PR #70, opened 2017-06-23
More fine-grained control over logging that allows us to use log levels.

ugermann:
These changes give us more fine-grained control over logging, allowing
us to send log messages at different log levels (not just info).

LOG(logger)->info()
LOG(logger)->warn()
etc.

### DIFF
--- a/src/command/s2s_translator.cpp
+++ b/src/command/s2s_translator.cpp
@@ -11,7 +11,7 @@ int main(int argc, char** argv) {
 
   boost::timer::cpu_timer timer;
   task->run();
-  LOG(info, "Total time: {}", timer.format());
+  LOG(info)->info("Total time: {}", timer.format());
 
   return 0;
 }
--- a/src/common/logging.cpp
+++ b/src/common/logging.cpp
@@ -24,6 +24,11 @@ std::shared_ptr<spdlog::logger> stderrLogger(
   return logger;
 }
 
+Logger checkedLog(std::string logger) {
+  Logger ret = spdlog::get(logger);
+  return ret ? ret : spdlog::get("devnull");
+}
+
 void createLoggers(const marian::Config* options) {
   std::vector<std::string> generalLogs;
   std::vector<std::string> validLogs;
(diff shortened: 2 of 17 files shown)


emjotde:
Can you then use this to set verbosity?

ugermann:
Yes. 

```src/3rd_party/spdlog/common.h``` defines
```
namespace spdlog
{

class formatter;

namespace sinks
{
class sink;
}

using log_clock = std::chrono::system_clock;
using sink_ptr = std::shared_ptr < sinks::sink >;
using sinks_init_list = std::initializer_list < sink_ptr >;
using formatter_ptr = std::shared_ptr<spdlog::formatter>;
#if defined(SPDLOG_NO_ATOMIC_LEVELS)
using level_t = details::null_atomic_int;
#else
using level_t = std::atomic<int>;
#endif

using log_err_handler = std::function<void(const std::string &err_msg)>;

//Log level enum
namespace level
{
typedef enum
{
    trace = 0,
    debug = 1,
    info = 2,
    warn = 3,
    err = 4,
    critical = 5,
    off = 6
} level_enum;
```

Then, e.g. logger->set_level(spdlog::level::debug); The only thing to keep in mind is that as log levels go up, verbosity goes down.

**Comment by emjotde** [comment]
Cool, can we add the command line option for that, too?

**Comment by ugermann** [comment]
Working on it. Are options defined only in training/config.cpp, or elsewhere as well?

**Comment by ugermann** [comment]
training/config.cpp now offers the option --verbose={trace|debug|info|warn|error|critical|off}. Should we call the option 'loglevel' instead?

**Comment by tomekd** [comment]
@ugermann : log level says everything.

**Comment by ugermann** [comment]
Since there were no objections, I merged my change to the logging mechanism into master.
@emjotde @tomekd @snukky: Note that now instead of
````
LOG(logger, <arglist>)
```
You should write
```
LOG(logger)-><loglevel>(<arglist>)
```
with <loglevel> being any of trace(), trace_if(), debug(), debug_if(), info(), info_if(), warn(), warn_if(), error(), error_if(), critical(), critical_if(). See the spdlog documentation.

PR #75, opened 2017-06-27
Fixed bug #74 with new logging mechanism.

### LINKED ISSUE #74: Segmentation fault from logger in s2s and rescorer
Logging is not yet initialized in `addOptions`, so we get segmentation fault in s2s and rescorer from here: 2def3251add6af4d959db127bfd96cd1c1de90e6:

    0x0000000000444cbf in spdlog::details::line_logger 
        spdlog::logger::_log_if_enabled<>(spdlog::level::level_enum, char const*) ()

I temporally commented the logging, but a fix is needed.

Another thing: is "verbose" [here](https://github.com/marian-nmt/marian-train/blob/master/src/common/logging.cpp#L76) correct?

ugermann:
Fixes #74, methinks. Not elegant, but works for the time being.

### DIFF
--- a/src/common/logging.cpp
+++ b/src/common/logging.cpp
@@ -73,8 +73,8 @@ void createLoggers(const marian::Config* options) {
   Logger devnull{stderrLogger("devnull", "%v")};
   devnull->set_level(spdlog::level::off);
 
-  if (options->has("verbose")) {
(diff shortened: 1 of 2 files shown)


snukky:
Already fixed with 81e5b72.

PR #82, opened 2017-07-05
Fix cuDNN 6 compilation

tomekd:
The problem was changed API for CNN in cuDNN 6.


tomekd:
Tested on zisa with local cuDNN 6 and fatboy (both with USE_CUDNN on and off). @geovedi confirmed it works for him too.

Issue #89, opened 2017-07-11
All cudaMalloc calls should be wrapped in CUDA_CHECK


XapaJIaMnu:
Hey,

I was getting some weird behavior when using the gradient drop, when approaching the maximum GPU memory, but I ended up in an infinite loop in stead of crashing. A quick search in the code revealed a few locations where cudaMalloc is not wrapped in guards. Those should be fixed.

https://github.com/marian-nmt/marian-train/blob/master/src/training/sparse_tensor.h#L65
https://github.com/marian-nmt/marian-train/blob/master/src/kernels/sparse.h#L89
https://github.com/marian-nmt/marian-train/blob/master/src/training/dropper.h#L111

In addition to that, we seem to have two guards that do exactly the same. One of them (I guess the nth_element one, because it's less idiomatic) should go.
https://github.com/marian-nmt/marian-train/blob/master/src/translator/nth_element.cu#L7
and 
https://github.com/marian-nmt/marian-train/blob/master/src/kernels/cuda_helpers.h#L7

If you agree with that I can quickly fix that and do a pull request. (Or i can haz write access to repo?)

Cheers,

Nick

emjotde:
You can haz, but there be dragons in these areas of code, so careful please.

emjotde:
@afaji Alham, this seems to be mostly your gradient dropping stuff.

XapaJIaMnu:
I promise to not edit commit before asking the authors.

XapaJIaMnu:
Fixed by 588d317

PR #90, opened 2017-07-17
Fix guards in a few files

XapaJIaMnu:
This fixes most outstanding issues in #89. There's plenty of outstanding issues with cudaMemCpy but those are less likely to fail silently compared to cudaMalloc calls.


emjotde:
Those silent fails are also due to exceptions not passing through thread termination. This is something @kpu might want to take a look at when he's back.

tomekd:
@emjotde : I need to find time and converse nth_element into a node look-alike object and withdraw memory allocation in this place. I need to have some workspace memory in this node. It seems to be ease to implement.

PR #93, opened 2017-07-20
Optimizer work

XapaJIaMnu:
This adds two different things which improve convergence: Automatically adjusting the learning rate and delaying the gradient update for a few batches, essentially faking large batches.


emjotde:
Thanks. And AsyncGraphGroup continues growing into a huge abomination.

PR #94, opened 2017-07-22
Style changes

DianaDespa:
Most of the changes are in graph_group.h and dropper.h, some variables were renamed for better code comprehension and to follow the camel case style.


emjotde:
Can you elaborate a bit what is happening and why this should go into master?

DianaDespa:
Hi, I'm an MSc student at the University of Edinburgh and I'm working on an optimisation for the sparse gradients, which mainly involves changes in the graph_group.h and dropper.h files. While trying to understand the existing code in those files, I thought it would be useful to rename some parameters/methods to more intuitive names. Also, I changed some of the existing names to camel case style, which I noticed is the predominant style throughout the repository. The first few commits are related to some experiments I was doing before deciding to separate the style changes into their own branch, experiments which I removed in the latter commits. Also, some of them are just merges after I pulled from master.

XapaJIaMnu:
I'm getting slightly higher error in the first few batches with Diana's branch and --tau 4 and --batch-flexible-lr but looking at the code nothing has changed.... (out of two runs) Could you check on your system? I think it's just random, but it's 2 AM here and i'm not certain... I'll leave two running overnight and see.

kpu:
Hi Diana, can you break this into a smaller pull request and rebase?  For instance "remove leftover code" could be its own.

Issue #99, opened 2017-08-31
nematus branch vs. SUMMA config files


ugermann:
1. For the source vocabularies, SUMMA config files use `source-vocab` as the parameter name, while s2s in the nematus branch expects `vocabs`. 

   I can work around for now by renaming parameter in SUMMA config files, but we should come to an agreement as to which parameter name we will use.

2. SUMMA: weights is a dictionary, s2s expects a vector.

snukky:
_amun_ options are not fully compatible with _s2s_ options. `vocabs` follows the convention we use in _s2s_, which, for example, allows specifying multiple source vocabularies.

I don't think we need the full backward-compatibility with _amun_ options.

tomekd:
To my best knowledge, there was never a point to make amun and s2s config files compactible. Moreover, amun guesses a lot of model features (layer-normalisation, transition depth) from the model file, whereas s2s requires those in the config file.

emjotde:
@ugermann I can be persuaded to turn weights into a dictionary for s2s too. That feature was quite early and I used it mostly for my own experiments. Out of curiosity, how are you using those?

emjotde:
And considering the vocabularies: in s2s there is no clear concept of source/target vocabularies anymore. There are just vocabularies and the model components can use any of those at will, hence the list.

ugermann:
@ejotde: I'm currently trying to get s2s to translate with WMT17. I assumed that the config files that Tomasz provided for the amun-based translation server in SUMMA were compatible, but I understand now that they are not. In the issues I just started listing things that s2s stumbled over when I tried to use SUMMA config files, under the misconception that they were compatible with s2s. I've now started to document the mapping from Nematus parameters to s2s parameters [here](https://github.com/marian-nmt/marian-dev/wiki/Mapping-Nematus-Model-Parameters-to-S2S). If someone wants to help out with the mapping, that would be great. Otherwises I'll try to solve the puzzle one parameter at a time.

snukky:
@ugermann: Here is an example of the config.yaml which I use with WMT17 en-de model:
```
# performance settings
beam-size: 5
normalize: yes
devices: [1]
workspace: 1024

# scorer configuration
type: nematus
enc-depth: 1
enc-cell-depth: 4
enc-type: bidirectional
dec-depth: 1
dec-cell-base-depth: 8
dec-cell-high-depth: 1
dec-cell: gru-nematus
enc-cell: gru-nematus
tied-embeddings: true
[...]

PR #100, opened 2017-08-31
Bug fix in: i < vocabPaths.size() - 1 => i + 1 < vocabPath.size()

ugermann:
in corpus.cpp

If vocabPaths.size() is empty in translation mode, s2s segfaults because 0 < size_t(0)-1 is always true.
Needless to say, vocabPaths should NOT be empty, but s2s should complain and exit instead of segfaulting.

See also #99.


emjotde:
Can you maybe make it complain, too? This is a good point.

ugermann:
Just added some checks that will throw exceptions with hopefully useful error messages.

PR #109, opened 2017-10-09
Fix python library path

lparam:


snukky:
Thanks. This will work with marian repo, but not with marian-dev used solely. 

BTW, if you need s2s in server mode, check `server` command.

lparam:
Do yout mean s2s_server.cpp?

snukky:
Yes. This corresponds to `build/server`.

snukky:
Open a new issue/question at the google group and provide `.npz.json` file. My guess is that you are using a WMT16 Edi model and you probably need `--dim-emb 500`. But that's only a guess.

emjotde:
Are these the en-de models? s2s does not currently recognize model properties for models that have been trained with nematus and now with marian. Here you would need to provide the model parameters:

`~/marian-dev/build/s2s -d 0 -m en-de/model-ens?.npz -v en-de/vocab.en.json en-de/vocab.de.json -i data/newstest2015.ende.en --dim-vocabs 85000 85000  --dim-emb 500`

For a model trained with marian this works automatically. @snukky did we plan to have a script that can do the conversion?

lparam:
Thanks to both of you. 
After adding “--dim-vocabs 85000 85000 --dim-emb 500”, it works fine.

PR #131, opened 2017-11-09
Update the way batch_flexible_lr is applied

XapaJIaMnu:
This is more of Request for comment than a pull request. It compiles but it's not completely tested.

I have simplified the way batch-flexible-lr is applied, using an array of the size of devices to keep what the last number of words for this batch was.

We can move this code to the scheduler pretty much as it is by adding functions such as "scheduler_->SetbatchSize(batch, threadID)" and "scheduler->getLRforThread(threadID)" but it'd not really simplify the code further, it will just call the scheduler instead of doing the work directly. If you have other ideas, please suggest!.

### DIFF
--- a/src/data/batch.h
+++ b/src/data/batch.h
@@ -11,6 +11,12 @@ class Batch {
 public:
   virtual size_t size() const = 0;
   virtual size_t words() const { return 0; };
+  virtual size_t width() const { return 0; };
+
+  virtual size_t sizeTrg() const { return 0; };
+  virtual size_t wordsTrg() const { return 0; };
+  virtual size_t widthTrg() const { return 0; };
+
   virtual void debug(){};
 
   virtual std::vector<Ptr<Batch>> split(size_t n) = 0;
--- a/src/data/corpus.h
+++ b/src/data/corpus.h
@@ -172,6 +172,14 @@ public:
 
   size_t words() const { return batches_[0]->batchWords(); }
 
+  size_t width() const { return batches_[0]->batchWidth(); }
+
+  size_t sizeTrg() const { return batches_.back()->batchSize(); }
+
+  size_t wordsTrg() const { return batches_.back()->batchWords(); };
+
+  size_t widthTrg() const { return batches_.back()->batchWidth(); };
+
   size_t sets() const { return batches_.size(); }
 
   static Ptr<CorpusBatch> fakeBatch(std::vector<size_t>& lengths,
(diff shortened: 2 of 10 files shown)


snukky:
Scheduler and TrainingState implement an observer pattern, so for SingletonGraph, I would propose something like the following:

1. add actAfterUpdate() (or actBeforeUpdate() if you wish) into TrainingObserver
2. implement this method in OptimizerBase
3. add newUpdate() into TrainingState
4. call newUpdate() before opt_->update() in SingletonGraph (through the scheduler)
5. implement logic in actBeforeUpdate() in Scheduler; use config options here
6. remove multiplyFactor_ from Optimizer, including update() methods which take it as an argument
7. remove scaleLearningRate_ and avgBatchWords_ from GraphGroup

This will require adding the size of the current batch into a TrainingState or passing it (or a Ptr<Batch>) into newUpdate and actBeforeUpdate.

For multiple workers, we would probably need multiple etas in TrainingState, and newUpdate and actBeforeUpdate methods should takes another argument indicating the ID of the worker.

emjotde:
Multiple workers will always work in different threads. Make eta thread local instead of having multiple etas. Unless this is done already?

XapaJIaMnu:
eta can't be thread_local because it's a class member... I can make it static and thread_local, but that feels wrong....

XapaJIaMnu:
Hey,

Could you look at the second attempt. This only addresses the graphGroupSingleton case for now, I have not implemented the multigraph yet, requesting comments first.

Issues that I identify with this approach:
1) We can't do thread local in the member variables so we have to  arrays for each thread.
2) If this approach is to work with delayed updates, we need another scheduler call "ActAfterOptimizer", so that we know when an optimizer step has been performed and we can reset the counter of words seen so far, which would seem veeery clunky to me.

Is there another cleaner way to include this in the scheduler? This is what I understood from Roman's comments.

Cheers,

Nick

emjotde:
@XapaJIaMnu could show something of this working in the end?

XapaJIaMnu:
Hey,
Sorry for the late reply! I have done some parameter sweeps and I can say there's a small benefit of using flexible learning rate when the mini-batches are very imbalanced (eg --maxi-batch 100 together with a shallow model and a very large -w and --optimizer-delay). In the rest of the cases the variance between mini-batches is not that great.

emjotde:
I will allow myself to close this without pulling then. We should probably remove the batch-flexible stuff then from master too?

PR #135, opened 2017-11-13
Tidy up CuDNN code

tomekd:
Tiding up cudnn depended code. I moved the code to new files. Moreover, the ``#ifdef CUDNN`` occurs only in those files.  I didn't change cmake configuration, so in default, so the compilation with cudnn is turned off. What I changed is the presence of e.g. mnist-lenet model. Whether someone tries to run mnist-lenet model without cudnn support, they get an informative message .


snukky:
All regression tests passed, including the updated test for MNIST LeNet, which uses CUDNN.

PR #138, opened 2017-11-20
reimplement gradient dropping

afaji:
Gradient Dropping re-implemented.

Additionally, I had to modify the graph_group_async class as follow:
 - create a function init(), and all variable and params initializations are now moved there (instead in execute())
 - Add an extra parameter "device_id" in pushGradient() and fetchParam().
 - make the functions and variable protected instead of private (so I can inherit them)

@kpu @emjotde


emjotde:
Looks alright. Did you run regression tests?

afaji:
manually. Do we have the automatic testing script now?

emjotde:
https://github.com/marian-nmt/marian-regression-tests

There's a README, you need to change the path in the bash script to point to your local copy with the correct branch. @snukky do we have anything that tests GraphGroupAsync?

snukky:
Not yet as our Jenkins works on a machine with 1 GPU. I'll add some multi-GPU tests just for local testing. Give me a day or two.

kpu:
Paging @dr-alan-turing who might want to base his code on this starting now. . .

emjotde:
Did you check the regression tests?

snukky:
I added a test for multi-GPU training. It will run only if you set the `NUM_DEVICES` environmental variable to > 1, e.g.:

     MARIAN=~/work/marian-dev NUM_DEVICES=2 ./run_mrt.sh

afaji:
how stable is the multi-gpu regression test? if you train on multi-gpu, the result is nondeterministic. I've run the async_sgd regression test with the unmodified marian (i.e from /tools/marian) and sometimes it failed.

snukky:
So it means it wasn't stable enough despite my 5/5 successful tests. I changed the tested conditions, so it should be more reliable now, but maybe the training is more nondeterministic on different machines/with different CUDA versions, etc.

afaji:
I've checked the regression test and obtained:
Ran 15 tests in 00:10:33.148s, 12 passed, 3 skipped, 0 failed

emjotde:
If @snukky says it's OK, I will pull it.

afaji:
any update on this?

emjotde:
Ping @snukky

snukky:
All regression tests pass after the merge. It seems that the last changes in master were not related to async graph group much. If training yet another example model is not required after the merge, we can pull.

emjotde:
@afaji @kpu 
BTW, would be interesting to test with the Transformer model. It is notoriously sensitive to stale gradients, but has a small model size even for large depths. Might be an interesting challenge?

PR #146, opened 2017-11-30
Add CUDNN flag in config parser

lparam:


emjotde:
We just did that in parallel. Thanks for noticing. Already updateing the regession tests to keep that from happening.

PR #151, opened 2017-12-13
CUDA 9 and gcc 6.x

tomekd:
Officially, cuda 9 supports gcc 6.x as a host code compiler. And it does in 95%. There is a well-known bug in this set-up, for instance it occured here:
https://github.com/ComputationalRadiationPhysics/alpaka/issues/426
https://github.com/pytorch/pytorch/issues/3963

The problem occurs also in marian. I workaround it by splitting the code into two cpp/cu files.


emjotde:
Do I see correctly, that there are now *.cpp and *.cu files? If that is the case and it is a bug in CUDA with gcc6 support I would rather wait for Nvida to fix that bug than adding this to the code base.

tomekd:
Agree. I don't push to merge it, but I want to keep and maintain it for local running.

XapaJIaMnu:
Cuda 9.1 is out, does it fix the issue for you?

hieuhoang:
CUDA 9.1 needs the latest driver. But it's also a tinny bit faster (5-10%)

XapaJIaMnu:
I believe this may need to be merged. It's impossible to compile using CUDA 9+ and nvidia didn't fix it in the last release.

tomekd:
@XapaJIaMnu : you know, there are only two of us who have such problem now :)
 More seriously, I'll wait til March 2018, when Ubuntu announces the next LTS version. I've seen that they're planning to move GCC to version 6/7.

emjotde:
Yes, please wait with this until it becomes more pressing. And even then I would prefer a different solution than the one proposed.

tomekd:
@XapaJIaMnu, @filipg I found easier solution for Arch users:
```
$ wget https://archive.archlinux.org/packages/g/gcc5/gcc5-5.4.0-1-x86_64.pkg.tar.xz
# pacman -U gcc5-5.4.0-1-x86_64.pkg.tar.xz
```
And compile marian with a flag ``-DCUDA_HOST_COMPILER=/usr/bin/gcc-5 ``.

PR #157, opened 2018-01-09
Add sparse momentum correction and masking

afaji:
Implement momentum correction and momentum masking on gradient dropping as in https://arxiv.org/pdf/1712.01887.pdf

Passed regression test


snukky:
It looks OK for me.

PR #178, opened 2018-02-14
Marian on CNTK (front-end). 

frankseide:
Temporary pull request for comparing the changes to make the transformer front-end module compile under Marian-on-CNTK/Dynamite.


frankseide:
With these modifications, all Marian models (except char-s2s) can be compiled unmodified as front-ends to CNTK Dynamite, on top of the Marian-on-CNTK layer.

(Supporting char-s2s requires some refactoring of the Convolution operation.)

emjotde:
Have you checked the regression tests?

PR #198, opened 2018-03-01
Add the ability to get statistics about target words

XapaJIaMnu:
Do batch flexible LR based on target words and add the ability to see the number of target words in a mini-batch.


emjotde:
Hi, could we make sure that things like that go through a review first?

Issue #195, opened 2018-02-27
Custom log message during training


emjotde:
Just a random thought. Would be nice to able to modify the the log message during training

```
Ep. 1 : Up. 6000 : Sen. 1193481 : Cost 2.68 : Time 86.83s : 32988.54 words/s
```

By providing a template, e.g.

```
--log-template "Ep. {%EPOCH%} : Up. {%UPDATE%} : Sen. {%SAMPLES%} : Cost {%COST%} : Time {%BATCH_SECONDS%} : {%WORDS_PER_SECOND%} words/s
```

And them provide more options like:
```
{%LR%}
{%TOTAL_WORDS%}
{%TOTAL_TIME%}
...
```

Not too urgent by kind of cute.

emjotde:
One could for instance generate latex tables from that directly :)

emjotde:
@ugermann maybe?

frankseide:
I locally implemented a modification that logs ce loss per target word, along with associated number of target words scored. This makes losses better comparable across systems and parallelization configurations, which I find useful. Example:
```
[2018-02-22 20:39:54] Ep. 1 : Up. 500 : Sen. 350559 : Cost 9.89 * 10696858 : Time 482.42s : 22173.46 words/s : L.r. 9.3750e-06
[2018-02-22 20:47:18] Ep. 1 : Up. 1000 : Sen. 699788 : Cost 8.54 * 10581251 : Time 444.68s : 23794.96 words/s : L.r. 1.8750e-05
[2018-02-22 20:54:42] Ep. 1 : Up. 1500 : Sen. 1047256 : Cost 7.92 * 10575207 : Time 443.88s : 23824.48 words/s : L.r. 2.8125e-05
```
This requires to recover the normalization done in `Cost()` and undo it. Here is the modified code in `scheduler.h`:
```
  void update(float cost, Ptr<data::Batch> batch) {
    auto batchSize = batch->size();
    auto batchTargetWords = batch->words(-1);
    // reconstruct sum cost
    auto costType = options_->get<std::string>("cost-type");
    cost *= // what was cost normalized with?
      /*if*/ (costType == "ce-sum") ?
        1
      /*else if*/ : ((costType == "ce-mean-words") ?
        batchTargetWords
      /*else*/ :  // use ce-mean for all others (not correct for some)
        batchSize);
    // undo averaging over workers
    cost *= options_->get<std::vector<size_t>>("devices").size();

[...]

PR #212, opened 2018-03-19
compatibility with MS-internal Philly cluster

frankseide:
This PR contains changes to make Marian compatible with the Microsoft-internal compute cluster. It also has a few changes made on the way:

  * vocabularies can be read from flat files
  * comments of dimensions in `transformer.h`
  * word penalty parameter
  * logged losses are now normalized per target word

(Note: please ignore the list of commit comments below, as it includes several temporary changes that have been removed again.)


emjotde:
Can you comment or show how the logging looks like now? Did you run regression tests with that?

frankseide:
Logging now reports the Cost per word, for ce-mean, ce-sum, ce-mean-words, "times" the number of target labels that the Cost was measured on. This allows to compare progress of ce-sum runs across different minibatch sizes. The log now also shows the aggregate number of target labels processed so far, which makes it easier to compare runs that ended up using different MB size (the Epoch number only allows that at epoch boundaries). These changes are aimed at reducing mental math. I also increased the number of decimals for the Cost, as that allows direct comparison of runs that are expected to be identical. Example of logging:

```
[2018-03-19 20:09:23] Ep. 38 : Up. 193500 : Sen. 3129009 : Cost 2.97881484 * 13341033 after 5203488403 : Time 635.33s : 20998.72 words/s : L.r. 8.6266e-05
[2018-03-19 20:20:04] Ep. 38 : Up. 194000 : Sen. 3582181 : Cost 2.95717525 * 13717641 after 5217206044 : Time 641.33s : 21389.45 words/s : L.r. 8.6155e-05
[2018-03-19 20:30:11] Ep. 38 : Up. 194500 : Sen. 4017920 : Cost 3.00738382 * 13358969 after 5230565013 : Time 606.33s : 22032.40 words/s : L.r. 8.6044e-05
[2018-03-19 20:40:17] Ep. 38 : Up. 195000 : Sen. 4441780 : Cost 2.97530913 * 12928926 after 5243493939 : Time 605.95s : 21336.71 words/s : L.r. 8.5934e-05
```

emjotde:
This will surely break a large number of our regression tests. Can we turn this into an option or solve this later with issue #195 ?

emjotde:
Memo to self. Wait with pull for update.

PR #215, opened 2018-04-04
fix scheduler update on delayed gradient

afaji:
@XapaJIaMnu @kpu this is my fix on delayed gradient updates:
 instead of calling scheduler every batch, we call scheduler every tau instead.

### DIFF
--- a/src/training/graph_group_async.cpp
+++ b/src/training/graph_group_async.cpp
@@ -1,6 +1,7 @@
 #include "training/graph_group_async.h"
 #include "functional/functional.h"
 #include "tensors/tensor_operators.h"
+#include "data/corpus_base.h"
 
 namespace marian {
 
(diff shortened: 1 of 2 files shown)


src/training/graph_group_async.cpp
```diff
@@ -207,30 +208,48 @@ void AsyncGraphGroup::execute(Ptr<data::Batch> batch) {
...
       // Wait until the thread that wants to do validation is finished.
       pool_->wait_for_one(lock);
 
-      scheduler_->update(cost, batch);
+      if (options_->get<std::string>("cost-type") != "ce-sum")
+        cost /= tau_;
+
+      if (tau_ > 1) {
+        std::vector<size_t> fakeLength = {1, 1}; 
+        auto fb = data::CorpusBatch::fakeBatch(fakeLength,
```
kpu: This seems hacky?

afaji: @kpu it is. Because we need to inform the scheduler the accumulated batch for statistic purpose (wps, when to start a new epochs, etc2). This is the "best" way I can find to create a fake batch for that purpose.

emjotde: Urgh. Hate this. Why update the scheduler at all?

PR #220, opened 2018-04-09
Fix ENDIF clause condition

CatarinaSilva:


CatarinaSilva:
@XapaJIaMnu @emjotde

emjotde:
And thanks again.

PR #238, opened 2018-05-13
Implement sychronous graph group multi-node

XapaJIaMnu:
I have not tested it yet, but I would appreciate comments. Will test it tomorrow.


emjotde:
Hi, we are in full-shared-task mode for the next few days. Will make sure to check it out during the week.

XapaJIaMnu:
Hey,
I didn't tag you or @snukky because it's an alpha version. I need to iron it out quite a bit before presenting it. There also number of places where memory access could be optimized.

PR #242, opened 2018-05-24
Add multi-node synchronous graph group

XapaJIaMnu:
Hey,
@emjotde @snukky @afaji @kpu 
This is our initial graph group multinode sync implementation based on MPI. The version should work with both CPU and GPU backend. We might be able to save one memory copy call, but other than that it should be quite optimal usage of MPI.

KNOWN ISSUES:
When using mini-batch-fit we run into a memory leak on GPU0. Not sure what the issue is yet, investigating.
Model saving results in a crash.


afaji:
More info on known issues about memory leak:
Only happens when using memory-fit mini-batch AND on CSD3. Cannot reproduce this issue on valhalla.

PR #247, opened 2018-06-06
Remove leftover references to Python bindings

### LINKED ISSUE #246: Running the python interface from example.py
Hello!

I'm trying to run `scripts/python/example.py` from the current master to look into Marian's Python interface. As I understand it, the compilation process is supposed to produce something like a `libmariannmt.so` file to be imported here:

https://github.com/marian-nmt/marian-dev/blob/8fbfa6560f10fff295f1f1b53372e15ae3549fa4/scripts/python/example.py#L7

However, I dug through the CMakes and found nothing related to generating this file, except for a commented out portion:

https://github.com/marian-nmt/marian-dev/blob/8fbfa6560f10fff295f1f1b53372e15ae3549fa4/src/CMakeLists.txt#L149-L151

Interestingly, in marian-nmt/marian, there are some Python related lines in CMakes, for example here:

https://github.com/marian-nmt/marian/blob/8b4e705a38f53d0ebb721eb1161a2e456ebf3455/CMakeLists.txt#L106-L113

But I can't seem to find those here, in marian-dev.

Thus, the question: is this `example.py` script still possible to run somehow? Is it left-over legacy code? Am I missing something silly and obvious with respect to compilation?

Thank you for your time!

StanczakDominik:
Closes #246, to which this PR is a direct followup.

I basically went through this with `ag python` and removed stuff that looked like legacy code.

It's a shame that my first contribution to the project is a completely negative one :disappointed: :laughing:


snukky:
Thanks! 

I think a proper optionally-compiled python bindings may be beneficial for the community, so you may try implementing that if you want to make a more challenging contribution. I'm willing to help. I'm just not sure we should use Boost for this, though.

The best use case I can think of would be decoding a language model trained with Marian from a Python script.

PR #272, opened 2018-08-02
Saving exponential smoothing for sync/async training

snukky:
This part needs a quick review:

* https://github.com/marian-nmt/marian-dev/blob/merge-exp-smooth/src/training/graph_group_sync.cpp#L59
* https://github.com/marian-nmt/marian-dev/blob/merge-exp-smooth/src/graph/expression_graph.h#L178


emjotde:
Why is the forceInit required for all nodes other than parameters?

Issue #266, opened 2018-07-09
Batched translation speed is very slow compared to Fairseq


wenyong-h:
I'm benchmarking marian and fairseq.
The model i'm using is Transformer big, beam size is 3.
The gpu i'm using is p100 with cuda 9.0.
Deocder argument is setted following https://marian-nmt.github.io/docs/#batched-translation.

The result i'm getting is in the following table.
As you can see, the speedup is much worse compared to Fairseq as batch size increase.
The result i'm getting is so different from https://twitter.com/marian_nmt/status/1008476022478532608.
Do you know what might be wrong?

| decoder        | batch size           | wps  |
| ------------- |:-------------:| -----:|
| marian 1.4    | 1 | 101 |
| marian 1.5     | 1 | 104 |
| fairseq     | 1 | 88 | 
| marian 1.4   | 16 | 268 |
| marian 1.5   | 16 | 345 |
| fairseq         | 16 | 881 |
| marian 1.4   | 32 | 267 |
| marian 1.5   | 32 | 376 |
| fairseq         | 32 | 1482 |

emjotde:
Hi,
The workspace setting here is for a much smaller model, I suppose it is re-allocating memory at the beginning and that is slow as hell. Try

```
./marian-decoder -m model.npz -v vocab.src.yml vocab.trg.yml -b 3 --normalize 0.6 \
    --mini-batch 64 --maxi-batch-sort src --maxi-batch 100 -w 10000 --max-length-factor 1.5
```

wenyong-h:
I have already set workspace bigger than 10000.
But i didn't set max-length-factor, after set it to 1.5, 
with marian 1.5, batch size 32,  wps doubled, becomes 700.
but fairseq under the same settings, wps is 1550.

I did some profiling using nprof, found that , with Marian, GEMM is taking much bigger percent of time than Fairseq.  
Then i found Fairseq is using cuda 8.0,while Marian is using cuda 9.0
So maybe gemm in cuda 9.0 have some performance problem.
so, i'm going to first try Marian with cuda 8.0, see how it turns out.

hieuhoang:
your analysis is probably correct, I've also found scalability issues in Marian, as discussed here
  https://github.com/marian-nmt/marian/issues/96
  https://arxiv.org/pdf/1805.09863.pdf
Probably due to the naive mini-batching implementation which should hopefully be sorted out soon.

As an extra big-up for my paper, it also has an optimization for your beam=3 scenario

[...]

PR #273, opened 2018-08-04
avoid cudaMalloc in CopyRows

wenyong-h:
This could increase wps of Transformer big by 250 when batch size is 32, see #266


emjotde:
Very nice! I intended to do this for while since we introduced typed tensors a while ago.

PR #303, opened 2018-09-12
Replace boost::program_options with a new CLI

snukky:


src/command/marian_conv.cpp
```diff
@@ -1,56 +1,36 @@
...
   configStr << config;
 
   auto graph = New<ExpressionGraph>(true, false);
   graph->setDevice(CPU0);
 
-  graph->load(vm["from"].as<std::string>());
+  graph->load(modelFrom);
   graph->forward();
-  graph->save(vm["to"].as<std::string>(), configStr.str());
+  graph->save(modelFrom, configStr.str());
```
emjotde: modelTo

src/common/config_parser.cpp
```diff
@@ -72,997 +37,712 @@ const std::set<std::string> PATHS = {"model",
...
+      "Display information every  arg  updates",
+      1000);
+  cli.add<bool>("--disp-label-counts",
+      "Display label counts when logging loss progress");
+  cli.add<size_t>("--save-freq",
+      "Save model file every  arg  updates",
+      10000);
+
+  addSupoptionsInputLength(cli);
+
```
emjotde: Sup - > Sub?

src/common/some_type.h
```diff
@@ -0,0 +1,133 @@
+#pragma once
+
+#include <utility>
+
+namespace marian {
+
+class some_type {
```
emjotde: Is this a reimplementation of `boost::any`? How about `any_type`?

snukky: Name changed

snukky:
It seems the proper order is answering review comments and then making changes.

src/common/cli_wrapper.h
```diff
@@ -0,0 +1,275 @@
...
+  // Name of the default option group
+  std::string defaultGroup_{""};
+  // Name of the current option group
+  std::string currentGroup_{""};
+
+  static std::string failureMessage(const CLI::App *app, const CLI::Error &e);
+
+  std::string keyName(const std::string &args) const {
+    return std::get<1>(CLI::detail::get_names(CLI::detail::split_names(args)))
+        .front();
```
frankseide: What is this specifically? Esp. why front(), why are there multiple? Add comment

src/common/cli_wrapper.h
```diff
@@ -0,0 +1,275 @@
...
+  }
+
+public:
+  /**
+   * @brief Create an instance of the command-line parser
```
frankseide: I would expect such a constructor to take argc and argv as arguments. Could you add a comment where those are fed in, and why this class must be stateful (where argc/argv are fed in later in a state update rather than creating an immutable class that gets constructed from the command line and is its representation)?

Maybe just roughly outline its use; something like First, call add() to create all teh options; then call ... to parse the cmd line, to make orientation for the reader easier.

Maybe the concept of option groups shoudl be touched upon here as well?

src/common/cli_wrapper.h
```diff
@@ -0,0 +1,275 @@
...
+   *
+   * Option --help, -h is automatically added.
+   *
+   * @param name Header for the main option group
+   * @param columnWidth Width of the column with option names
+   * @param screenWidth Maximum allowed width for help messages
+   */
+  CLIWrapper(const std::string &name = "General options",
+             size_t columnWidth = 35,
+             size_t screenWidth = 0);
```
frankseide: Does 0 have a special meaning? add comment

src/common/cli_wrapper.h
```diff
@@ -0,0 +1,275 @@
...
+   *
+   * @param args Comma-separated list of short and long option names
+   * @param help Help message
+   * @param val Default value
+   *
+   * @return Option object
+   */
+  template <typename T>
+  CLI::Option *add(const std::string &args, const std::string &help, T val) {
+    return add_option<T>(keyName(args), args, help, val, true, true);
```
frankseide: For these 'true' 'true' parameters, please use an inline comment to include the parameter name, e.g.

, /*x=*/true, /*y=*/true)

src/common/cli_wrapper.h
```diff
@@ -0,0 +1,275 @@
...
+  }
+
+  /**
+   * @brief Define an option without an explicit default value. The implicit
+   * default value is T()
+   *
+   * The option will be defined in the config file even if not given as a
```
frankseide: A more general question: We have run into situations where Marian runs as a library, and models are loaded that were created before a specific parameter existed. Those will fail when the option is read out. The fix is always to add a default to the read-out operation. So in that context, should we at all have options that have implicit defaults? Why not just always use defaults in the get() calls?

src/common/cli_wrapper.h
```diff
@@ -0,0 +1,275 @@
...
+   * command-line argument. The implicit default value for a numeric option is
```
frankseide: Should we have default defaults? Would it not be clearer to require to always state the default when creating teh parser?

src/common/cli_wrapper.h
```diff
@@ -0,0 +1,275 @@
...
+
+  // Get the current value for the option
+  template <typename T>
+  T get(const std::string &key) const {
+    ABORT_IF(
+        vars_.count(key) == 0, "An option with key '{}' is not defined", key);
+    return vars_.at(key)->as<T>();
+  }
+
+  YAML::Node getConfig() const;
```
frankseide: How does this class relate to YAML::Node? Is this function just an accessor, or does it perform an actual deep conversion of a data structure? Say both in a comment.

src/common/cli_wrapper.h
```diff
@@ -0,0 +1,275 @@
...
+   * unchanged
+   *
+   * @param node YAML config with new default values for options
+   */
+  void overwriteDefault(const YAML::Node &node);
+
+private:
+  template <
+      typename T,
+      CLI::enable_if_t<!CLI::is_bool<T>::value && !CLI::is_vector<T>::value,
```
frankseide: For non-trivial template magic, a comment should say what this is for and why

src/common/cli_wrapper.h
```diff
@@ -0,0 +1,275 @@
...
+                          T val,
+                          bool defaulted,
+                          bool addToConfig) {
+    // std::cerr << "CLI::add(" << key << ") as vector" << std::endl;
+
+    if(addToConfig)
+      config_[key] = val;
+    vars_.insert(std::make_pair(key, std::make_shared<any_type>(val)));
+
+    CLI::callback_t fun = [this, key](CLI::results_t res) {
```
frankseide: One-line comment what this function does and what it is for

src/common/cli_wrapper.h
```diff
@@ -0,0 +1,275 @@
...
+      opt->group(currentGroup_);
+    if(defaulted)
+      opt->default_str(CLI::detail::join(val));
+
+    opts_.insert(std::make_pair(key, opt));
+    return opts_[key];
+  }
+
+  template <typename T,
+            CLI::enable_if_t<CLI::is_bool<T>::value,
```
frankseide: cf. commetn above

src/common/config.cpp
```diff
@@ -11,10 +12,76 @@ namespace marian {
 
 size_t Config::seed = (size_t)time(0);
 
+Config::Config(const std::string options,
+               cli::mode mode /*= cli::mode::training*/,
```
frankseide: What is this mode for? How is a command-line parser modal? Add a comment

src/common/config.cpp
```diff
@@ -11,10 +12,76 @@ namespace marian {
...
+               bool validate /*= false*/) {
+  std::vector<std::string> sargv;
+  utils::Split(options, sargv, " ");
```
frankseide: What is the options string? This looks very dangerously as not parsing a command line correctly, w.r.t. escaping etc. Maybe this funcition needs a comment stating what it is for and in which context it is meant to be used.

src/common/config.cpp
```diff
@@ -27,21 +94,13 @@ YAML::Node& Config::get() {
...
-void Config::log() {
-  YAML::Emitter out;
-  cli::OutputYaml(config_, out);
-  std::string configString = out.c_str();
-
-  std::vector<std::string> results;
-  boost::algorithm::split(results, configString, boost::is_any_of("\n"));
-  for(auto& r : results)
-    LOG(info, "[config] {}", r);
+const std::vector<DeviceId>& Config::getDevices() {
```
frankseide: I remember a funciton called getDevices() that performs complex parsing of the device string. Is this teh same function, and what happened to it? Or if not, why is there a special accessor for devices, but not for other parameters? Does this belong into this class?

src/common/config.cpp
```diff
@@ -56,4 +115,21 @@ void Config::loadModelParameters(const void* ptr) {
   override(config);
 }
 
+void Config::override(const YAML::Node& params) {
```
frankseide: Please add a comment to specify exactly how that override process happens. E.g. does it replace entire nodes (so that existing entries in those nodes get deleted if the new one does not have one of the same key), or is the result the union set, with parameters occuring in both set to the overriding one?

src/common/config.cpp
```diff
@@ -56,4 +115,21 @@ void Config::loadModelParameters(const void* ptr) {
...
+  for(auto& it : params) {
+    config_[it.first.as<std::string>()] = it.second;
+  }
+}
+
+void Config::log() {
+  YAML::Emitter out;
+  cli::OutputYaml(config_, out);
```
frankseide: Why do we need this complicated interface? Why is there no single function in YAML that does this and just returns a string? I suggest to add such a function. The Emitter protocol seems an irrelevant concern at this level of code.

src/common/config.cpp
```diff
@@ -56,4 +115,21 @@ void Config::loadModelParameters(const void* ptr) {
...
+  std::string configString = out.c_str();
+
+  std::vector<std::string> results;
+  boost::algorithm::split(results, configString, boost::is_any_of("\n"));
```
frankseide: Please adda  one-line comment what this is (prints the stirng but prepending each line with [config]). Don't make readers reverse-engineer source code if not necessary.

src/common/config.h
```diff
@@ -26,74 +21,22 @@ class Config {
...
-      try {
-        if(!get<bool>("ignore-model-config"))
-          loadModelParameters(model);
-      } catch(std::runtime_error&) {
-        LOG(info, "[config] No model configuration found in model file");
-      }
-    }
-    log();
-
-    if(has("version"))
```
frankseide: Is this version info still printed? (please just make sure that it is)

src/common/config.h
```diff
@@ -108,24 +51,21 @@ class Config {
...
 
-  const YAML::Node& get() const;
-  YAML::Node& get();
-
   YAML::Node getModelParameters();
   void loadModelParameters(const std::string& name);
   void loadModelParameters(const void* ptr);
 
-  const std::vector<DeviceId>& getDevices() { return devices_; }
+  const std::vector<DeviceId>& getDevices();
```
frankseide: Why replace this function by a non-inlined one that does exactly the same thing? This makes the code harder to read, since the reader now must jump between two files. I suggest to revert this back.

src/common/config.h
```diff
@@ -108,24 +51,21 @@ class Config {
...
 
-  void save(const std::string& name) {
-    OutputFileStream out(name);
-    (std::ostream&)out << *this;
-  }
+  void save(const std::string& name);
```
frankseide: Is this save() function now more complex than before? If not, it may be appropriate to keep the original.

src/common/config_parser.cpp
```diff
@@ -16,50 +16,15 @@
...
 #include "common/cli_helper.h"
-#include "common/config.h"
 #include "common/config_parser.h"
+#include "common/config_validator.h"
 #include "common/file_stream.h"
 #include "common/logging.h"
-#include "common/regex.h"
-#include "common/version.h"
-
-#define SET_OPTION(key, type)                    \
```
frankseide: Great to see these macros go.

src/common/config_parser.cpp
```diff
@@ -72,997 +37,712 @@ const std::set<std::string> PATHS = {"model",
...
-     "All paths are relative to the config file location")
-    ("dump-config", po::value<bool>()->zero_tokens()->default_value(false),
-     "Dump current (modified) configuration to stdout and exit")
-    ("version", po::value<bool>()->zero_tokens()->default_value(false),
-      "Print version number and exit")
-    ("help,h", po::value<bool>()->zero_tokens()->default_value(false),
-      "Print this help message and exit")
-  ;
+  cli.add_nondefault<bool>("--version",
+      "Print version number and exit");
```
frankseide: I presume all of these are purely mechanical conversions without actual changes of strings or defaults?

src/common/config_parser.h
```diff
@@ -16,48 +13,74 @@
...
     parseOptions(argc, argv, validate);
   }
 
+  /**
+   * @brief Parse command-line options
+   *
+   * Options are parsed in the following order, later config options overwrite
+   * earlier:
+   *  * predefined default values
+   *  * options from the config files provided with --config, from left to right
```
frankseide: Does still work for interpolating environment variables? (Maybe could you add a test case? This is critical for our production training.)

src/common/config_parser.h
```diff
@@ -16,48 +13,74 @@
...
+   * @param argc
+   * @param argv
+   * @param validate Do or do not validate parsed options
+   */
   void parseOptions(int argc, char** argv, bool validate);
 
   YAML::Node getConfig() const;
   std::vector<DeviceId> getDevices();
 
+  // Check if the config has a boolean option for the given key and it is
```
frankseide: What if it has such option but it is not a boolean? I kind of lost the context where this is happening, but it seems that if it is next to parseOptions(), it would be clearer to rename this to hasBoolean()

src/common/config_parser.h
```diff
@@ -16,48 +13,74 @@
...
-  void addOptionsValid(boost::program_options::options_description&);
-  void addOptionsTranslate(boost::program_options::options_description&);
+  // Change relative paths to absolute paths relative to the config file's
+  // directory
+  void makeAbsolutePaths(const std::vector<std::string>&);
 
-  void validateOptions() const;
-  void validateDevices() const;
+  // Create a new config from all provided config files
+  YAML::Node loadConfigFiles(const std::vector<std::string>&);
```
frankseide: I don't understand this comment. What is that new config? How is it created from the config files? Is this just loading multiple config files into a YAML::Node? Maybe change the comment then.

The problem is that tehre are so many config classes in here, plus "configuration" can refer to the high-level concept of the entire program's configuration, and/or any of its representations. Let us avoid the term "config" and always qualify it what kind of config. E.g. "new Options instance"

src/common/config_validator.cpp
```diff
@@ -0,0 +1,142 @@
...
+#include "3rd_party/exception.h"
+#include "common/logging.h"
+#include "common/regex.h"
+#include "common/utils.h"
+
+#include <boost/filesystem.hpp>
+
+namespace marian {
+
+bool ConfigValidator::has(const std::string& key) const {
```
frankseide: Why would a validator be asked whether it has a key? Please add a comment. If this is a priviate method, it may clarify to add a comment /*private*/

src/common/config_validator.cpp
```diff
@@ -0,0 +1,142 @@
...
+      validateOptionsScoring();
+      break;
+    case cli::mode::training:
+      validateOptionsParallelData();
+      validateOptionsTraining();
+      break;
+  }
+}
+
+void ConfigValidator::validateOptionsTranslation() const {
```
frankseide: What is not fully clear to me is the different levels. There are classes for
 * parsing cmd-line options
 * representing a set of options
 * interpreting these options in the context of Marian

Which class does what, which class knows what? it looks like ConfigValidator knows about Marian. How about Config? How about Options? Should we qualify classes that know about Marian by including it in the name? E.g. MarianConfig?

src/common/config_validator.cpp
```diff
@@ -0,0 +1,142 @@
...
+      "Decay strategies 'epoch+batches' and 'epoch+stalled' require two "
+      "values specified with --lr-decay-start option");
+  UTIL_THROW_IF2(
+      (strategy == "epoch" || strategy == "batches" || strategy == "stalled")
+          && get<std::vector<size_t>>("lr-decay-start").size() != 1,
+      "Single decay strategies require only one value specified with "
+      "--lr-decay-start option");
+}
+
+void ConfigValidator::validateDevices(cli::mode mode) const {
```
frankseide: Does it make sense to keep a sepaerate validate funciton? Why can the validate function not just call the actual function that interprets this, and let it fail?

src/common/config_validator.h
```diff
@@ -0,0 +1,33 @@
+#pragma once
+
+#include "3rd_party/yaml-cpp/yaml.h"
+#include "common/config_parser.h"
+
+namespace marian {
+
+class ConfigValidator
```
frankseide: comment what this is, and what it is meant to be used for

src/common/logging.cpp
```diff
@@ -56,14 +56,17 @@ void createLoggers(const marian::Config* options) {
   std::vector<std::string> generalLogs;
   std::vector<std::string> validLogs;
 
-  if(options && options->has("log")) {
+  if(options && options->has("log")
+     && !options->get<std::string>("log").empty()) {
```
frankseide: Should the empty string be an allowed "disable" value for this? Why not just not specify it? If this is the only way to find out whether the variable is not specified (because it is declared as defaulting to the empty string), then consider to add a function next to has() that contains the above expression, and expresses this in its name.

src/examples/mnist/mnist_ffnn.cpp
```diff
@@ -24,7 +24,7 @@ const std::vector<std::string> VALID_SET
 using namespace marian;
 
 int main(int argc, char** argv) {
-  auto options = New<Config>(argc, argv, ConfigMode::training, false);
+  auto options = New<Config>(argc, argv, cli::mode::training, false);
```
frankseide: Are we mixing abstraction levels? Should this not be Config::mode::training, rather than cli::? cli seems to be the external parser library that we use, while 'training' clearly refers to a Marian concept, and Config has knowledge of Marian.

src/command/marian_vocab.cpp
```diff
@@ -10,39 +9,18 @@ int main(int argc, char** argv) {
...
+
+  cli.parse(argc, argv);
 
   LOG(info, "Creating vocabulary...");
 
   auto vocab = New<Vocab>();
   InputFileStream corpusStrm(std::cin);
   OutputFileStream vocabStrm(std::cout);
-  vocab->create(corpusStrm, vocabStrm, vm["max-size"].as<size_t>());
+  vocab->create(corpusStrm, vocabStrm, cli.get<size_t>("max-size"));
```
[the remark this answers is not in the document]
emjotde: Yes, that's a short term goal. This PR is only supposed to get do a minimal pain replacement of boost::program_options.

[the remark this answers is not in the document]
snukky: Yes, that's the goal it will be easier now.

src/examples/mnist/mnist_ffnn.cpp
emjotde: Was thinking about that, but the only place where this is actually used is CLI options. Once we start rewriting the whole options/config chaos you addressed earlier, we should be able to do away with this by having hierarchical options.

src/common/config_validator.cpp
emjotde: As said above. This is currently chaotic, but this PR is not meant to fix this yet.

src/common/cli_wrapper.cpp
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  if(has("version")) {
+    std::cerr << PROJECT_VERSION_FULL << std::endl;
+    exit(0);
+  }
+}
+
+bool CLIWrapper::has(const std::string &key) const {
+  return opts_.count(key) > 0 && !opts_.at(key)->empty();
```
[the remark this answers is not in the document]
snukky: This is to preserve the messy logic from config parser using boost::program_options. Will be fixed.

src/common/cli_wrapper.h
emjotde: As above. We are at the moment mimicking old behavior. It's not necessarily better, just getting rid of boost.

src/common/cli_wrapper.h
emjotde: Worth remembering, not in this PR.

src/common/config.cpp
emjotde: Copied over from previous code, just moved from *.h to *.cpp.

src/common/logging.cpp
emjotde: It should be pointing to a file location. I think an empty log should even abort? 
@snukky Is there a setting for that says if set it should be given with a value?

snukky:
@frankseide I agree with most or all your suggestions. This PR is just a first step and it emulates the same behavior we had with boost::program_options as close as possible. I'd like to avoid postponing these changes for much longer and blocking anyone else, but we can make a revert to add more comments.

src/common/any_type.h
```diff
@@ -0,0 +1,132 @@
+#pragma once
+
+#include <utility>
+
+namespace marian {
+
+class any_type {
```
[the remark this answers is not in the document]
emjotde: @snukky this is a copy of std::any or inspired?

src/common/config_parser.h
emjotde: Concerning second paragraph, we will address this in a different phrase. This is a first step.

src/command/marian_vocab.cpp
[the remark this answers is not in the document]
snukky: get() method can be already removed from CLI wrapper

src/common/any_type.h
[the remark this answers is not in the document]
snukky: Based on some code from this codereview: https://codereview.stackexchange.com/questions/48344 with modifications, inspired by boost::any. I added a comment and moved the file to src/3rd_party

src/common/cli_wrapper.cpp
```diff
@@ -0,0 +1,135 @@
+#include "common/cli_wrapper.h"
+#include "common/version.h"
+
+namespace marian {
+namespace cli {
+
+uint16_t guess_terminal_width(uint16_t max_width, uint16_t default_width) {
```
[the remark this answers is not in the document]
snukky: The comment was in .h file. I added TODOs

src/common/cli_wrapper.cpp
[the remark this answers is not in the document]
snukky: I moved handling --version to the config parser and has() method here is not needed anymore

src/common/cli_wrapper.h
```diff
@@ -0,0 +1,275 @@
...
+class CLIFormatter : public CLI::Formatter {
+public:
+  CLIFormatter(size_t columnWidth, size_t screenWidth);
+  virtual std::string make_option_desc(const CLI::Option *) const;
+
+private:
+  size_t screenWidth_{0};
+};
+
+class CLIWrapper {
```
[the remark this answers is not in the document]
snukky: Fixed

src/common/cli_wrapper.h
```diff
@@ -0,0 +1,275 @@
...
+private:
+  // Stores option variables
```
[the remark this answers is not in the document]
snukky: I disagree with line-end comments as these decrease readability even if formatted for 80 or 100 character-length lines. I've updated comments

src/common/cli_wrapper.h
```diff
@@ -0,0 +1,275 @@
...
+  std::map<std::string, Ptr<any_type>> vars_;
+  // Stores option objects
+  std::map<std::string, CLI::Option *> opts_;
+  // Command-line argument parser
+  Ptr<CLI::App> app_;
+  // Stores options as YAML object
+  YAML::Node config_;
+
+  // Name of the default option group
```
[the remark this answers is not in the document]
snukky: Done

src/common/cli_wrapper.h
snukky: Added a comment

src/common/cli_wrapper.h
snukky: Done

src/common/cli_wrapper.h
snukky: Fixed

src/common/cli_wrapper.h
snukky: Fixed

src/command/marian_vocab.cpp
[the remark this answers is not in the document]
frankseide: Then could you do that? And would you mind adding a @TODO comment into the code saying what you just replied here, then Resolve this?

src/common/cli_wrapper.cpp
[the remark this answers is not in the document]
frankseide: "Will be fixed": Please add a @TODO for this, then Resolve this.

src/common/cli_wrapper.h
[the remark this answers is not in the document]
frankseide: I disagree with vertical comments, as they decrease readability since less code fits on a screen.

The thing to do is to horizontally align the comments. Some auto-formatters, like the one we used in CNTK, can do that. Then they are very nicely readable, while getting out of the way of the actual source code. This is the standard we should follow.

src/common/cli_wrapper.h
frankseide: Would you mind adding that note to the source-code comment?

src/common/cli_wrapper.h
frankseide: Could you please add a @TODO to the source code?

src/common/config_parser.h
frankseide: Lets add a @TODO at an appropriate place in teh soruce code, then Resolve this.

src/examples/mnist/mnist_ffnn.cpp
frankseide: We could import the cli::mode into Config::mode? E.g. in Config, using cli::mode or so?

src/command/marian_vocab.cpp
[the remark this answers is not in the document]
snukky: I did do that. So, as we need to create another PR anyway, maybe I will add comments to all suggestions first and then paste links to the code.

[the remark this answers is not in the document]
snukky: And leave everything unresolved for now.

src/common/cli_wrapper.h
snukky: Done

src/command/marian_vocab.cpp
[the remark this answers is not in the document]
emjotde: When you think you are done, do the PR first. We will see what it looks like and decide then.

src/common/cli_wrapper.h
snukky: I created an issue: #305

src/common/cli_wrapper.h
snukky: Added

src/common/cli_wrapper.h
snukky: Done

src/common/cli_wrapper.h
snukky: Done

src/common/cli_wrapper.h
snukky: Done

src/common/config.cpp
snukky: Added TODO

src/common/config.cpp
snukky: Added TODO

src/common/config.cpp
snukky: Added TODO

src/common/config.cpp
snukky: Done

src/common/config.h
snukky: Yes, it is, and we have a regression test for this

src/common/config.h
snukky: The class has been blindly separated into .h and .cpp, as it will be majorly refactorized later. Moreover, this method might be removed as suggested in another comment.

src/common/config_parser.h
snukky: I was pretty sure I've added a test case for that, but I'll check.

src/common/config_parser.cpp
snukky: Yes. `add_nondefault` has been added only to preserve compatibility with the old code and hopefully it will be removed soon. Added TODO

src/common/config_parser.h
snukky: Fixed

src/common/config_validator.cpp
snukky: The class was just extracted from the old ConfigParser to make removing boost::program_options easier.

src/common/config_validator.cpp
snukky: Added TODO

src/common/config_validator.h
snukky: That's an ugly class created from code pieces extracted from the old config parser (slightly re-organized) to make sure that removing boost::program_options doesn't break too much. Will be changed in the upcoming refactorization of CLI.

Added TODO.

src/common/config_validator.cpp
snukky: It doesn't. I moved it here from the old ConfigParser. 
I've made it private and now it's called only from validateOptions().

src/common/logging.cpp
snukky: The extended condition was necessary before I introduced `add_nondefault`. It's not needed anymore, so I reverted the change.

src/examples/mnist/mnist_ffnn.cpp
snukky: I added a TODO related to this resolving one of the previous suggestions.

src/common/config_parser.h
snukky: Added a basic test here: https://github.com/marian-nmt/marian-regression-tests/blob/master/tests/interface/envvars/test_interpolate_envvars.sh

PR #307, opened 2018-09-14
Updates after removing boost::program_options

snukky:
This PR is a continuation of #303 and resolves all mentioned issues/suggestions.

A new PR is created because the previous one has been already merged and can not be re-opened.

All regression tests passed.

### DIFF
--- a/src/command/marian_conv.cpp
+++ b/src/command/marian_conv.cpp
@@ -13,10 +13,9 @@ int main(int argc, char** argv) {
   cli.add<std::string>("--from,-f", "Input model", "model.npz");
   cli.add<std::string>("--to,-t", "Output model", "model.bin");
 
-  cli.parse(argc, argv);
-
-  auto modelFrom = cli.get<std::string>("from");
-  auto modelTo = cli.get<std::string>("to");
+  auto opts = cli.parse(argc, argv);
+  auto modelFrom = opts["from"].as<std::string>();
+  auto modelTo = opts["to"].as<std::string>();
 
   LOG(info, "Outputting {}", modelTo);
 
--- a/src/command/marian_conv.cpp
+++ b/src/command/marian_conv.cpp
@@ -9,13 +9,15 @@ int main(int argc, char** argv) {
 
   createLoggers();
 
-  cli::CLIWrapper cli("Allowed options");
-  cli.add<std::string>("--from,-f", "Input model", "model.npz");
-  cli.add<std::string>("--to,-t", "Output model", "model.bin");
-
-  auto opts = cli.parse(argc, argv);
-  auto modelFrom = opts["from"].as<std::string>();
-  auto modelTo = opts["to"].as<std::string>();
+  auto options = New<Options>();
+  {
+    auto cli = New<cli::CLIWrapper>(options, "Allowed options");
+    cli->add<std::string>("--from,-f", "Input model", "model.npz");
+    cli->add<std::string>("--to,-t", "Output model", "model.bin");
+    cli->parse(argc, argv);
+  }
+  auto modelFrom = options->get<std::string>("from");
+  auto modelTo = options->get<std::string>("to");
 
   LOG(info, "Outputting {}", modelTo);
 
--- a/src/common/cli_helper.h
+++ b/src/common/cli_helper.h
@@ -91,7 +91,11 @@ static void ProcessPaths(
   }
 }
 
-// helper to convert a Yaml node recursively into a string
+// helper to convert a YAML node recursively into a string
+//
+// TODO: create a helper function that converts a YAML node into a string
+// without an emitter; consider extracting YAML-related helper functions to a
+// separate file
 static void OutputYaml(const YAML::Node node, YAML::Emitter& out) {
   std::set<std::string> sorter;
   switch(node.Type()) {
--- a/src/common/cli_helper.h
+++ b/src/common/cli_helper.h
@@ -96,6 +96,7 @@ static void ProcessPaths(
 // TODO: create a helper function that converts a YAML node into a string
 // without an emitter; consider extracting YAML-related helper functions to a
 // separate file
+// TODO: Look for Frank's function that does that.
 static void OutputYaml(const YAML::Node node, YAML::Emitter& out) {
   std::set<std::string> sorter;
   switch(node.Type()) {
--- a/src/common/cli_wrapper.cpp
+++ b/src/common/cli_wrapper.cpp
@@ -1,5 +1,4 @@
 #include "common/cli_wrapper.h"
-#include "common/version.h"
 
 namespace marian {
 namespace cli {
@@ -89,16 +88,7 @@ YAML::Node CLIWrapper::parse(int argc, char** argv) {
     exit(app_->exit(e));
   }
 
-  if(has("version")) {
-    std::cerr << PROJECT_VERSION_FULL << std::endl;
-    exit(0);
-  }
-
   return config_;
-}
-
-bool CLIWrapper::has(const std::string &key) const {
-  return opts_.count(key) > 0 && !opts_.at(key)->empty();
 }
 
 std::string CLIWrapper::failureMessage(const CLI::App *app,
--- a/src/common/cli_wrapper.cpp
+++ b/src/common/cli_wrapper.cpp
@@ -1,4 +1,5 @@
 #include "common/cli_wrapper.h"
+#include "common/options.h"
 
 namespace marian {
 namespace cli {
@@ -56,12 +57,14 @@ std::string CLIFormatter::make_option_desc(const CLI::Option *opt) const {
   return desc;
 }
 
-CLIWrapper::CLIWrapper(const std::string &name,
+CLIWrapper::CLIWrapper(YAML::Node &config,
+                       const std::string &name,
                        size_t columnWidth,
                        size_t screenWidth)
     : app_(std::make_shared<CLI::App>()),
       defaultGroup_(name),
-      currentGroup_(name) {
+      currentGroup_(name),
+      config_(config) {
   // set group name for --help option
   app_->get_help_ptr()->group(defaultGroup_);
   // set custom failure message
@@ -72,6 +75,12 @@ CLIWrapper::CLIWrapper(const std::string &name,
   app_->formatter(fmt);
 }
 
+CLIWrapper::CLIWrapper(Ptr<marian::Options> options,
+                       const std::string &name,
+                       size_t columnWidth,
+                       size_t screenWidth)
+    : CLIWrapper(options->getOptions(), name, columnWidth, screenWidth) {}
+
 CLIWrapper::~CLIWrapper() {}
 
 void CLIWrapper::switchGroup(const std::string &name) {
@@ -81,14 +90,12 @@ void CLIWrapper::switchGroup(const std::string &name) {
     currentGroup_ = name;
 }
 
-YAML::Node CLIWrapper::parse(int argc, char **argv) {
+void CLIWrapper::parse(int argc, char **argv) {
   try {
     app_->parse(argc, argv);
   } catch(const CLI::ParseError &e) {
     exit(app_->exit(e));
   }
-
-  return config_;
 }
 
 std::string CLIWrapper::failureMessage(const CLI::App *app,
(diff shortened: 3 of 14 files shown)


src/common/config_parser.h
```diff
@@ -51,14 +51,14 @@ class ConfigParser {
...
-  // enabled
-  bool has(const std::string& key) const;
-
 private:
   cli::mode mode_;
   YAML::Node config_;
 
+  // Check if the config has a boolean option for the given key and it is
+  // enabled
+  bool hasBool(const std::string& key) const;
```
emjotde: Not a fan of the name "hasBool" we have "has" everywhere else

src/tests/cli_test.cpp
```diff
@@ -46,15 +46,15 @@ int main(int argc, char** argv) {
...
-  w.get<bool>("bool");
+  opts["int"].as<int>();
+  opts["str"].as<std::string>();
+  opts["vec"].as<std::vector<float>>();
+  opts["defvec"].as<std::vector<std::string>>();
+  opts["bool"].as<bool>();
   //w.get<std::string>("long");
-  w.get<std::string>("file");
+  opts["file"].as<std::string>();
   //w.get<color>("enum");
```
emjotde: Also wondering about this part. We usually do not expose the YAML object anywhere directly at least if it can be avoided. How about using "common/options.h" ? That probably not be much of a difference?

src/command/marian_conv.cpp
```diff
@@ -13,10 +13,9 @@ int main(int argc, char** argv) {
   cli.add<std::string>("--from,-f", "Input model", "model.npz");
   cli.add<std::string>("--to,-t", "Output model", "model.bin");
 
-  cli.parse(argc, argv);
-
-  auto modelFrom = cli.get<std::string>("from");
-  auto modelTo = cli.get<std::string>("to");
+  auto opts = cli.parse(argc, argv);
```
frankseide: Similar to a separate comment by Marcin, I hope we can use the same mecahnism everywhere.

Would it be so bad to do something like this:

    auto opts = Options(cli.parse(argc, argv));
    auto modelFrom = opts->get<string>("from");

src/common/cli_helper.h
```diff
@@ -91,7 +91,11 @@ static void ProcessPaths(
   }
 }
 
-// helper to convert a Yaml node recursively into a string
+// helper to convert a YAML node recursively into a string
+//
+// TODO: create a helper function that converts a YAML node into a string
+// without an emitter; consider extracting YAML-related helper functions to a
```
frankseide: I am somewhat sure that I have written that function. Did you accidentally delete it?

src/common/config_parser.h
emjotde: What's the use case here? And what about a more generic has<T> ?

src/common/cli_wrapper.cpp
```diff
@@ -82,21 +81,14 @@ void CLIWrapper::switchGroup(const std::string &name) {
     currentGroup_ = name;
 }
 
-void CLIWrapper::parse(int argc, char** argv) {
+YAML::Node CLIWrapper::parse(int argc, char **argv) {
```
frankseide: Now I am confused. Does `parse()` perform a state change to the CLIWrapper class, or is it a pure function that creates a YAML::Node object?

It seems it does both: It updates config_, and then returns it. I feel this is incorrect design.

frankseide: One possible solution is to make this `void` again, but then add an automatic type cast for this class to YAML::Node. Then you can say

    clip.parse(...);
    YAML::Node opts = cli;

src/common/cli_wrapper.h
```diff
@@ -36,15 +44,30 @@ class CLIFormatter : public CLI::Formatter {
...
+ * getConfig().
+ *
+ * Options are organized in option groups. Each option group has a header that
+ * preceeds all options in the group. The header for the default option group
+ * can be set from the class constructor.
+ */
 class CLIWrapper {
 private:
-  // Stores option variables
+  // Map with option names and variables
```
frankseide: Don't want to be nit-picky, but an even clearer way would be to say what gets mapped to what. Here is how I woudl write it:

    std::map<std::string, Ptr<any_type>> vars_;   // [option name] -> option value

and maybe rename `vars_` to be more specific, e.g. `allVars_` or whatever it contains.

src/common/cli_wrapper.h
```diff
@@ -54,20 +77,25 @@ class CLIWrapper {
...
 
+  // Extract an option name from comma-separated list of command-line arguments,
+  // e.g. 'help' from '--help,-h'
   std::string keyName(const std::string &args) const {
-    return std::get<1>(CLI::detail::get_names(CLI::detail::split_names(args)))
-        .front();
+    // re-use existing functions from CLI11 to keep option names consistent
+    return std::get<1>(CLI::detail::get_names(CLI::detail::split_names(
+                           args)))  // get long names only
+        .front();                   // get first long name
```
frankseide: (Can there be multiple long names? Should there be? If not, maybe add to comment "(there should really only be one)")

src/common/cli_wrapper.h
```diff
@@ -209,37 +275,47 @@ class CLIWrapper {
...
       }
+      // update YAML entry
       config_[key] = vec;
       return (!vec.empty()) && ret;
     };
 
     auto opt = app_->add_option(args, fun, help);
-    opt->type_name(CLI::detail::type_name<T>())->type_size(-1);
+    // set human readable type value: VECTOR and
+    opt->type_name(CLI::detail::type_name<T>());
```
frankseide: Is this a setter? It reads like a getter. If this is our own code, I would prefer `setTypeName`, same for size.

src/common/config_parser.cpp
```diff
@@ -637,19 +641,18 @@ void ConfigParser::parseOptions(int argc, char** argv, bool doValidate) {
     config_ = cli.getConfig();
   }
 
-  if(has("interpolate-env-vars")) {
+  if(hasBool("interpolate-env-vars")) {
     cli::ProcessPaths(config_, cli::InterpolateEnvVars, PATHS);
   }
 
-  if(has("relative-paths") && !has("dump-config")) {
+  if(hasBool("relative-paths") && !hasBool("dump-config")) {
```
frankseide: Is this correct? Are you really testing whether the option is present, or rather whether it is specified to be true? Then it would be more like

    get<bool>("relative-paths", false)

src/common/config_parser.h
frankseide: From the several instances, I think this function is not actually needed; instead, all calls should be rewritten as:

    get<bool>("key", false)

src/tests/cli_test.cpp
```diff
@@ -46,15 +46,15 @@ int main(int argc, char** argv) {
...
 
-  w.parse(argc, argv);
+  auto opts = w.parse(argc, argv);
 
-  w.get<int>("int");
-  w.get<std::string>("str");
-  w.get<std::vector<float>>("vec");
-  w.get<std::vector<std::string>>("defvec");
-  w.get<bool>("bool");
+  opts["int"].as<int>();
```
frankseide: What does this do? Check the types by reading them out and ignoring the ret val, which will throw if the type is incorrect? Then maybe add a comment // validate types

src/common/cli_wrapper.cpp
emjotde: I agree and I am making it void again. 

I'll also add a @TODO here. We need to clean the entire Options/Config/CLI chaos anyway (which is one of the next steps). If we spend to much time here now, will not go forward.

src/common/config_parser.h
emjotde: Changed this to get, but considering to remove this and go back to YAML function. We use YAML for everything else here, the function is a bit obfuscating.

src/tests/cli_test.cpp
emjotde: CLIWrapper is now a proper wrapper which holds a reference to the YAML tree.

src/common/cli_wrapper.cpp
emjotde: Went back to void and removed the get/set functions for the YAML tree, this now can only be passed in the constructor as a reference. Hence proper wrapper.

src/tests/cli_test.cpp
emjotde: New way with a proper wrapper. 

```c++
  auto options = New<Options>();
  {
    auto cli = New<cli::CLIWrapper>(options, "Allowed options");
    cli->add<std::string>("--from,-f", "Input model", "model.npz");
    cli->add<std::string>("--to,-t", "Output model", "model.bin");
    cli->parse(argc, argv);
  }
  auto modelFrom = options->get<std::string>("from");
  auto modelTo = options->get<std::string>("to");
```

src/common/cli_wrapper.h
emjotde: That's the API of the library: https://github.com/CLIUtils/CLI11
Couldn't find if it support multiple long options (why not?), but no reason to dispute this.

src/common/cli_helper.h
emjotde: What's bad about the emitter?

src/common/cli_wrapper.h
emjotde: Done

emjotde:
@frankseide I believe I addressed your comments and a little clean-up seems to go a long way. 
I turned the CLIWrapper into a proper wrapper now that does not hold a copy of the YAML tree anymore but a reference that can only be passed with the constructor. I also remove all accessors to options and the tree itself:

```c++
  auto options = New<Options>();
  {
    auto cli = New<cli::CLIWrapper>(options, "Allowed options");
    cli->add<std::string>("--from,-f", "Input model", "model.npz");
    cli->add<std::string>("--to,-t", "Output model", "model.bin");
    cli->parse(argc, argv);
  }
  auto modelFrom = options->get<std::string>("from");
  auto modelTo = options->get<std::string>("to");
```

I think that's actually quite acceptable for now. Thanks for your help.

src/common/config_parser.cpp
```diff
@@ -634,22 +639,20 @@ void ConfigParser::parseOptions(int argc, char** argv, bool doValidate) {
     auto config = loadConfigFiles(configPaths);
     // combine loaded options with the main YAML config
     cli.overwriteDefault(config);
-    config_ = cli.getConfig();
   }
 
-  if(has("interpolate-env-vars")) {
+  if(get<bool>("interpolate-env-vars")) {
```
frankseide: Don't these need the optional default value? `get<bool>("...", false)`?

src/tests/cli_test.cpp
```diff
@@ -34,31 +34,34 @@ S &operator>>(S &s, color &r) {
...
-  w.switchGroup("My group");
-  w.add<std::vector<std::string>>("--defvec,-d", "help message")->default_val("foo");
-  w.add<bool>("-b,--bool", "help message");
-  w.add<std::string>("--a-very-long-option-name-for-testing-purposes", "A very long text a very long text a very long text a very long text a very long text a very long text");
-  w.switchGroup();
-  w.add<std::string>("-f,--file", "help message")->check(validators::file_exists);
-  //w.add<color>("-e,--enum", "help message for enum");
+  auto options = New<Options>();
+  {
+    auto w = New<CLIWrapper>(options);
```
frankseide: This is so much better!!

src/tests/cli_test.cpp
```diff
@@ -34,31 +34,34 @@ S &operator>>(S &s, color &r) {
...
-  w.parse(argc, argv);
+    w->parse(argc, argv);
+  }
 
-  w.get<int>("int");
-  w.get<std::string>("str");
-  w.get<std::vector<float>>("vec");
-  w.get<std::vector<std::string>>("defvec");
-  w.get<bool>("bool");
+  options->get<int>("int");
```
frankseide: Comment what's this for?

src/common/config_parser.cpp
emjotde: This has a default value of "false", since we are inside the config parser which sees the option definitions, this is safe.

src/tests/cli_test.cpp
emjotde: I think this can actually be removed, @snukky?

PR #308, opened 2018-09-14
Enable development on Windows with GPU support

cedrou:
This PR adds the necessary files to enable the development on Windows with CUDA/CuDNN and the Intel MKL.

3 alternatives are proposed for the IDE:
- Visual Studio 2017+ with built-in CMake support  
  Enable directory view, intellisense, build with CMake inside VS, debug, …

- Visual Studio with the solution file generated by CMake  
  This is the standard method with CMake on Windows with all the disadvantages: project file with all compiled files in the same folder, ALL_BUILD target that rebuilds all every time, bad debugging experience, …

- Any editor  
   Use your favorite editor and build with a script that have the right flags for CMake

Please see the [README file](https://github.com/cedrou/marian-dev/blob/build_on_win/vs/README.md) for more information, especially on the changes done in the source files to make it compile on Windows.

I initially put the files in the `vs` subdirectoy introduced by @frankseide, but I subsequently found the folder `contrib/other-builds` that already contains project files for Eclipse: maybe I should move there.


src/common/file_stream.h
```diff
@@ -15,82 +15,119 @@
...
+#ifdef _MSC_VER
+    char* name = _tempnam(base.c_str(), "marian.");
+    ABORT_IF(name == NULL,
+			"Error while making a temporary based on '{}'",
+      base);
+
+    int oflag = _O_RDWR | _O_CREAT | _O_EXCL;
+    if (unlink_) oflag |= _O_TEMPORARY;
+
+		int ret = _open(name, oflag, _S_IREAD | _S_IWRITE);
```
frankseide: I dislike the _ prefix, and that VS forces them upon us. These are the standard Linux functions, there is no value in renaming them on Windows. In my view, we should insist to use the Linux names, and disable the warning explicitly. My version already does that. This is one of the few cases where disabling a warning is more than justified. (The other warning about the so-called "safe" CRT functions like `fopen_s()` which we should not use here.)

Once you change that, I also think you can reconcile at least some of these diverging code paths (e.g. I saw a `_close()` call below that was identical to a `close()` in the Linux version)

src/data/corpus_sqlite.cpp
```diff
@@ -82,7 +82,7 @@ void CorpusSQLite::fillSQLite() {
       for(size_t i = 0; i < files_.size(); ++i) {
         cont = cont && utils::GetLine((std::istream&)*files_[i], line);
         if(cont)
-          ps.bind(i + 2, line);
+          ps.bind((int)(i + 2), line);
```
frankseide: I would instead change the type of `i` to `int` in the for loop.

src/data/corpus_sqlite.cpp
```diff
@@ -114,7 +114,7 @@ SentenceTuple CorpusSQLite::next() {
     SentenceTuple tup(curId);
 
     for(size_t i = 0; i < files_.size(); ++i) {
-      auto line = select_->getColumn(i + 1);
+      auto line = select_->getColumn((int)(i + 1));
```
frankseide: `int i`

src/models/transformer_factory.h
```diff
@@ -13,6 +13,6 @@ Ptr<EncoderBase> NewEncoderTransformer(Ptr<Options> options);
 Ptr<DecoderBase> NewDecoderTransformer(Ptr<Options> options);
 }  // namespace marian
 
-#ifndef _WIN32  // TODO: remove this once I updated the Linux-side makefile
+//#ifndef _WIN32  // TODO: remove this once I updated the Linux-side makefile
```
frankseide: It's OK to actually remove it :)

src/rescorer/score_collector.cpp
```diff
@@ -91,10 +91,10 @@ void ScoreCollectorNBest::Write(long id,
                "Entry {} < {} already read but not in buffer",
                id,
                lastRead_);
-      std::string line;
-      while(lastRead_ < id && utils::GetLine((std::istream&)*file_, line)) {
+      std::string line1;
```
frankseide: (I presume this is due to shadowing. Consider to simply not dedeclare this variable, and just use the previous `line`, as long as the code remains correct. I did a similar change elsewhere.)

src/tensors/gpu/backend.h
```diff
@@ -18,14 +18,14 @@
 namespace marian {
 namespace gpu {
 
-class Backend : public marian::Backend {
+class gpuBackend : public marian::Backend {
```
frankseide: Please start class names with upper-case letters. GpuBackend? GPUBackend?

src/tensors/gpu/cuda_helpers.h
```diff
@@ -4,7 +4,7 @@
 #include "3rd_party/exception.h"
 #include "common/logging.h"
 
-const float CUDA_FLT_MAX = 1.70141e+38;
+__constant__ float CUDA_FLT_MAX = 1.70141e+38;
```
frankseide: What is this?

src/tensors/gpu/cudnn_wrappers.cu
```diff
@@ -170,17 +170,24 @@ void ConvolutionWrapper::setConvDescriptor(int hPad,
                                            int wStride) {
   CUDNN_CALL(cudnnCreateConvolutionDescriptor(&convDesc_));
 
+#if CUDNN_MAJOR > 5
```
frankseide: If this only differs in one parameter, consider doing the #if only for that section, as it will be easier for the reader to see where it differs. Seems that's what the old version was?

vs/CheckDeps.bat
```diff
@@ -0,0 +1,288 @@
+@echo off
```
frankseide: Do you mind adding a top-level comment to this file on what it does, and in what context it is meant to be used?

vs/CreateVSProjects.bat
```diff
@@ -0,0 +1,57 @@
+@echo off
```
frankseide: Same here, please add a top-level comment

vs/README.md
```diff
@@ -0,0 +1,200 @@
+# Build Marian on Windows with GPU support
```
frankseide: Build -> Building? How to build?

vs/README.md
```diff
@@ -0,0 +1,200 @@
...
+## Build the project
+
+There are 3 alternatives to build the project:
+1. Use Visual Studio 2017 built-in support for CMake
+2. Create a Solution file for Visual Studio
+3. Use a script (MSBuild)
+
+### 1. Use VS2017 with built-in support for CMake
+
+VS2017 now allows to develop projects built with CMake without the need to generate VS projects and solutions. For more information, please read [this article](https://blogs.msdn.microsoft.com/vcblog/2016/10/05/cmake-support-in-visual-studio/) from the Visual C++ Team.
```
frankseide: Nice!

emjotde:
This is deserving of a blog post on the website :)

src/3rd_party/CLI/ConfigFwd.hpp
```diff
@@ -117,7 +117,7 @@ class ConfigINI : public Config {
         std::vector<ConfigItem> output;
 
         while(getline(input, line)) {
-            std::vector<std::string> items;
+            std::vector<std::string> sitems;
```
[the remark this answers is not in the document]
cedrou: Yes, `items` is already defined in the parent class and is hidden by this declaration. I added the prefix `s` for `string`, because … I had to find something … I could have renamed it `items1` or so.

src/3rd_party/cnpy/cnpy.h
```diff
@@ -350,15 +350,19 @@ namespace cnpy {
...
+#endif
+		bad = bad || (rename(tmpname.c_str(), zipname.c_str()) == -1);
+
+		if (bad)
+		{
+			unlink(tmpname.c_str());
+			throw std::runtime_error("npz_save: error saving to file: " + zipname);
+		}
+
+	}
```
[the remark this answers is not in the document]
cedrou: OK. Got it. The VS setup on my laptop is wrongly configured with Tabs. Thanks for that !

src/common/cli_wrapper.cpp
```diff
@@ -4,7 +4,8 @@
 namespace marian {
 namespace cli {
 
-uint16_t guess_terminal_width(uint16_t max_width, uint16_t default_width) {
+/*
+static uint16_t guess_terminal_width(uint16_t max_width, uint16_t default_width) {
```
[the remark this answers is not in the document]
cedrou: This caused a warning "unused local function was deleted". My first idea was to remove it, but in the code below there is a TODO to compute the terminal size with this function. So I just commented it for now.

src/common/file_stream.h
cedrou: I also disabled these 2 warnings. I'll take a look at all those changes.

cedrou:
Sure, I'll do that

src/data/corpus_sqlite.cpp
cedrou: I had the choice between this fix (only one cast) or the fix you propose which implies 3 casts below in lines 119, 121 and 124.

src/3rd_party/simple-websocket-server/server_ws.hpp
```diff
@@ -334,8 +334,8 @@ namespace SimpleWeb {
 
         auto send_stream = std::make_shared<SendStream>();
 
-        send_stream->put(status >> 8);
-        send_stream->put(status % 256);
+        send_stream->put((unsigned char)(status >> 8));
```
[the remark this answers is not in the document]
cedrou: This is probably in code for the CUDA specific.

cedrou:
I have addressed all the comments: thanks for the review. The compilation is now warning free, expect for 2 that look tricky to fix:

- the first one is a member shadowing in `src/3rd_party/simple-websocket-server/server_ws.hpp`, line 422 : "declaration of endpoint hides class member". The point is that the class member is used but I cannot find where it is set. I don't really understand what is done here, and I don't want to add a bug...

- the second is for a function that is declared in a header, but not used every time the header is used. It should be declared in the header and implemented only once in a c++ file, but I don't know where...

Concerning the format configuration, the last updates of VS2017 now support .editorconfig and .clang-format files.

cedrou:
The compiler was confused by a name conflict with these 3 Backend classes, even if they are in different namespace. I haven't dig much more on this issue, and I took the fast way by just renaming them. I can look at it more closely if you're worried about that.

emjotde:
Very weird. Maybe don't waste time on that and will later try to solve this.

src/tensors/gpu/algorithm.h
```diff
@@ -5,12 +5,12 @@
 namespace marian {
 namespace gpu {
 template <typename T>
-void copy(Ptr<Backend> backend, const T* begin, const T* end, T* dest);
+void copy(Ptr<marian::Backend> backend, const T* begin, const T* end, T* dest);
```
cedrou: This commit allowed me to rollback some previous changes. The `cpu::Backend` and `gpu::Backend` classes were renamed to `cpu::CpuBackend` and `gpu::GpuBackend` because of a compile error related to this function. The compiler was confused between the `Backend` and the `gpu::Backend` classes when instantiating this function. This change alone fix the name ambiguity for the compiler and I renamed the classes back to their original names.

src/common/cli_wrapper.cpp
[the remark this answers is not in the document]
frankseide: (Un-resolving only so that this does not get missed. Please just Resolve it once you saw this, no need for action.)

In such cases, I found it is sometimes easier to change the function definition to 'static inline'. I use that trick for static functions headers. Probably not here though.

src/tensors/gpu/cuda_helpers.h
frankseide: Would you mind adding a note as a very brief line-end comment? E.g. // fixes a missing constant in CUDA-side headers

vs/README.md
frankseide: No, just a grammar fix of this one comment line. Grammatically, it should be either "Building Marian..." or "How to build Marian..."

vs/README.md
```diff
@@ -0,0 +1,200 @@
...
+
+3. __Fix Warning: D9002: ignoring unknown option '-m64'__  
+   This one is related to a compiler flag added while finding the package MKL that does not exists for MS compiler. 
+
+4. __Fix marian::Backend, marian::cpu::Backend and marian::gpu::Backend conflicts__  
+   There were name conflicts between the 3 `Backend` classes that confused the compiler:
+   
+   >  template instantiation resulted in unexpected function type of "type" (the meaning of a name may have changed since the template declaration -- the type of the template is "type").
+
+   I renamed the CPU and GPU as `cpuBackend` and `gpuBackend`.
```
frankseide: Is this still accurate, or was this rolled back? Would you mind quickly going over these once again to check whether some of these changes were undone or changed further since you wrote this? Then please close this Conversation.

vs/README.md
cedrou: OK. Get it. Thanks

emjotde:
@cedrou great thanks!

PR #312, opened 2018-09-21
Fix logging version; support .yaml extension

snukky:
Changes:
* Fixing the message "Model created with Marian false"
* Logging Marian version for newly started training
* Adding .yaml as a supported vocab extension

### DIFF
--- a/src/common/config.cpp
+++ b/src/common/config.cpp
@@ -2,6 +2,7 @@
 #include "common/file_stream.h"
 #include "common/logging.h"
 #include "common/utils.h"
+#include "common/version.h"
 
 #include <algorithm>
 #include <boost/algorithm/string.hpp>
@@ -28,11 +29,18 @@ void Config::initialize(int argc, char** argv, cli::mode mode, bool validate) {
 
   createLoggers(this);
 
+  // set random seed
   if(get<size_t>("seed") == 0)
     seed = (size_t)time(0);
   else
     seed = get<size_t>("seed");
 
+  // set Marian version for newly started training, can be overwritten by a
+  // version from loaded model parameters
+  if(mode == cli::mode::training)
+    config_["version"] = PROJECT_VERSION_FULL;
+
+  // load model parameters
   if(mode != cli::mode::translation) {
     if(filesystem::exists(get<std::string>("model"))
        && !get<bool>("no-reload")) {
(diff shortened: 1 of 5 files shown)


src/common/config_parser.cpp
```diff
@@ -44,8 +44,7 @@ void ConfigParser::addOptionsGeneral(cli::CLIWrapper& cli) {
 
   // clang-format off
   cli.add<bool>("--version",
-     "Print version number and exit",
-     false);
+     "Print version number and exit");
```
emjotde: Is this false by default?

src/common/config_parser.cpp
```diff
@@ -635,6 +634,8 @@ void ConfigParser::parseOptions(int argc, char** argv, bool doValidate) {
   if(get<bool>("version")) {
     std::cerr << PROJECT_VERSION_FULL << std::endl;
     exit(0);
+  } else {
+    config_["version"] = PROJECT_VERSION_FULL;
```
emjotde: Maybe add a comment. It took me a moment to understand what this does.

src/common/config_parser.cpp
snukky: I removed the definition of --version from ConfigParser. Now it's automatically added like --help is and not present in `config`

src/common/config_parser.cpp
snukky: Now, `config["version"]` is a version number and not related to the --version flag

src/common/config.cpp
```diff
@@ -63,10 +58,31 @@ void Config::initialize(int argc, char** argv, cli::mode mode, bool validate) {
 
   log();
 
-  if(has("version"))
+  // Log version of Marian that has been used to create the model.
+  //
+  // Key "version" is present only if loaded from model parameters and is not
+  // related to --version flag
+  if(has("version")) {
+    auto version = get("version").as<std::string>();
```
emjotde: Better use `get<std::string>("version")`

src/common/config.cpp
```diff
@@ -63,10 +58,31 @@ void Config::initialize(int argc, char** argv, cli::mode mode, bool validate) {
...
+  }
+  // If this is a newly started training
+  else if(mode == cli::mode::training) {
     LOG(info,
-        "[config] Model created with Marian {}",
-        get("version").as<std::string>());
+        "[config] The model is being created with Marian {}",
+        PROJECT_VERSION_FULL);
+    config_["version"] = PROJECT_VERSION_FULL;
+  }
```
emjotde: I think we should log the Marian version every time, not only during training?

src/common/config.cpp
snukky: Fixed

src/common/config.cpp
snukky: It's also logged for translation with a Marian model. We have a regression test for this.

I see these cases: 
* (A) If a loaded model has been trained with Marian, the version will be loaded from the model configuration file and logged for training/translation/scoring.
* If there is no model configuration file, it means that it's either (B) a newly started training (message "Model is being created with...")
* or (C) translation/scoring with a model trained with Nematus or a very old Marian (no message).

"[config] version: xyz" will be displayed only if a version is loaded from the model configuration file.

PR #316, opened 2018-09-26
Remove more boost reference and reorganize code for easier future removal

emjotde:
* Thin wrapper around cpu_timer and auto_cpu_timer
* Remove references to <boost/thread.hpp>
* Attempt to remove boost from filestream.h, keeping original file for now.


src/command/marian_scorer.cpp
```diff
@@ -2,13 +2,14 @@
 
 #include "models/model_task.h"
 #include "rescorer/rescorer.h"
+#include "common/timer.h"
```
snukky: Inspired by missing autoformatting: what do you think about running a small subset of rules from the code autoformatter in a regular basis, for example, weekly? Only some mechanical and  usually very safe rules, such as trailing whitespaces, needless new lines, tabs instead spaces, unsorted includes, spaces after keywords, etc. Jenkins could open pull requests with those for us.

src/common/file_stream.h.bac
```diff
@@ -0,0 +1,201 @@
...
+  }
+
+  template <typename T>
+  size_t write(const T* ptr, size_t num = 1) {
+    ostream_->write((char*)ptr, num * sizeof(T));
+    return num * sizeof(T);
+  }
+
+  std::string path() { return file_.string(); }
+};
```
snukky: Do we need to add the unused file into master? I think we should keep it in this branch only. Maybe the same with zstr.

src/graph/node_operators_binary.h
```diff
@@ -410,18 +410,15 @@ class DotBatchedNodeOp : public NaryNodeOp {
...
-  template <typename... Args>
-  ScalarProductNodeOp(Expr a, Expr b, Args... args)
-      : NaryNodeOp({a, b}, newShape(a, b, args...)) {}
+  ScalarProductNodeOp(Expr a, Expr b, int axis)
+      : NaryNodeOp({a, b}, newShape(a, b, axis)) {}
 
-  template <typename... Args>
-  Shape newShape(Expr a, Expr b, Args... args) {
-    int ax = keywords::Get(keywords::axis, -1, args...);
+  Shape newShape(Expr a, Expr b, int axis) {
```
snukky: An empty line.

src/graph/node_operators_unary.h
```diff
@@ -455,10 +454,9 @@ struct SumNodeOp : public UnaryNodeOp {
     return {NodeOp(Add(_1, child(0)->grad(), adj_))};
   }
 
-  template <class... Args>
-  Shape newShape(Expr a, Args... args) {
+  Shape newShape(Expr a, int axis) {
     Shape shape = a->shape();
-    ax_ = shape.axis(keywords::Get(keywords::axis, -1, args...));
+    ax_ = shape.axis(axis);
```
snukky: I think `ax_` can be renamed to `axis_` for consistency.

src/command/marian_scorer.cpp
emjotde: Maybe put this question into an issue?

emjotde:
@frankseide do you want to take a look at this too? This is a larger PR, but fairly trivial.

Issue #240, opened 2018-05-22
Training goes astray in the second epoch (with --right-left)


bhaddow:
Training with approx 60M sentences, using the --right-left flag, I noticed that development scores collapsed just after the start of the second epoch. This happened on 4 identical runs, but did not happen on other (smaller) data sets. I am not using --sqlite

End of first epoch
Ep. 1 : Up. 350000 : Sen. 61128422 : Cost 42.18 : Time 677.69s : 2063.19 words/s

Start of second epoch:
Ep. 2 : Up. 351000 : Sen. 82422 : Cost 84.35 : Time 735.34s : 1399.12 words/s

This was observed on a private Edinburgh branch, but I do not think that the corpus handling code was touched.

emjotde:
Hi, don't think this happened because of right2left. This is applied in the batch level and does not touch actual data. 

Did you maybe run the training processes on the same machine at the same time? Can l see your config?

emjotde:
Maybe we can try to dump the result of the shuffled data somewhere to see what happened.

bhaddow:
Maybe related, but shuffling without sqlite could be swallowing errors. The evidence for this is that I had some runs where the shuffling was happening in /tmp, which caused the root filesystem to fill up making the node (and the cluster admin) unhappy but Marian kept going. With --sqlite Marian crashed when it filled up the filesystem.

emjotde:
Yes, I have similar suspicions.

XapaJIaMnu:
I experienced the same issue (always happens with right-left, but only after second epoch on left-right and seems to be dependent on the random seed). I have a few systems on csd3 on which i can reliably reproduce the issue with left-right.

I would imagine that if disk space was issue, we would experience it every time, not dependent on the random seed, mm?

emjotde:
Oh? OK. I will stare hard at the R2L option. Weird. I have used it all the time for my system.

snukky:
I've got training settings and data from @XapaJIaMnu and I'm checking if I can reproduce the bug.

@bhaddow did you use symbolic links for the training data files in this experiment?

bhaddow:
Symbolic links - Yes I would have used them.

On 28 May 2018 12:46:22 CEST, Roman Grundkiewicz <notifications@github.com> wrote:
>I've got training settings and data from @XapaJIaMnu and I'm checking
>if I can reproduce the bug.
>
>@bhaddow did you use symbolic links for the training data files in this
>experiment?
>
>-- 
[...]

PR #318, opened 2018-09-28
Abort if an error occured during an input operation on a stream

snukky:
Hopefully, a solution for #240. Not sure if we need to/want to add an explicit message related to shuffling in /tmp for the input operation here: https://github.com/marian-nmt/marian-dev/blob/d82aff455187f1ae741dc8d59e7908515cf5c086/src/data/corpus.cpp#L119

### DIFF
--- a/src/common/file_stream.h
+++ b/src/common/file_stream.h
@@ -205,12 +205,16 @@ public:
   template <typename T>
   friend OutputFileStream& operator<<(OutputFileStream& stream, const T& t) {
     stream.ostream_ << t;
+    ABORT_IF(!stream.ostream_,
+             "Exception writing to file '{}'",
+             stream.file_.string());
     return stream;
   }
 
   template <typename T>
   size_t write(const T* ptr, size_t num = 1) {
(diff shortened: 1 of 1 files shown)


emjotde:
Didn't we identify an issue with exceptions not being enabled?

snukky:
An issue is that we're not checking error codes. As far as I understand, enabling and catching exceptions is another way to do the same thing. The IO library was written before exceptions had been added.

emjotde:
I mentioned this in #240:

http://www.cplusplus.com/reference/ios/ios/exceptions/

emjotde:
I would hold on with this change. There are a couple of issues here. 
1) We should merge my larger PR first. 
2) In the branch rm-boost-3 I have fix for operator<< not working properly with our file streams. Turns it was an issue with std::endl. With this we can get rid of all the the conversions to (std::istream&/std::ostream&).
3) This gives us finally a single place and guarantee that our << and >> are actually being used and not the STL versions. 
4) Only then it makes sense to try to catch these errors as otherwise we will probably miss most of them anyway.

Issue #315, opened 2018-09-26
(Re)move model options from scorer/decoder?


emjotde:
Should we put model options at the end of the options list for `--help` with the scorer and decoder? We should also think if we want to remove them. They are only used for compatibility with Nematus, and maybe we should enforce using a script for that to add the model options to the npz file.

snukky:
I vote for removing those options and forcing using a script.
A note for myself: remove the options, update regression tests, update information on the website, update examples.

emjotde:
Maybe let's just move them to the bottom.

frankseide:
Marcin's suggestion is based on me whining. During development, it sometimes happens that parameters written to the model file are not consistent with parameters used. E.g. initial training uses a hard-coded parameter, where the proper parameter is only added once training looks like it is doing something. Another example is that it is very easy to forget to include new parameters in the set of params saved to the model file. Now I can fix this by passing those missing parameters when running decoding.

The problem with an update script is that it breaks processes in more complex ways than just passing a few extra command-line args. One now must have an additional way to manage additional patched versions of a file. The alternative, updating a model-file in-place, is no good either since it changes its date stamp. With `--overwrite`, the date stamp is an important indicator of which model generation a file represents.

The conclusion of our discussion was to not remove the parameters *for now*. Instead, when some day in the future we move to a way to define parameters in a module-specific way that allows to define a parameter quickly and safely (that is, in a single place where they are used instead of today's 3 in different source files), we can then remove the parameters.

emjotde:
Yupp. We will get back to that when we do a rewrite of the option hierarchy. I am keeping this issue open.

snukky:
With boost-based CLI (or actually before loading config files has been rewritten), it was possible to pass any option you wanted via a config file. That was useful during development as you didn't have to even define a command-line argument. As a possible solution we could consider bringing this "feature" back, of course not as a default behaviour.

BTW. We should at least warn if an unrecognised option occurs in a loaded config file.

emjotde:
I think I would still leave things as they are now, but implement the warning you mentioned in the "BTW".

PR #320, opened 2018-09-29
Warn if an unrecognized option is found in a config file

snukky:
The simplest solution for the missing warning mentioned in #315 that prints the message to the standard error stream. The warning message was there but not displayed because loggers cannot be initialized at that time.


emjotde:
If it cannot be logged correctly, maybe it should be an abort?

snukky:
Maybe, I have no opinion. Warning will allow to provide any config file you want, which might be useful during development.

emjotde:
Lack of logging doesn't allow to track this down. Will an early abort be logged?

snukky:
Yes, ABORT_IF will work here.

emjotde:
Uhm, does the message from ABORT_IF get logged then? If yes, why would the previous message not get logged?

emjotde:
Let's do ABORT then. I think it's logical to have that in the same vein as unrecognized options on the CLI also abort.

snukky:
Replaced by #363

PR #321, opened 2018-09-29
Make use of InputFileStream and OutputFileStream consistent

emjotde:
Make use of InputFileStream and OutputFileStream consistent

* Moved to namespace `io`
* Disallow other file and stream IO besides Input/OutputFileStream
* Add error handling for stream, now aborts instead of failing silently
* Add a number of missing functions like `bad()` and `widen()`
* Fix `operator<<` for `OutputFileStream` with `std::endl`

### DIFF
--- a/src/common/file_stream.h
+++ b/src/common/file_stream.h
@@ -18,8 +18,6 @@
 #include <iostream>
 #include "common/logging.h"
 
-namespace io = boost::iostreams;
-
 #ifdef _MSC_VER
 
 #include <fcntl.h>
@@ -27,6 +25,9 @@ namespace io = boost::iostreams;
 #include <stdlib.h>
 
 #endif
+
+namespace marian {
+namespace io {
 
 class TemporaryFile {
 private:
@@ -140,12 +141,12 @@ public:
         !marian::filesystem::exists(file_), "File '{}' does not exist", file);
 
     if(file_.extension() == marian::filesystem::Path(std::string(".gz")))
-      istream_.push(io::gzip_decompressor());
+      istream_.push(boost::iostreams::gzip_decompressor());
     istream_.push(ifstream_);
   }
 
   InputFileStream(TemporaryFile& tempfile)
-      : fds_(tempfile.getFileDescriptor(), io::never_close_handle) {
+      : fds_(tempfile.getFileDescriptor(), boost::iostreams::never_close_handle) {
     lseek(tempfile.getFileDescriptor(), 0, SEEK_SET);
     istream_.push(fds_, 1024);
   }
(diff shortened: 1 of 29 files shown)


src/common/file_stream.h
```diff
@@ -205,12 +250,27 @@ class OutputFileStream {
   template <typename T>
   friend OutputFileStream& operator<<(OutputFileStream& stream, const T& t) {
     stream.ostream_ << t;
+    ABORT_IF(!stream.ostream_,
```
snukky: Here we are checking for both a badbit and failbit as `!` is a synonym of `fail()`. Above (in `getline` and read operations) only a badbit is checked with `bad()`. This should be unified. Or is this some difference between Input and OutputFileStream?

src/examples/iris/helper.cpp
```diff
@@ -26,10 +26,10 @@ void readIrisData(const std::string fileName,
   }
   std::string line;
   std::string value;
-  while(marian::utils::getline(in, line)) {
+  while(marian::io::getline(in, line)) {
```
snukky: This will not compile as `in` is `std::ifstream`. I think using `std::getline` should be safe here as the Iris data file is provided and it doesn't contain CRs. I'll fix this.

snukky:
Fixed compilation with examples and tests.

src/examples/iris/helper.cpp
```diff
@@ -26,10 +26,10 @@ void readIrisData(const std::string fileName,
   }
   std::string line;
   std::string value;
-  while(marian::io::getline(in, line)) {
+  while(std::getline(in, line)) {
     std::stringstream ss(line);
     int i = 0;
-    while(marian::io::getline(ss, value, ',')) {
+    while(std::getline(ss, value, ',')) {
```
emjotde: Wrong approach. Please use InputFileStream and io::getline.

src/examples/iris/helper.cpp
```diff
@@ -26,10 +26,10 @@ void readIrisData(const std::string fileName,
   }
   std::string line;
   std::string value;
-  while(marian::io::getline(in, line)) {
+  while(std::getline(in, line)) {
```
emjotde: Wrong approach, please rather use InputFileStream and io::getline. I would like to discourage using std::ifstream and std::getline, so we can intercept this.

src/examples/iris/helper.cpp
emjotde: This should be done rather with split?

src/common/file_stream.h
emjotde: I think you cannot do that for InputFileStream in the same way. Consider:
```
while(io::getline(stream, line)) {
   …
}
```

This is also a Boolean check on steam and stops when EOF is reached. I think there is no equivalent for OutputFileStream.

snukky: My question was whether we should use `fail()` instead of `bad()`, not about boolean check.

emjotde: Where in the code is that?

emjotde: OK found it. Rather comment in the code where the error is, not where it is not :)
It seems that checking for fail() checks both, failbit and badbit, so we should indeed use that.

snukky: Yes, `fail()` does check for both. OK, I'll change that.

src/examples/iris/helper.cpp
snukky: Why not just leaving `std::getline` here?

emjotde: It's not a line. Seems wrong semantically.

kpu: I think it's fine using generalized getline to scan for a character.  The real evil is std::stringstream when one should be using a `string_view`.

src/common/file_stream.h
emjotde: Reaching EOF seems to set failbit. So this cannot be used for std::getline, I guess?

PR #336, opened 2018-10-09
Fix compilation of multi-node training

snukky:
This removes leftover scaled learning rate that breaks compilation of multi-node training.

### DIFF
--- a/src/training/graph_group_multinode.cpp
+++ b/src/training/graph_group_multinode.cpp
@@ -288,14 +288,8 @@ void MultiNodeGraphGroup::launchServerThread() {
               cudaStreamSynchronize(0);
 
               // Run optimizer on GPU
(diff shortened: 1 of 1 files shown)


emjotde:
We are not pulling this, as we have some changes internally where we did that already. This would only cause more merge conflicts.

kpu:
Is the plan to release Microsoft's internal changes?  
I'm asking whether Alham / Roman should be working on multi-node.  
It's really Microsoft's problem to do merges if you've chosen to have an internal fork that's opaque to external people who don't know what you have.  The public repo should continue to improve.

emjotde:
Yes. Otherwise I would not complain. We have a working version internally, the only reason this is not public is not secrecy, but convenience. 

What makes you think the public repo is not improving? :)

emjotde:
And currently, do not work on multi-node.

Issue #333, opened 2018-10-09
Missing tensors decoder_c_tt in Amun-type models


Proyag:
When I train a model with `--type amun` and try to decode it with Amun, I get
```
[Tue Oct 9 13:26:45 2018] (I) Loading scorers...
Error: Tensor not found:decoder_c_tt
Aborted (core dumped)
```

Comparing the old and new models, I see the following two keys are missing in the new models: `{'ff_logit_W', 'decoder_c_tt'}`.

I wasn't sure whether to post this here or in the Amun repo, but I think this is caused by changes in Marian and not Amun, since models with the exact same training parameters trained with older versions of Marian still work with Amun.

snukky:
How old version of Marian does work here? Could you provide the commit?

Proyag:
Models trained with https://github.com/marian-nmt/marian-dev/commit/91c19625e02f4598fb73f0a7b767b3f177487cf9 (v1.5.0) work, but not https://github.com/marian-nmt/marian-dev/commit/8b799b2d2889132d3607cde2d2e54ed8f81b2546 (v1.6.0). I just checked that marian-decoder still works for the new model.

snukky:
`decoder_c_tt` seems to be a dummy matrix that has gone with this commit: https://github.com/marian-nmt/marian-dev/commit/a2fbe9#diff-e1c381307655a2f7c4b6ff06f2c777deL105

You may fix your model by add it back using the script: https://github.com/marian-nmt/marian-dev/blob/master/scripts/contrib/inject_ctt.py

We will need to fix Marian or Amun.

emjotde:
Let's add c_tt back to marian. It was meant as a dummy from the start.

PR #341, opened 2018-10-12
Add missing 'decoder_c_tt' to amun and nematus models

snukky:
This addresses #333 and https://github.com/marian-nmt/amun/issues/6 restoring the incidentally removed 'decoder_c_tt': https://github.com/marian-nmt/marian-dev/commit/a2fbe9#diff-e1c381307655a2f7c4b6ff06f2c777deL105


emjotde:
Comment for myself. We will try to approach this differently, but loading io::Item directly and passing them to the graph. This requires a modification of the public interface of expression_graph, but only of those functions that were actually custom added for this particular problem. So we will only change one approach for another that should not  affect any other situations.

src/models/amun.h
```diff
@@ -89,7 +89,14 @@ class Amun : public EncoderDecoder {
...
 
-    graph->load(name, nameMap, true, {"decoder_c_tt"});
+    auto items = io::loadItems(name);
+    for(auto it = items.begin(); it != items.end(); ++it) {
+      if(it->name == "decoder_c_tt") {
+        items.erase(it);
+        break;
+      }
+    }
+    graph->load(name, items, nameMap, true);
```
emjotde: Let's add a few comments what's happening here and in related places.

src/graph/expression_graph.h
```diff
@@ -475,27 +475,31 @@ class ExpressionGraph : public std::enable_shared_from_this<ExpressionGraph> {
 
 public:
   void load(const std::string& name,
+            const std::vector<io::Item>& items,
             const std::map<std::string, std::string>& nameMap,
```
emjotde: Do we use the nameMap now anywhere? If not, let's remove it.

PR #351, opened 2018-10-21
Replace Config where ever possible with an Option object

emjotde:
Opening this PR on behalf of @snukky as I want to have a platform for discussion here. This is not ready to be merged yet.


src/common/config.cpp
```diff
@@ -139,78 +142,14 @@ void Config::log() {
...
-      deviceNos.resize(numDevices);
-    }
-    // form the final vector
-    for (auto d : deviceNos)
-      devices.push_back({ d, DeviceType::gpu });
-  }
-#ifdef MPI_FOUND
-  for (auto d : devices)
-    LOG(info, "[MPI rank {} out of {}]: {}[{}]", myMPIRank, numMPIProcesses, d.type == DeviceType::cpu ? "CPU" : "GPU", d.no);
-#endif
```
emjotde: Why this move?

src/training/validator.h
```diff
@@ -58,31 +57,33 @@ class ValidatorBase : public TrainingObserver {
...
-  Validator(std::vector<Ptr<Vocab>> vocabs,
-            Ptr<Config> options,
-            bool lowerIsBetter = true)
-      : ValidatorBase(lowerIsBetter), vocabs_(vocabs), options_(options) {}
+  Validator(std::vector<Ptr<Vocab>> vocabs, Ptr<Options> options, bool lowerIsBetter = true)
+      : ValidatorBase(lowerIsBetter),
+        vocabs_(vocabs),
+        // options_ is a copy of global options, so it can be safely modified within the class
+        options_(New<Options>(*options)) {
+    // set options common for all validators
```
emjotde: This is fixing the issue is asked about above? Is this correctly being propagated?
We should indeed have a clone() function.

src/training/validator.h
```diff
@@ -125,13 +126,13 @@ class Validator : public ValidatorBase {
...
-    opts->merge(options);
-    opts->set("inference", true);
-    opts->set("cost-type", "ce-sum");
-    builder_ = models::from_options(opts, models::usage::scoring);
+    // set the cost type temporarily to 'ce-sum' for model builder
+    auto ctype = options_->get<std::string>("cost-type");
+    options_->set("cost-type", "ce-sum");
+    builder_ = models::from_options(options_, models::usage::scoring);
+    options_->set("cost-type", ctype);
   }
```
emjotde: Seems to me like CEValidator should have it's own option object that can be safely modified?

src/training/validator.h
```diff
@@ -180,6 +177,9 @@ class CrossEntropyValidator : public Validator<data::Corpus> {
       }
     }
 
+    // get back to the original cost type
+    options_->set("cost-type", ctype);
+
```
emjotde: As said above. Modification should be safer. This seems too fragile.

src/training/validator.h
```diff
@@ -193,15 +193,11 @@ class CrossEntropyValidator : public Validator<data::Corpus> {
 
 class ScriptValidator : public Validator<data::Corpus> {
 public:
-  ScriptValidator(std::vector<Ptr<Vocab>> vocabs, Ptr<Config> options)
+  ScriptValidator(std::vector<Ptr<Vocab>> vocabs, Ptr<Options> options)
       : Validator(vocabs, options, false) {
-    Ptr<Options> opts = New<Options>();
-    opts->merge(options);
-    opts->set("inference", true);
```
emjotde: Just to be sure, inherited from above? Maybe there should be `ABORT_IF(!options->get<bool>("inference", false), ...)` in every validator.

src/training/validator.h
```diff
@@ -229,48 +225,32 @@ class ScriptValidator : public Validator<data::Corpus> {
...
+    auto batchGenerator = New<BatchGenerator<data::Corpus>>(corpus, options_);
     batchGenerator->prepare(false);
 
     // Create scorer
     auto model = options_->get<std::string>("model");
-
-    auto mopts = New<Options>();
-    mopts->merge(options_);
-    mopts->set("inference", true);
-
```
emjotde: As above.

src/translator/scorers.cpp
```diff
@@ -1,24 +1,22 @@
...
 namespace marian {
 
 Ptr<Scorer> scorerByType(const std::string& fname,
                          float weight,
                          const std::string& model,
-                         Ptr<Config> config) {
-  Ptr<Options> options = New<Options>();
-  options->merge(config);
+                         Ptr<Options> options) {
   options->set("inference", true);
```
emjotde: clone for cleanliness?

src/translator/scorers.cpp
```diff
@@ -91,11 +89,15 @@ std::vector<Ptr<Scorer>> createScorers(Ptr<Config> options,
...
+    // load options specific for the scorer
+    auto modelOptions = New<Options>(*options);
     try {
-      if(!options->get<bool>("ignore-model-config"))
-        modelOptions->loadModelParameters(ptr);
+      if(!options->get<bool>("ignore-model-config")) {
+        YAML::Node modelYaml;
+        io::getYamlFromModel(modelYaml, "special:model.yml", ptr);
+        modelOptions->merge(modelYaml, true);
+      }
```
emjotde: This should be hidden. Maybe a static function in Config:: ? It is modifying an options object, but also clearly a configuration-specific operation.

src/examples/mnist/validator.h
```diff
@@ -14,12 +14,8 @@ namespace marian {
 
 class AccuracyValidator : public Validator<data::MNISTData> {
 public:
-  AccuracyValidator(Ptr<Config> options)
-      : Validator(std::vector<Ptr<Vocab>>(), options, false) {
-    Ptr<Options> temp = New<Options>();
-    temp->merge(options);
-    temp->set("inference", true);
-    builder_ = models::from_options(temp, models::usage::scoring);
```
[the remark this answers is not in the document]
snukky: The superclass does this upon construction, so options_ can be modified in subclasses.

src/training/scheduler.h
```diff
@@ -305,7 +305,7 @@ class Scheduler : public TrainingObserver {
 
   void save(const std::string& name) {
     // Save config options
-    YAML::Node config = options_->get();
+    YAML::Node config = options_->getOptions();
```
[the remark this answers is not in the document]
snukky: I agree, gettingOptions from an Options object is confusing. getYaml is better.

src/training/validator.h
snukky: Yes, it should enable subclasses to conveniently modify options_. It's seems to be correctly propagated as tests fail without this.

src/training/validator.h
snukky: It's not because options_ can not be modified here for some reason, but because "cost-type" seems to mean something different for models and for validators? Very confusing, but I haven't investigated this deeper as I assumed to be out of the scope of this PR.

src/training/validator.h
snukky: See comment above.

src/training/validator.h
snukky: Agree. Adding this to Validator should be enough, though.

src/translator/scorers.cpp
snukky: IMHO clone here would be redundant as `scorerByType` is only used in `createScorers` using a newly cloned options specific for each scorer. 

To be able to test this I have added regression tests for ensembles of models with different architectures.

src/translator/scorers.cpp
snukky: Agree. I aimed for minimal changes.

snukky:
@emjotde I responded to your comments so far, but will not update the code until all suggestions are clear/complete.

snukky:
Merged with the current master.

PR #353, opened 2018-10-22
Allow compilation with user-specified BUILD_ARCH instead of just -march=native.

ugermann:
The previous attempt (with -DCMAKE_BUILD_TYPE=Nonative) had no effect.
Now -DBUILD_ARCH=... allows compilation of portable binaries that
will also run on CPUs older than the one Marian was compiled on. -DBUILD_ARCH=x86-64 gives the most generic version.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,6 +16,7 @@ option(USE_CUDNN "Use CUDNN library" OFF)
 option(USE_NCCL "Use NCCL library" ON)
 option(USE_MPI "Use MPI library" OFF)
 option(USE_SENTENCEPIECE "Download and compile SentencePiece" OFF)
+option(COMPILE_NATIVELY "Generate native code (-march=native => less portable)" ON)
 
 # Project versioning
 find_package(Git QUIET)
@@ -26,6 +27,12 @@ message(STATUS "Project version: ${PROJECT_VERSION_STRING_FULL}")
 
 execute_process(COMMAND git submodule update --init --recursive --no-fetch
                 WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
+              
+if (COMPILE_NATIVELY)
+  set(MARCH native)
+else()     
+  set(MARCH "x86-64 -mavx")
+endif()
 
 # Set compilation flags
 if(MSVC)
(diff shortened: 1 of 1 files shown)


emjotde:
I am wondering about the -msse4.1 flag. How is that one special compared to others?

@kpu any onions about how this should be done?

kpu:
So far as I know `-msse4.1` is only _needed_ to compile a matrix multiply in low precision in `src/tensors/cpu/sharp/avx_gemm.cpp`.  As such it should be set in CMakeLists.txt with `set_source_files_properties`.  This can be attached to just that file and should be harmless to @ugermann's use case because he isn't using it and there's CPUID dispatch around that code.  

The vast majority of Marian CPU code (everything except MKL and the above code) uses implicit SIMD with some pragmas scattered about.  Compiler people like it that way because it genericizes to other architectures.  That means generically compiled kernels will be slower i.e. use 128-bit SIMD instead of 256-bit SIMD.  

So I think it's sensible to let the user specify their own `CFLAGS` and `CXXFLAGS` in the normal environment variable way, default to adding `-march=native` and allow that addition to be turned off in a `cmake` argument.

emjotde:
So in short, `-march=native` is smart enough to do well for us in the general case and you would recommend to set `-msse4.1` for `avx_gemm.cpp` only.

kpu:
Yes, but I believe Uli's chief complaint is he wants the ability to turn off `-march=native`.

emjotde:
Sure, I am fine with the change and the optional change of `-march`. Just wanted to know what it entails. Thanks for your help.

emjotde:
Also I am seeing now that autocorrect hilariously changed opinions into onions.

ugermann:
@emjotde: Just for the record: as far as I understand the -m<...> flags enable specific instruction sets; the -march option is a shorthand for the list of instruction sets supported by the specific CPU that the code is being compiled on, breaking portability to older CPUs that might not support some of the flags. You can find out for your CPU what -march translates into by running this:

```
gcc -march=native -E -v - </dev/null 2>&1 | grep cc1
```

So -march=native implies -msse4.1 for most CPUs after 2006 or so. Without -msse4.1, I wasn't able to compile with MKL when using the most generic -march=x86-64.

You can also tune for a specific CPU (portable to older CPUs but tuned for a specific CPU by using -mtune instead of or in addition to -march).

emjotde:
Oh, that gcc trick is useful!

kpu:
@ugermann We care mostly about SIMD width, which `mtune` won't change.  There is of course instruction latency that `mtune` can take advantage of.

PR #371, opened 2018-11-09
Const diligence

ugermann:
Let's try this again ...


emjotde:
@ugermann can you open the PR once more? If you cannot reproduce the compilation issues we can probably help.

Issue #343, opened 2018-10-16
Tied-layers cannot be overwritten with empty vector from command line


emjotde:
Tied-layers cannot be overwritten with empty vector from command line when specified in config.

frankseide:
Example:

 * in config file:
 `transformer-tied-layers: [1,1,1,1,1,1]`
 * on command line:
  `--config ... --transformer-tied-layers  --no-shuffle ...`

leads to `tiedLayers` variable to contain `{1, 1, 1, 1, 1, 1}`.

snukky:
This is somehow related to setting boolean flags to `false` via command line if they are `true` in a config file.

frankseide:
And also to the problem that some variables can have the value "not set." For example, to not use guided alignments, one should not define `--guided-alignment`. But if it is already defined in a config file, it cannot be overwritten to "unset". For now, I use a hack in that if it is set to the filename "none", it is treated as if unset, which has the same problem as naming stdin "stdin"--what if that's the actual filename?

We would either need an "unset" or "undefined" value, or not give "unset" any other meaning than "default", consistently for all parameters.

emjotde:
OK. Using '[]' works, but what is weird now is that you can provide empty vector arguments that have no effect. That seems wrong. Either they should overwrite with [] too (preferred) or there should be an abort.

For instance `--optimizer-params` with no value is completely ignored at the moment, which is fine if it has never been set, but problematic if there is a value in the config. What is technically possible here? 

It seems that the call-back function is not being called for empty vectors?

PR #377, opened 2018-11-10
Updates to vector-like options: empty vectors and implicit values

snukky:
This fixes the issue mentioned in #343, e.g.:

```
./build/marian -t toy.bpe.de toy.bpe.en -m model.npz -v vocab.de.yml vocab.en.yml \
    --config config.yml --transformer-tied-layers
Error: Expected at least 1 arguments to --transformer-tied-layers, got 0
Run with --help for more information.
```

The fix required a modification of CLI11.


emjotde:
hm. Is this what we actually wanted?

snukky:
Now, this PR also fixes using implicit values for vector-like options.

snukky:
Updated with the current master.

PR #380, opened 2018-11-13
fix compile error on Ubuntu 18.04, gcc 7.3

hieuhoang:


emjotde:
We are getting rid of boost right now for that file. This fix also messes with warnings. Since this will go away soon without that fix I would rather not pull it.

PR #383, opened 2018-11-25
Make SentencePiece trainable from within Marian

emjotde:
* When SentencePiece support is compiled in, Marian can now train SentencePiece models for its input files. 
* Added grouping of training files when they correspond to the same vocabulary, used for joint vocabularies.
* SentencePiece will by default sample a representative subset of lines if there are more than 10,000,000 lines. 
* Removed hard-attention models and corresponding special vocab items.


src/common/config_parser.cpp
```diff
@@ -255,8 +255,13 @@ void ConfigParser::addOptionsTraining(cli::CLIWrapper& cli) {
       "If these files do not exist they are created");
 #ifdef USE_SENTENCEPIECE
   cli.add<std::vector<float>>("--sentencepiece-alphas",
-                              "Sampling factors for SentencePieceVocab;"
-                              "i-th factor corresponds to i-th vocabulary");
+      "Sampling factors for SentencePieceVocab; i-th factor corresponds to i-th vocabulary");
```
snukky: SentencePieceVocab is from our code, not a term used in sentence piece toolkit, so I would prefer to use something less confusing for users, for instance, "SentencePiece vocabulary".

src/data/default_vocab.cpp
```diff
@@ -187,59 +181,62 @@ class DefaultVocab : public VocabBase {
...
-            dir.string());
+    if(vocabPath != "stdout") {
+      filesystem::Path path(vocabPath);
+      auto dir = path.parentPath();
+      if(dir.empty())
+        dir = filesystem::currentPath();
 
-    ABORT_IF(!dir.empty() && !filesystem::canWrite(dir),
-            "No write permission in vocab directory {}",
-            dir.string());
```
snukky: I think we have a test for this, so it will need to be removed or updated.

src/data/corpus_base.cpp
```diff
@@ -92,9 +96,25 @@ CorpusBase::CorpusBase(Ptr<Options> options, bool translate)
       if(maxVocabs.size() < vocabPaths.size())
         maxVocabs.resize(paths_.size(), 0);
 
+      // Group training files based on vocabulary path. If the same
+      // vocab path corresponds to different training files, this means
+      // that a single vocab should combine tokens from all files.
+      typedef std::pair<std::set<std::string>, size_t> PathsAndSize;
+      std::map<std::string, PathsAndSize> groupVocab;
```
snukky: Consider introducing a simple structure to improve readability a bit.

src/data/default_vocab.cpp
```diff
@@ -30,14 +30,14 @@ class DefaultVocab : public VocabBase {
...
+    const std::unordered_map<std::string, size_t>& counter_;
 
   public:
-    VocabFreqOrderer(std::unordered_map<std::string, size_t>& counter)
+    VocabFreqOrderer(const std::unordered_map<std::string, size_t>& counter)
         : counter_(counter) {}
 
     bool operator()(const std::string& a, const std::string& b) const {
-      return counter_[a] > counter_[b] || (counter_[a] == counter_[b] && a < b);
+      return counter_.at(a) > counter_.at(b) || (counter_.at(a) == counter_.at(b) && a < b);
```
snukky: Please add a comment explaining what it does.

src/data/sentencepiece_vocab.cpp
```diff
@@ -28,9 +32,12 @@ class SentencePieceVocab : public VocabBase {
   Ptr<Options> options_;
   size_t batchIndex_{0};
 
+  std::mt19937 generator_;
```
snukky: In other classes we use https://github.com/marian-nmt/marian-dev/blob/master/src/data/rng_engine.h It has been introduced for serialization of the seed. Maybe we can use it here as well?

PR #385, opened 2018-11-25
Another step towards removing boost

emjotde:
* Adds Pathie as a replacement for boost::filesystem
* Adds zstr as a replacement for boost::filtering_stream


src/common/config_validator.cpp
```diff
@@ -84,9 +84,6 @@ void ConfigValidator::validateOptionsTraining() const {
   ABORT_IF(!modelDir.empty() && !filesystem::isDirectory(modelDir),
            "Model directory does not exist");
 
-  ABORT_IF(!modelDir.empty() && !filesystem::canWrite(modelDir),
-           "No write permission in model directory");
```
snukky: This will very likely break one of our regression tests, so we will need to update it.

src/common/logging.cpp
```diff
@@ -150,9 +150,11 @@ void switchtoMultinodeLogging(std::string nodeIdStr) {
 
 
 namespace marian {
-  void noinline logCallStack(size_t skipLevels)
-  {
-    auto callStack = ::Microsoft::MSR::CNTK::DebugUtil::GetCallStack(skipLevels + 2, /*makeFunctionNamesStandOut=*/true);
-    checkedLog("general", "critical", "Call stack:{}", callStack);
+  std::string getCallStack(size_t skipLevels) {
+    return ::Microsoft::MSR::CNTK::DebugUtil::GetCallStack(skipLevels + 2, /*makeFunctionNamesStandOut=*/true);
```
snukky: Does this require some external dependency or should be used only with a specific compilation settings? Regardless of whether it does or not, a comment might be helpful here.

src/common/logging.h
```diff
@@ -46,13 +47,19 @@ namespace marian {
...
+    if(logger == nullptr)                                                      \
+      logger = createStderrLogger("general", "[%Y-%m-%d %T] Error: %v");       \
+    else                                                                       \
+      logger->set_pattern("[%Y-%m-%d %T] Error: %v");                          \
+    checkedLog("general", "critical", __VA_ARGS__);                            \
+    checkedLog("general", "critical", "Aborted from {} in {}:{}",              \
+               FUNCTION_NAME, __FILE__, __LINE__);                             \
+    logger->set_pattern("%v");                                                 \
+    checkedLog("general", "critical", marian::getCallStack(/*skipLevels=*/0)); \
+    std::abort();                                                              \
```
snukky: This refactorization is not related to removing Boost dependency, right? Just a question.

src/layers/word2vec_reader.h
```diff
@@ -18,8 +18,6 @@ class Word2VecReader {
     LOG(info, "[data] Loading embedding vectors from {}", fileName);
 
     io::InputFileStream embFile(fileName);
-    ABORT_IF(!embFile.isOpen(),
-             "Unable to open file with embeddings: " + fileName);
```
snukky: We don't need this or the condition is checked implicitly by InputFileStream?

src/data/default_vocab.cpp
```diff
@@ -199,10 +199,6 @@ class DefaultVocab : public VocabBase {
             "Specified vocab directory {} does not exist",
             dir.string());
 
-    ABORT_IF(!dir.empty() && !filesystem::canWrite(dir),
-            "No write permission in vocab directory {}",
-            dir.string());
```
snukky: As above, we need to remember that this might need updating one of the regression tests.

src/common/logging.h
emjotde: No, this was a general aesthetic thing.

src/common/logging.cpp
emjotde: There are no dependencies here. Just a weird cascade of namespaces.

src/data/default_vocab.cpp
emjotde: I think our regression tests will remind us :)

src/layers/word2vec_reader.h
emjotde: We are now aborting on badbit and failbit. This was not working before. So a failure to read aborts anyway.

PR #400, opened 2018-12-06
Integrate NCCL build with Marian

emjotde:
* Link statically against CUDA libraries with -DUSE_STATIC_LIBS
* Add NCCL fork as a submodule
* Add custom command to build with NCCL makefile from CMake
* Pass build directory and CXX compiler into NCCL makefile
* Add target to make "make clean" for NCCL outputs
* Include generated "nccl.h"
* Test for CPU intrinsics


src/CMakeLists.txt
```diff
@@ -182,7 +181,7 @@ endif(COMPILE_SERVER)
 foreach(exec ${EXECUTABLES})
   target_link_libraries(${exec} marian ${EXT_LIBS} ${EXT_LIBS} ${CMAKE_THREAD_LIBS_INIT})
   if(CUDA_FOUND)
-    target_link_libraries(${exec} marian marian_cuda ${EXT_LIBS} ${CMAKE_THREAD_LIBS_INIT})
+    target_link_libraries(${exec} marian marian_cuda nccl ${EXT_LIBS} ${CMAKE_THREAD_LIBS_INIT})
```
emjotde: Note to self: why is CMake not screaming at me that target nccl or libnccl does not exist when I am building with -DUSE_NCCL=off ?

CMakeLists.txt
```diff
@@ -91,23 +137,49 @@ if(CUDA_FOUND)
...
-
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DCUDA_FOUND")
   list(APPEND CUDA_NVCC_FLAGS -DCUDA_FOUND; )
 
   if(MSVC)
     list(APPEND CUDA_NVCC_FLAGS -DBOOST_PP_VARIADICS=0; )
   endif()
 
+  if(USE_NCCL)
+    # define and set the include dir for the generated nccl.h header
```
frankseide: I would add to the comment: "Who generates header files anyway!!?"

CMakeLists.txt
```diff
@@ -91,23 +137,49 @@ if(CUDA_FOUND)
...
+    set(NCCL_HEADER_LOCATION "${CMAKE_CURRENT_BINARY_DIR}/nccl/include")
+    include_directories(${NCCL_HEADER_LOCATION})
+
+    # set the path for the generated static lib
+    set(NCCL_STATIC "${CMAKE_CURRENT_BINARY_DIR}/nccl/lib/libnccl_static.a")
```
frankseide: Should there be a "LIB" in the variable name? Not sure what the conventions in CMake files are.

CMakeLists.txt
```diff
@@ -91,23 +137,49 @@ if(CUDA_FOUND)
...
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DUSE_NCCL")
+
+    LIST(APPEND CUDA_NVCC_FLAGS -DUSE_NCCL; )
+
+    # disables compilation for sm_30 to avoid ptxas warning... that's general Kepler support. But K80s are supported for instance by sm_35
```
frankseide: Does Marian still support Kepler? Does it actually work? I had a problem with back compat, where some CUDA operations on an old architecture would just compute zeroes, without getting an error.

CMakeLists.txt
```diff
@@ -91,23 +137,49 @@ if(CUDA_FOUND)
...
+    set(GENCODE "-gencode=arch=compute_35,code=sm_35 -gencode=arch=compute_50,code=sm_50 -gencode=arch=compute_60,code=sm_60 -gencode=arch=compute_61,code=sm_61")
+
+    # sets output to build, passes cuda location from FindCUDA, sets c++ compiler to the same one CMake uses.
```
frankseide: (minor: cuda -> CUDA ?)

CMakeLists.txt
```diff
@@ -91,23 +137,49 @@ if(CUDA_FOUND)
...
+    add_custom_command(OUTPUT ${NCCL_STATIC}
```
frankseide: add comment: # build using NVidia's custom makefile

CMakeLists.txt
frankseide: maybe clarify: "set output to our usual build folder..."

cmake/FindSSE.cmake
```diff
@@ -0,0 +1,148 @@
+# Check if SSE/AVX instructions are available on the machine where
```
frankseide: This seems not a good criterion, unless it is merely for use as a default. I suggest to add to the comment right here how to override it when calling CMake.

src/tensors/cpu/device.cpp
```diff
@@ -15,18 +15,26 @@ Device::~Device() {
   size_ = 0;
 }
 
+#ifdef _WIN32
```
frankseide: Please add a comment what this is used for, e.g. // allocate function for tensor reserve() below. Needed for AVX2, while not available on all compilers.

src/tests/CMakeLists.txt
```diff
@@ -12,7 +12,6 @@ foreach(test ${UNIT_TESTS})
 
     if(CUDA_FOUND)
       target_link_libraries("run_${test}" marian marian_cuda ${EXT_LIBS} Catch)
```
frankseide: Does this mean the tests are now part of the repo (submodule or not)?

src/training/communicator_nccl.h
```diff
@@ -4,8 +4,11 @@
 #include "3rd_party/threadpool.h"
 #include "tensors/gpu/cuda_helpers.h"
 
-#include "cuda_runtime.h"
+// Generated by NCCL make files in build/nccl/include
```
frankseide: Maybe change to
// we explicitly include nccl.h here first to ensure that it uses our own version (cuda_runtime.h includes it as <nccl.h> which would not find it)
(or whatever the correct explanation is)

CMakeLists.txt
emjotde: NCCL adds a version number. I think we do the same? :)

src/tests/CMakeLists.txt
emjotde: These are unit tests, not regression tests. They are enabled with options. Only check for a number of core functionalities.

PR #403, opened 2018-12-09
Add options to disable tensor cores

emjotde:
This allows to switch off tensor cores when running on the GPU. The option appears when compiled with CUDA support. Currently this is enabled by default on CUDA 9.0 and higher, but has no effect on GPUs that do not support Tensor Cores. I noticed that using Tensor Cores does not seem to be entirely free. With CUDA 9.2 I saw NANs appearing.


src/common/config.cpp
```diff
@@ -55,6 +63,13 @@ void Config::initialize(int argc, char** argv, cli::mode mode, bool validate) {
     seed = get<size_t>("seed");
   }
 
+  #ifdef CUDA_FOUND
+  // set if Tensor Cores should be used
+  useTensorCores = !get<bool>("no-tensor-cores");
```
snukky: This is more like a question: since we can make a boolean option `true` by default and disable it by `--bool-option false|off|no|0`, maybe we should try to avoid introducing more options like `--no-something`? Double negations are not easy to read.

src/tensors/gpu/backend.h
```diff
@@ -29,9 +29,19 @@ class Backend : public marian::Backend {
 
   cublasHandle_t getCublasHandle() { return cublasHandle_; }
 
+#if CUDA_VERSION >= 9000
+  bool useTensorCores() {
+    return useTensorCores_;
+  }
```
snukky: Wouldn't it be better to have this function always defined and just returning `false` if CUDA < 9.0? That would possibly minimize the uses of `#if` statements in other places. 

If CUDA_VERSION implies CUDA_FOUND, maybe put this here explicitly?

src/common/config.cpp
emjotde: Yeah, I was thinking about this. This is like a diagnostic options, where we specifically disable a rather useful feature. We also have --no-restore and --no-shuffle. Usually you would want in all those cases to have them enabled. Using them seems very explicit.

src/tensors/gpu/backend.h
emjotde: yes, that's probably better, I am going to change this.

src/tensors/gpu/backend.h
```diff
@@ -29,9 +29,19 @@ class Backend : public marian::Backend {
...
+#endif
+
 private:
   cublasHandle_t cublasHandle_;
 
+#if CUDA_VERSION >= 9000
+  bool useTensorCores_{Config::useTensorCores};
```
frankseide: How about removing the #if, and instead turn this into a request? requestTensorCores_

Which will throw a runtime error if it cannot be fulfilled.

emjotde:
I am abandoning this PR for now as we have the ENV variable for switching off TC, although I really dislike that solution. Also I can never remember the name of the variable and the only way to find that out is reading prod.cpp.

Need to look at that again and come up with something smarter.

PR #405, opened 2018-12-11
Some updates in CLI, tests, utils

snukky:
A few updates related to some issues or code TODOs:
* log unexpected options occurring in config files (related to #402)
* handle "version" in config files (#402)
* add regression tests as a submodule (#379)
* organize tests in `src/tests/`
* refactorize `utils::split*`
* clean `TextInput`
* change some options to floats
* remove `add_nondefault()` in config parser and add `Options::nonempty`, which is `Options::has` + nonempty string or vector
* remove `--special-vocab`
* restrict `--port` to marian server only

There are two extra regression tests in the `some-updates` branch that I'll merge later.

Issue #394, opened 2018-11-29
Adding a --best-practice option


emjotde:
I would like to add a `--best-practice` option to Marian which should serve as a short-cut to certain settings. For instance one value would be 
```
marian --best-practice transformer-big
```
which would correspond to
```
marian --model transformer \
--dim-emb 1024 --transformer-dim-ffn 4096 --transformer-heads 16 \
--enc-depth 6 --dec-depth 6 \
--transformer-postprocess dan --transformer-preprocess d \
--transformer-ffn-activation relu \
--transformer-dropout 0.1 --transformer-attention-dropout 0.1 --transformer-ffn-dropout 0.1 \
--learn-rate 0.0002 --lr-decay-sqrt-inv 8000 --lr-warmup 8000 --label-smoothing 0.1 \
```

Other values for best-practice might be combinations of the following and more:
```
- (rnn|transformer|rnn|lm|tranformer-lm)-(base|big)(-(low_resource|multi_source|...))?
- WMT17-en-de
- WMT18-en-de
- ...
```

Thoughts?

hieuhoang:
It's a good idea. Our maybe a little program that asks done questions and gives you the command you can run. Then people can see exactly what they're getting

emjotde:
ah, that's clever. `./marian-expert`

kpu:
Presumably this would be tweaked over time.  So make it clear that it's not going to behave consistently across versions.

frankseide:
An alternative would be to ship config files with these things. If you store them right next to the executable, then we can allow --config to find them, or a different option.

Baking these into the compiled binary feels opaque. On the other hand, explicit human-readable config files are a good basis for modification.

emjotde:
@frankseide that's a good idea too.

emjotde:
These options could be generated via `marian --best-practice=blabla --dump-config=minimal > myconfig.yml`

emjotde:
@snukky What do other toolkits call these kind of options?

snukky:
T2T calls it `--problem` (https://github.com/tensorflow/tensor2tensor#translation). Sockeye has an autopilot that uses `--model` (https://github.com/awslabs/sockeye/tree/master/sockeye_contrib/autopilot) . Fairseq seems to use `--arch` that stands for "architecture".

emjotde:
Out of these I kinda like `--problem` best, the others don't mean anything.

emjotde:
[...]

PR #413, opened 2019-01-15
Adding a --task option

snukky:
This PR will:
* add a `--problem` option (mentioned in #394) with example aliases for "transformer" and "transformer-big"
* add `--dump-config explain` for dumping a config file with expanded alias options
* bypass some validation rules if `--dump-config` is used, e.g. training sets and vocabs don't have to be set

**Note:** the pull request adds changes to the `some-updates` branch, not `master`. It is separated from PR #405 to have a clear view on modifications and a distinct space for discussion.


emjotde:
Did we say the option should be called "problem" in the end? It sounds so negative (might be honest though). I'd prefer "experiment" or maybe "task"?

emjotde:
Ha, I actually said "problem". Funny, I hate it now :)

emjotde:
God.... I think I am getting old and moody. I want "task" now.

snukky:
Renamed. I also like "task" much more than "problem". We should check if we can add  more aliases in this pull request and if the currently defined "transformer" and "transformer-big" are OK.

src/common/cli_wrapper.cpp
```diff
@@ -128,6 +128,38 @@ void CLIWrapper::parse(int argc, char **argv) {
   }
 }
 
+void CLIWrapper::parseAliases() {
```
emjotde: Since this is a new functionality, can you add more comments about this function and also inside the function?

src/common/cli_wrapper.cpp
```diff
@@ -128,6 +128,38 @@ void CLIWrapper::parse(int argc, char **argv) {
...
+  if(aliases_.empty())
```
emjotde: aliases are YAML nodes?

emjotde: No, AliasTuple apparently, this deserves a comment, where to find the actual object definition.

src/common/cli_wrapper.cpp
```diff
@@ -128,6 +128,38 @@ void CLIWrapper::parse(int argc, char **argv) {
...
+        auto aliasOpts = config_[alias.key].as<std::vector<std::string>>();
+        expand = std::find(aliasOpts.begin(), aliasOpts.end(), alias.value) != aliasOpts.end();
+      } else {
+        // Note: options values are always compared as strings
+        expand = config_[alias.key].as<std::string>() == alias.value;
+      }
+
+      if(expand) {
+        updateConfig(alias.config,
+                     "Unknown option(s) in alias '" + alias.key + ": " + alias.value + "'");
```
emjotde: Is this a hidden abort message? When does it fire?

src/common/cli_wrapper.cpp
```diff
@@ -128,6 +128,38 @@ void CLIWrapper::parse(int argc, char **argv) {
...
+      }
+      aliasKeys.insert(alias.key);
+    }
+  }
+
+  // Remove aliases from the config
```
emjotde: Comment: Why?

src/common/cli_wrapper.h
```diff
@@ -51,16 +51,21 @@ class CLIFormatter : public CLI::Formatter {
...
- */
+// The helper structure storing an option object, the associated variable and creation index
 struct CLIOptionTuple {
   CLI::Option *opt;
   Ptr<any_type> var;
   size_t idx{0};
   bool modified{false};
 };
 
+// Helper structure used for aliases and storing an option key, value, and YAML node
```
emjotde: Maybe also explain what an alias is.

src/common/config_parser.cpp
```diff
@@ -79,7 +79,8 @@ void ConfigParser::addOptionsGeneral(cli::CLIWrapper& cli) {
   cli.add<bool>("--relative-paths",
      "All paths are relative to the config file location");
   cli.add<std::string>("--dump-config",
-     "Dump current (modified) configuration to stdout and exit. Possible values: full, minimal")
+     "Dump current (modified) configuration to stdout and exit. "
+     "Possible values: full, minimal, explain")
```
emjotde: Add comment: what do the values mean. I think we already saw that "minimal" is actually a bit counterintuitive.
Please explain "explain". Is that a good name?

src/common/config_parser.cpp
```diff
@@ -413,8 +414,12 @@ void ConfigParser::addOptionsTraining(cli::CLIWrapper& cli) {
   cli.add<bool>("--multi-node-overlap",
      "Overlap model computations with MPI communication",
      true);
+
   // add ULR settings
   addSuboptionsULR(cli);
+
+  cli.add<std::vector<std::string>>("--task",
```
emjotde: Another proposition for the name was "--defaults-for-task". I am conflicted.

src/common/config_parser.cpp
```diff
@@ -687,11 +692,10 @@ void ConfigParser::addSuboptionsULR(cli::CLIWrapper& cli) {
   // clang-format on
 }
 
-void ConfigParser::expandAliases(cli::CLIWrapper& cli) {
-  YAML::Node config;
+void ConfigParser::addAliases(cli::CLIWrapper& cli) {
   // The order of aliases does matter as later options overwrite earlier
 
-  if(config_["best-deep"].as<bool>()) {
+  cli.alias("best-deep", "true", [](YAML::Node& config) {
```
emjotde: I think these should be moved to a separate file, either by moving the function, or by using an `#include`.

```
void ConfigParser::addAliases(cli::CLIWrapper& cli) {
#include "common/aliases.inc"
}
```

@frankseide Is this ugly?

emjotde: Or better have the entire function inside `common/aliases.inc`

src/common/config_parser.cpp
```diff
@@ -701,11 +705,42 @@ void ConfigParser::expandAliases(cli::CLIWrapper& cli) {
...
     config["dec-depth"] = 4;
     config["skip"] = true;
-  }
-
-  if(config) {
-    cli.updateConfig(config, "Unknown option(s) in aliases");
-  }
+  });
+
+  cli.alias("task", "transformer", [](YAML::Node& config) {
```
emjotde: I guess these should also have some comments, I guess this is something I can do later.

src/common/config_validator.cpp
```diff
@@ -10,7 +10,10 @@ bool ConfigValidator::has(const std::string& key) const {
   return config_[key];
 }
 
-ConfigValidator::ConfigValidator(const YAML::Node& config) : config_(config) {}
+ConfigValidator::ConfigValidator(const YAML::Node& config)
+    : config_(config),
+      dump_(config["dump-config"] && !config["dump-config"].as<std::string>().empty()
```
emjotde: What is this doing?

src/common/config_validator.h
```diff
@@ -5,11 +5,10 @@
 
 namespace marian {
 
-// TODO: Finally refactorize Config, Options, ConfigParser and ConfigValidator
-// classes.
 class ConfigValidator {
 private:
   const YAML::Node& config_;
+  bool dump_{false};
```
emjotde: Maybe a longer more descriptive name?

src/common/config_parser.cpp
emjotde: Roman suggested to just put the function definition into a separate aliases.cpp which is of course much better. I have done something like that in the `bert` branch.

emjotde:
Problem: when `task` is used with `dump-config` the options that were specified on the command line are not not identified anymore as options that should override the task options. When marian is then called with

```
marian --task transformer --dim-emb 256 --dump-config > test.yml
```

This will not still train with `dim-emb: 512` because task overrides `dim-emb` from the config:
```
marian --config test.yml 
```

snukky:
Recent changes:
* added comments
* fixed the mentioned problem by introducing priority levels for options
* moved alias definitions to a separate file

Additional regression tests are in https://github.com/marian-nmt/marian-regression-tests/tree/some-updates

src/common/cli_wrapper.cpp
snukky: Function documentation was in the header file. Added more comments inside the function.

src/common/cli_wrapper.cpp
snukky: See documentation of updateConfig() in the header file.

src/common/config_parser.cpp
snukky: My vote for `--task` as it's shorter.

PR #415, opened 2019-01-23
C++ source should not have execute bits

kpu:


emjotde:
I should hold this for a moment, want to check if this will cause merge problems. A couple of bigger ones is coming.

emjotde:
@frankseide would this get in your way somehow?

frankseide:
I have no idea. However, this will keep happening. It is an artefact of Windows and Visual Studio, which somehow keeps messing with the X permissions on Windows, which doesn't have the concept of X permissions. If this does not cause a problem, I suggest to ignore it until VS handles this correctly.

emjotde:
I will pull this in and let's see what happens. Does not seem to cause merge conflicts as no real content has changed.

PR #420, opened 2019-02-02
reset gradients in default comm.

afaji:
default communicator does not reset gradient in scatterReduceAndResetGrads() function.

Do you think it will be clearer if we remove resetGrads functionality in that function, instead just reset the full gradients after update as in https://github.com/marian-nmt/marian-dev/blob/50d64de62cb3e47b912fdd3438ffacb566187263/src/training/graph_group_sync.cpp#L389?

### DIFF
--- a/src/training/communicator.h
+++ b/src/training/communicator.h
@@ -175,7 +175,17 @@ public:
       }
     };
 
+    // reset gradients outside current shard
+    auto reset = [this, shardSize](size_t idx, size_t begin, size_t end) {
+      auto grad = graphs_[idx]->params()->grads();
+      if (begin > 0)
+        grad->subtensor(0, begin)->set(0);
+      if (end < grad->size())
+        grad->subtensor(end, grad->size()-end)->set(0);
+    };
+
     foreach(scatter);
+    foreach(reset);
   }
 
   void allGatherParams() const override {


frankseide:
Yikes, that's a serious bug. However, the reason for this grouping of functionality is that communication may be lossy. Not in the current version, but maybe some day in the future (and in fact in a research branch that hasn't panned out so far).

The idea is that the communicator may decide to not transfer gradients in full fidelity, but remember the residual error, which is done by storing the residual in the gradient, so that the next minibatch will add the new gradient on top of the residual.

This technique comes from [1-bit SGD](https://www.microsoft.com/en-us/research/publication/1-bit-stochastic-gradient-descent-and-application-to-data-parallel-distributed-training-of-speech-dnns/), and is also used in Amazon's [sparse gradient SGD](https://www.isca-speech.org/archive/interspeech_2015/i15_1488.html).

So let's fix this in the `DefaultCommunicator`, not in the various versions of group group update().

We should also see how we can catch this in testing. We should have at least one test run with `--no-nccl`, which will force using`DefaultCommunicator`.

emjotde:
I also find the current solution highly confusing.

emjotde:
Especially since it circumvents the intended usage of `graph->backward(/*reset=*/true)` which we had before for dealing with delayed updates.

afaji:
Hi, do we have any update on this?

frankseide:
Would you mind fixing it consistently with the current API?

afaji:
This pull request follows the current API behavior, i.e. scatterReduceAndResetGrads()  will reset the gradients to 0.

Please let me know if I make a mistake

kpu:
Ping.

kpu:
There is a bug in master for the MPI-based multi-node training.  This pull request fixes it.  I did poke Alham about maintaining the MPI codepath and he commented that this request hadn't been approved.  

I would also respectfully disagree with "It is not currently needed" by saying it might not be needed at Microsoft but there are users with gigabit ethernet.

frankseide:
Approved.

PR #422, opened 2019-02-05
Update NCCL version to 2.4.2 - use ExternalProject

emjotde:
Small PR:
* Updates NCCL to version 2.4.2
* Starts using CMake's ExternalProject for building NCCL and potentially other external dependencies which is much cleaner than the previous custom command. This also makes the Make build server work with `make -j8`


emjotde:
Wrong branch

PR #423, opened 2019-02-05
Update NCCL to 2.4.2, build with CMake's ExternalProject

emjotde:
Small PR:
* Updates NCCL to version 2.4.2
* Starts using CMake's ExternalProject for building NCCL and potentially other external dependencies which is much cleaner than the previous custom command. This also makes the Make build server work with `make -j8`


CMakeLists.txt
```diff
@@ -172,37 +172,10 @@ endif(USE_STATIC_LIBS)
   # Cmake. This is also fairly untested, let's hope it does not explode.
```
frankseide: Is this remark still applicable? It seems to work now, no?

emjotde: True, that can be removed.

PR #434, opened 2019-03-01
Add switch fallthroughs for marian-nmt/marian#190

kpu:


rihardsk:
It seems that support for `[[fallthrough]]` was added in GCC/G++ 7, so when using GCC/G++ 5.4.0 i get
```
/home/rkroot/src/marian-dev/src/tensors/cpu/sharp/avx_gemm.cpp:498:7: error: attributes at the beginning of statement are ignored [-Werror=attributes]
       [[fallthrough]];
```
What works instead is `/* fall through */`.

emjotde:
What dark magic is this?

kpu:
There's a switch that falls through intentionally to handle matrices of different sizes.  GCC 7 complains about missing `[[fallthrough]]` (and warnings are errors in Marian...) while GCC 5.4.0 is confused by `[[fallthrough]]` :'(

emjotde:
So this is a deliberate marker that you are OK with not having an explicit "break;" here? Since this is new behavior, we could just switch off that warning. @frankseide ?

kpu:
That would also work, depending on the scope with which the flag is used.

frankseide:
This is a good thing. However, it is a C++ 17 feature. Older compilers do not support this. I think we need to continue to support gcc 5.4 for a while.

How about a macro "fallthrough;" that is defined for all compilers in just the right way (possibly empty)?

BTW our build with Visual Studio 2017 says this:

    1>\\mt-gpu-009\home\fseide\marian-dev\src\common\utils.cpp(369): error C2429:
        attribute 'fallthrough' requires compiler flag '/std:c++17'

so we possibly could make it work here by just enabling this flag (if gcc 7 has no other issues, then we are likely already quite C++ 17 compliant, so the fallout from this switch at other places may be limited).

emjotde:
Is g++-7 just assuming C++17? Isn't that weird?

frankseide:
Why would the latest compiler not default to the latest standard?

emjotde:
Not how it used to be with gcc. Usually you have to enable the standards. Didn't know that changed.

rihardsk:
> This is a good thing. However, it is a C++ 17 feature. Older compilers do not support this. I think we need to continue to support gcc 5.4 for a while.
>
> How about a macro "fallthrough;" that is defined for all compilers in just the right way (possibly empty)?

@frankseide how about using `/* fall through */` instead of `[[fallthrough]]` (see my previous comment)? I've tested that it works on both gcc 5.4.0 and gcc 7.

PR #437, opened 2019-03-22
Don't close named pipes during corpus reset.

### LINKED ISSUE #425: Marian decoder produces garbage when the input file is a FIFO pipe
In bash, 
```
marian-decoder -c config.yml -i <(cat file)
```
should produce the same result as 
```
marian-decoder -c config.yml < file
```
(Let's just assume that the user has a good reason to do the former rather than the latter.)

Marian does not handle the first variant properly and produces garbage. The input reader should check (e.g. via S_ISFIFO(m) from sys/stat.h) if the input is  a regular file or a pipe.

ugermann:
Corpus::reset() closed and reopened input files for corpora. So files were opened once
during the construction of the Corpus instance, then closed and repopened during
data->reset() in BatchGenerator::prepare(false). With normal files that's not a problem,
but if the "file" is a named pipe, the closing triggers a SIGPIPE (broken pipe) on
the writing end of the pipe. With this commit, Corpus::reset() leaves open pipes
alone.

This resolves #425.

### DIFF
--- a/src/common/filesystem.cpp
+++ b/src/common/filesystem.cpp
@@ -1,16 +1,26 @@
 #include "filesystem.h"
 
+#ifndef _MSC_VER
+// don't include these on Windows:
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <unistd.h>
+#endif
 
 namespace marian {
 namespace filesystem {
 
 bool is_fifo(char const* path) {
+#ifdef _MSC_VER
+  // Pretend that Windows knows no named pipes. It does, by the way, but
+  // they seem to be different from pipes on Unix / Linux. See
+  // https://docs.microsoft.com/en-us/windows/win32/ipc/named-pipes
+  return false;
+#else
   struct stat buf;
   stat(path, &buf);
   return S_ISFIFO(buf.st_mode);
+#endif
 }
 
 bool is_fifo(std::string const& path) {
--- a/src/common/filesystem.cpp
+++ b/src/common/filesystem.cpp
@@ -13,5 +13,9 @@ bool is_fifo(char const* path) {
   return S_ISFIFO(buf.st_mode);
 }
 
+bool is_fifo(std::string const& path) {
+  return is_fifo(path.c_str());
+}
+
 } // end of namespace marian::filesystem
 } // end of namespace marian
--- a/src/common/filesystem.cpp
+++ b/src/common/filesystem.cpp
@@ -1,0 +1,17 @@
+#include "filesystem.h"
+
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <unistd.h>
+
+namespace marian {
+namespace filesystem {
+
+bool is_fifo(char const* path) {
+  struct stat buf;
+  stat(path, &buf);
+  return S_ISFIFO(buf.st_mode);
+}
+
+} // end of namespace marian::filesystem
+} // end of namespace marian
--- a/src/data/corpus.cpp
+++ b/src/data/corpus.cpp
@@ -46,7 +46,6 @@ SentenceTuple Corpus::next() {
       }
       else {
         bool gotLine = io::getline(*files_[i], line);
-        // LOG(debug,"[{}][{}] {}", i, pos_ - 1, line);
         if(!gotLine) {
           eofsHit++;
           continue;
--- a/src/data/corpus.cpp
+++ b/src/data/corpus.cpp
@@ -90,6 +90,8 @@ void Corpus::shuffle() {
 void Corpus::reset() {
   corpusInRAM_.clear();
   ids_.clear();
+  if (pos_ == 0) // no data read yet
+    return;
   pos_ = 0;
   for (size_t i = 0; i < paths_.size(); ++i)
     {
@@ -98,9 +100,12 @@ void Corpus::reset() {
         // Probably not necessary, unless there are some buffers
         // that we want flushed.
       }
-      else if (!filesystem::is_fifo(paths_[i].c_str())) {
+      else {
+        ABORT_IF(files_[i] and filesystem::is_fifo(paths_[i]),
+                 "File '", paths_[i], "' is a pipe and cannot be re-opened.");
         // Do NOT reset named pipes; that closes them and triggers a SIGPIPE
-        // (lost pipe) at the writing end.
+        // (lost pipe) at the writing end, which may do whatever it wants
+        // in this situation.
         files_[i].reset(new io::InputFileStream(paths_[i]));
       }
     }
--- a/src/data/corpus.cpp
+++ b/src/data/corpus.cpp
@@ -92,15 +92,14 @@ void Corpus::reset() {
   if (pos_ == 0) // no data read yet
     return;
   pos_ = 0;
-  for (size_t i = 0; i < paths_.size(); ++i)
-    {
+  for (size_t i = 0; i < paths_.size(); ++i) {
       if(paths_[i] == "stdin") {
         files_[i].reset(new io::InputFileStream(std::cin));
         // Probably not necessary, unless there are some buffers
         // that we want flushed.
       }
       else {
-        ABORT_IF(files_[i] and filesystem::is_fifo(paths_[i]),
+        ABORT_IF(files_[i] && filesystem::is_fifo(paths_[i]),
                  "File '", paths_[i], "' is a pipe and cannot be re-opened.");
         // Do NOT reset named pipes; that closes them and triggers a SIGPIPE
         // (lost pipe) at the writing end, which may do whatever it wants
--- a/src/data/corpus.cpp
+++ b/src/data/corpus.cpp
@@ -4,6 +4,8 @@
 #include <random>
 
 #include "common/utils.h"
+#include "common/filesystem.h"
+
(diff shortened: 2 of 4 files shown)


frankseide:
I am not clear what this does. A corpus reset means to restart reading from the start of the data. But I am not sure how that is possible with a pipe.

ugermann:
The point is that Marian **unnecessarily** closes and reopens the input files **before** even processing the first sentence. And that breaks pipes, because the writing side is informed that the reader is gone and thus stops writing. 

- files are opened in the Corpus constructor.
- then they are closed (BatchGenerator::prepare(false) triggers a data reset before the first iteration) and opened again, and that's what I don't want.

In my scenario, I have a data sampler that deliver an infinite stream of training data to named pipes so that Marian trains in a single, infinite epoch until convergence or lack of improvement on the validation data. 

Another use case is training with in-memory shuffle with training data in multiple files 
```
marian -c config.yml -t <(cat *.de) <(cat *.en) 
```
without having to concatenate all the training data into a single file first.

Yet another use case is a set-up where you do not have enough local storage (e.g., in the cloud), so that you want to stream the training data from a remote location:
```
marian -c config.yml -t <(curl http:://data.server.com/train.de) <(curl http://data.server.com/train.en)
```

frankseide:
Thanks. I do not feel comfortable to base this decision on `is_pipe()`, since that seems an indirect dependency.

So with pipes, there are only two scenarios that work (correct me if I am wrong):
* `--no-shuffle` with infinite input: Pipe is written by an external program that keeps writing into the pipe forever. Marian stopping criterion decides when to stop. When Marian is done, pipe gets closed, and driving code will die with a pipe-lost error, but you don't care because training is complete. This variant invokes `reset()`, which causes the problem.
* `--no-shuffle` from finite file: Requires `close()`/`open()` in either case.
* `--shuffle-in-ram` with finite input from pipe or file: Marian reads all data from pipe upfront, and sees an EOF when done. After that, it will never open the pipe/file again, since it fetches data from RAM. This code never calls `reset()`, therefore there is no problem here.

Is this correct so far?

So maybe the problem is that the `CorpusBase` constructor opens the file in the first place. Is that actually needed? As I understand, at the start, either `reset()` or `shuffle()` is guaranteed to be called. So can we just move the initial file-opening step out from the constructor and into `shuffle()`? (`reset()` already re-opens the file anyway.) I.e. in the `CorpusBase` constructor, replace all `files_.emplace_back(new...)` with `files_push_back(nullptr)` (so that `files_` already has the right size, but no actual file)?

ugermann:
The Corpus constructor includes some sanity checks (e.g., for empty files). I'ts nice to detect those things before doing more costly things like model initialization, only to realize later that there are problems with the data. Another option would be to have Corpus::reset() reopen files only if pos_ > 0. If pos_ == 0, no data has been retrieved from the corpus, so there's no need to reopen files. Not sure if ids should be cleared in the code below();

```
void Corpus::reset() {
  // files_.clear(); // moved down; see below
  corpusInRAM_.clear();
  ids_.clear(); // move this too?

  if (pos_ == 0) // this is new: nothing to do if no data has been read yet. 
    return;
  files_.clear();
  pos_ = 0;
  for(auto& path : paths_) {
    if(path == "stdin")
      files_.emplace_back(new io::InputFileStream(std::cin));
    else
      files_.emplace_back(new io::InputFileStream(path));
  }
}
```

frankseide:
This looks quite right: If `pos_` is already 0, we are already in the desired state. If not, we make it so.

I think `ids_clear()` should be kept where it is. The `ids_` are owned by shuffle. A comment there says `reset()` and `shuffle()` are mutually exclusive. Praying that comment is correct, I read this as that whole function as doing two things:
* switch all variables to no-shuffle mode
* reopen the files

The `ids_clear()` is part of the first step, so it should remain before the check of `pos_`.

I think you can shorten that comment simply to

```
  if (pos_ == 0) // no data read yet
```

ugermann:
Btw, why does Corpus::reset() clear corpusInRAM? If I understand things correctly, corpusInRAM is used only when shuffling in RAM, but if shuffle is true, Corpus::reset() never gets called, so this is basically a no-op, no?

Not that it's a realistic scenario, but currently, reset will fail on finite data from pipes when training for more than one epoch without shuffling, because the data won't be cached, and pipes can't be re-opened. Obviously, in that scenario (training for more than one epoch on finite data), one would want to shuffle, but technically, such scientifically unwise user behavior should not cause Marian to crash.

One  hack might be to use --shuffle-in-ram to tell Marian to keep all data in ram and --no-shuffle to tell it not to actually shuffle. In this case, Corpus::reset() should not clear corpusInRAM.

frankseide:
We could replace the clearing of `corpusInRAM` with an `ABORT_IF(!corpusInRAM.empty(), "...")`.

I think I added this out of completeness--when `reset()` is called, that switches shuffle mode off. I am not entirely happy with the API where one calls either `shuffle()` or `reset()`. It would be better to have a single API function that knows what to do, and not let the caller even have to know about it.

`corpusInRAM` was written for shuffling, since, unless one uses the pipe method, one does not need to keep it in RAM, since it is cheap to reread from disk. If you feel that is a worthwhile feature, may I suggest to separate this from the present PR?

ugermann:
Pinging this PR. Are there any obstacles to merging this?

emjotde:
We are working through the PRs now. Give us a few more days. I would like to clean these out during the Marathon at the latest.

emjotde:
Looking through this now.

src/common/filesystem.cpp
```diff
@@ -0,0 +1,21 @@
+#include "filesystem.h"
+
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <unistd.h>
```
emjotde: This would fail on Windows, wouldn't it? Assuming that we are not going to use pipes on Windows, can you surround that with appropriate #ifdefs and make sure things work with the compilation of these parts turned off?

src/data/corpus.cpp
```diff
@@ -85,16 +88,27 @@ void Corpus::shuffle() {
...
+    return;
   pos_ = 0;
-  for(auto& path : paths_) {
-    if(path == "stdin")
-      files_.emplace_back(new io::InputFileStream(std::cin));
-    else
-      files_.emplace_back(new io::InputFileStream(path));
-  }
+  for (size_t i = 0; i < paths_.size(); ++i)
+    {
```
emjotde: Nitpicking: we don't do newlines here.

src/data/corpus.cpp
```diff
@@ -85,16 +88,27 @@ void Corpus::shuffle() {
...
+      if(paths_[i] == "stdin") {
+        files_[i].reset(new io::InputFileStream(std::cin));
+        // Probably not necessary, unless there are some buffers
+        // that we want flushed.
+      }
+      else {
+        ABORT_IF(files_[i] and filesystem::is_fifo(paths_[i]),
```
emjotde: "and" is not C++ standard, if I am not wrong. gcc allows this for some reason. Many other compilers don't or at least didn't a few years ago.

src/common/filesystem.cpp
```diff
@@ -0,0 +1,21 @@
...
+namespace filesystem {
+
+bool is_fifo(char const* path) {
+  struct stat buf;
+  stat(path, &buf);
+  return S_ISFIFO(buf.st_mode);
+}
+
+bool is_fifo(std::string const& path) {
+  return is_fifo(path.c_str());
```
emjotde: Probably just return "false" on Windows?

src/data/corpus.cpp
```diff
@@ -44,6 +46,7 @@ SentenceTuple Corpus::next() {
       }
       else {
         bool gotLine = io::getline(*files_[i], line);
+        // LOG(debug,"[{}][{}] {}", i, pos_ - 1, line);
```
emjotde: Delete

src/common/filesystem.cpp
emjotde: We seem to be using "#ifdef _MSC_VER" for that purpose. @frankseide is that the preferred way?

src/data/corpus.cpp
ugermann: Resolving as I fix things.

src/data/corpus.cpp
ugermann: Resolving as I fix things.

src/data/corpus.cpp
ugermann: Resolving as I fix things.

src/common/filesystem.cpp
ugermann: Resolving as I fix things.

src/common/filesystem.cpp
ugermann: Resolving as I fix things.

ugermann:
Code changes made as requested.

emjotde:
@frankseide can you take a quick look if handling Windows here is OK to you?

frankseide:
Yes, it seems OK to me to ignore named pipes on Windows for now, as done in this PR. If we ever need this, we can implement it later.

emjotde:
Thanks Uli and Frank.

PR #442, opened 2019-04-27
Save models before exiting after receiving signal SIGTERM

ugermann:
Some of us run Marian on a cluster with time limits for jobs. When a training job gets killed due to timeout, we lose all the progress since the last checkpoint save (--save-freq=...). This branch / pull request installs a signal handler in class Scheduler that triggers a model save before exit after receiving a SIGTERM (i.e., the default signal for kill <pid>), so that we can pick up where we left it after a training process was killed.

I've tested it to some degree but not exhaustively. Feedback welcome.


snukky:
Thanks, I'll take a look. Do regression tests pass with your branch?

ugermann:
How do I run them?

snukky:
Please follow https://github.com/marian-nmt/marian-regression-tests/blob/master/README.md

ugermann:
```
Skipped:
  - tests/training/features/exp-smoothing/test_expsmooth_sync.sh
  - tests/training/restoring/optimizer/test_adam_params_async.sh
  - tests/training/restoring/optimizer/test_adam_params_sync.sh
  - tests/training/restoring/exp-smoothing/test_expsmooth.sh
  - tests/training/restoring/exp-smoothing/test_expsmooth_s2s.sh
  - tests/training/restoring/exp-smoothing/test_expsmooth_sync.sh
  - tests/training/restoring/multi-gpu/test_async.sh
  - tests/training/restoring/multi-gpu/test_sync.sh
  - tests/training/multi-gpu/test_async_sgd_runs.sh
  - tests/training/multi-gpu/test_sync_sgd.sh
  - tests/examples/iris/test_iris.sh
  - tests/examples/mnist/test_mnist_ffnn.sh
  - tests/server/test_ende.sh
  - tests/server/test_ende_align.sh
  - tests/server/test_ende_batch32.sh
  - tests/server/test_ende_cpu.sh
  - tests/server/test_ende_with_empty_lines.sh
---------------------
Ran 157 tests in 00:31:58.166s, 140 passed, 17 skipped, 0 failed
```

emjotde:
I assume this leaves SIGINT alone?

emjotde:
This has surprisingly many changes that seem unrelated to the SIGTERM issue, no?

snukky:
Maybe thre should be an option to turn it off (or on), as sometimes automatic saving might be an unwanted behaviour?

ugermann:
@emjotde: In order to understand what was going on, I refactored the code a bit ti make it more readable, apart form mere reformatting of some lambda functions, I moved blocks of code into separate auxiliary functions (e.g., initMPI, prepareData) to declutter the functions that originally contained those code blocks. And yes, this leaves SIGINT alone.

@snukky Making this configurable is trivial, but I'd prefer to have #488 merged first before I make this configurable. In principle, SIGTERM is intended for shutdown with cleanup, and SIGKILL for immediate exit without cleanup.

frankseide:
Similar changes may have been made in our internal code. @emjotde, are the changes re variable-sized batches already in public master?

ugermann:
Can we pretty-please merge this soon? It's been more than four months since the original filing.

emjotde:
I'm merging in our internal branch now. Will take another look at this afterwards. I am a bit wary because of the changes not related to SIGTERM.

emjotde:
Looking also at #488

ugermann:
I've minimized the differences between master and this branch. Can this be merged, please? This pull request has been waiting for Godot for more than half a year now.

emjotde:
Will take a look tomorrow.

src/training/communicator.h
```diff
@@ -175,7 +175,17 @@ class DefaultCommunicator : public ICommunicator {
...
+    auto reset = [this, shardSize](size_t idx, size_t begin, size_t end) {
+      auto grad = graphs_[idx]->params()->grads();
+      if (begin > 0)
+        grad->subtensor(0, begin)->set(0);
+      if (end < grad->size())
+        grad->subtensor(end, grad->size()-end)->set(0);
+    };
+
     foreach(scatter);
+    foreach(reset);
```
emjotde: Where does this change come from?

ugermann: By cherry-picking communicator.h from master in the latest merge with master. The change comes from 
https://github.com/marian-nmt/marian-dev/commit/2111c280362fcf1b76bb13688e5a8f372e435479

ugermann: I actually don't understand why github lists this as a difference.

emjotde: I have been running in to this recently. Seems merge diffs are done against last common ancestor or something. Makes no sense to me either. At least give me a way to see an actual 3-way diff, not a fictional one.

emjotde:
Running full regressions tests now, and playing around with it.

src/training/scheduler.cpp
```diff
@@ -0,0 +1,32 @@
+#include "scheduler.h"
+#include <signal.h>
+#include <cassert>
+
+namespace marian {
+bool Scheduler::sigterm_{false};
+
+void
+Scheduler::
+signalHandler_(int sig) {
```
emjotde: Odd formatting, please keep this in one line.

src/training/scheduler.cpp
```diff
@@ -0,0 +1,32 @@
...
+    break;
+  default:
+    ABORT("No action defined for signal {}", sig);
+  }
+}
+
+
+void
+Scheduler::
+installSignalHandlers_() {
```
emjotde: As above.

src/training/scheduler.cpp
```diff
@@ -0,0 +1,32 @@
...
+
+void
+Scheduler::
+signalHandler_(int sig) {
+  // Note: sys_siglist[sig] or stdsignal() describe the effect (e.g.,
+  // 'Terminated' rather than provide the signal name (which are #define(s)
+  // in signal.h), so we have to do custom log messages here.
+  switch (sig) {
+  case SIGTERM: // save models and exit
+    LOG(info, "[training] Scheduler received signal SIGTERM");
```
emjotde: indent case

src/training/scheduler.h
```diff
@@ -19,6 +19,10 @@ class Scheduler : public TrainingObserver {
   timer::Timer timer_;
   timer::Timer heartBeatTimer_;
 
+  static bool sigterm_;
+  void installSignalHandlers_();
+  static void signalHandler_(int sig);
```
emjotde: Please add comments to all of these.

src/training/scheduler.h
```diff
@@ -146,9 +150,14 @@ class Scheduler : public TrainingObserver {
       : options_(options), state_(state) {
     ABORT_IF(state_->factor != 1, "state.factor unexpectedly not 1 at this point??");
     updateLearningRate(*state);
+    installSignalHandlers_();
```
emjotde: This seems like an odd place to put that and rather not well motived. @snukky can you suggest a better location?

src/training/scheduler.h
emjotde: Also, we do not use _ for function names.

src/training/scheduler.h
```diff
@@ -175,7 +184,13 @@ class Scheduler : public TrainingObserver {
   }
 
   void started() { LOG(info, "Training started"); }
-  void finished() { LOG(info, "Training finished"); }
+  void finished() {
+    if (keepGoing(false)) // false means: ignore sigterm flag
```
emjotde: ```keepGoing(/*checkForSigTerm=*/false)```

src/training/scheduler.h
```diff
@@ -204,7 +219,8 @@ class Scheduler : public TrainingObserver {
                 bool final = false) {
     // Do not validate if already validated (for instance, after the model is
     // loaded) or if validation is scheduled for another update
-    if(state_->validated
+    if(sigterm_
```
emjotde: Add comment how this is affected by sigterm_

emjotde:
General functionality is nice. Have a few requests for changes. Will run regression tests afterwards once more.

emjotde:
Regression tests all pass.

src/training/scheduler.cpp
```diff
@@ -0,0 +1,32 @@
...
+    ABORT("No action defined for signal {}", sig);
+  }
+}
+
+
+void
+Scheduler::
+installSignalHandlers_() {
+  // TODO: use sigaction instead of signal
+  signal(SIGTERM, Scheduler::signalHandler_);
```
snukky: Why `sigaction` would be better than `signal` and what is the reason it's only a TODO? A quick googling suggests that `sigaction` is not supported on Windows?

src/training/scheduler.h
snukky: Is this the only use of `keepGoing(false)`, i.e. only for logging a different message? If so, I would suggest removing the argument from `keepGoing()` (and keeping it simple) and adding here `if(sigterm_)` or similar.

src/training/scheduler.h
snukky: It's in the constructor, so should be OK.

emjotde: Actually, ignore that comment. GitHub cut off the function name so weirdly that I thought it was inside getScheduledLRDecayFactor().

emjotde: The constructor is fine.

src/training/scheduler.cpp
ugermann: > Why sigaction would be better than signal
https://stackoverflow.com/questions/231912/what-is-the-difference-between-sigaction-and-signal

> what is the reason it's only a TODO
Convenience of implementation. Old dogs, new tricks and so forth.

src/training/scheduler.h
ugermann: It is possible that both are true: SIGTERM recevied and training finished. In that case, we want to log "Training finished.", not "Training interrupted.", so we'd have to say
```
if(sigterm&&!keepGoing()){
...
```
And that's the opposite of "keeping things simple".

emjotde:
Let us know when you are ready.

ugermann:
Compiles now. Didn't run regtests b/c I get different results on different machines.

emjotde:
OK, will try again.

emjotde:
Not compiling for me, did you push?

emjotde:
```
/home/marcinjd/marian-dev/src/training/scheduler.cpp: In member function ‘void marian::Scheduler::installSignalHandlers()’:
/home/marcinjd/marian-dev/src/training/scheduler.cpp:25:19: error: ‘signalHandler_’ is not a member of ‘marian::Scheduler’
   signal(SIGTERM, Scheduler::signalHandler_);
```

ugermann:
Weird. Did compile even with make -B. Fixed bug now, methinks. Try again.

emjotde:
Looks good, running tests.

emjotde:
I am satisfied. Regression tests pass. @snukky, your turn.

src/training/scheduler.h
snukky: OK, but `keepGoing()` is a part of the public interface of the class, so I would value its simpler version more than a simpler if-statement hidden in a function (plus a comment). It also seems the argument is used only internally inside the class.

src/training/scheduler.cpp
snukky: OK, would you mind adding this link into that TODO for futher reference? Thanks

src/training/scheduler.h
emjotde: I think @snukky is right here. If that's only for logging purposes just checking `sigterm_` seems fine.

PR #458, opened 2019-06-28
Allow allocator to increase the workspace size by more than step_

mateuszchudyk:
For now, if the allocator receives a memory allocation request that doesn't fit to any gap then the workspace is increased by step_. This procedure is repeated until an enough gap occurs so in case of big allocation, the allocator performs a series of workspace reallocations instead of one big reallocation. This commit change policy of allocator in the following way: if an allocation doesn't fit to any gap then the allocator increases workspace size by max(step_, allocation_size).

UPDATE: I forgot about a minor optimization I made. Workspace is increased by max(step_, allocation_size - last_gap_size).

### DIFF
--- a/src/tensors/allocator.h
+++ b/src/tensors/allocator.h
@@ -127,6 +127,10 @@ private:
         throw AllocationException(available_, size);
       }
 
+      // Extend the workspace to fit requested allocation. To avoid lots of small
+      // reallocations, the workspace is extended by at least step_.
+
+      // Find size of the last gap
       size_t last_gap_size = 0;
       for (const auto& gap : gaps_) {
         if (gap.data() + gap.size() == device_->data() + device_->size()) {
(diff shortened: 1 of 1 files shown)


src/tensors/allocator.h
```diff
@@ -122,13 +122,21 @@ class Allocator {
...
 
-    if(throw_ && it == gaps_.end()) {
-      throw AllocationException(available_, size);
-    }
+    if (it == gaps_.end()) {
+      if (throw_) {
+        throw AllocationException(available_, size);
+      }
+
+      size_t last_gap_size = 0;
```
emjotde: Please add comments what exactly this is doing and how this is different to before. It took me a while to figure this out. From what I am seeing it either grows step size or required minimal size, depending what is larger. 

Otherwise a good change probably.

PR #459, opened 2019-07-03
Abort if shapes aren't the same in reshape() or concatenate()

fiqas:
I just added to ABORT_IF statements in operators to avoid confusing myself anymore when I try to modify Expr in a wrong way and I catch it much later.
No idea how Marian behaves when trying reshaping/concatenating with different elements/shapes, probably takes a chunk of memory it shouldn't and calculates with it.

### DIFF
--- a/src/graph/expression_operators.cpp
+++ b/src/graph/expression_operators.cpp
@@ -185,6 +185,9 @@ Expr operator/(float a, Expr b) {
 /*********************************************************/
 
 Expr concatenate(const std::vector<Expr>& concats, int ax) {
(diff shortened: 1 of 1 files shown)


emjotde:
The checks aren't correct though. One can concatenate tensors, but specific dims have to match, for instance [3, 5] + [1, 5] for axis 1 is [4, 5]. Dimension of axis 2 (here 5) has to match.

emjotde:
It's correct for reshape.

emjotde:
Closing as I am implementing this right now.

PR #474, opened 2019-08-11
Some updates - based on PR 413, reviewed earlier

emjotde:
We already reviewed and approved these changes in #413 but for some reason only merged into a separate branch. Catching up. Waiting for regression-tests to go through.


emjotde:
@snukky Can you make sure these changes fully overlap with PR #413 ?

emjotde:
All regression-tests pass.

PR #478, opened 2019-08-12
Fix CMakeLists.txt and bug with overlapping CXX flags

### LINKED ISSUE #477: spm_train built with marian compilation crashes
```
git clone https://github.com/marian-dev 
cd marian-dev
mkdir build
cd build
cmake -DUSE_SENTENCEPIECE=on ..
make -j
./spm_train -h
```
results in a segmentation fault.

If I build sentencepiece directly from the submodule like so:
```
mkdir build.sp
cd build.sp
cmake ../src/3rd_party/sentencepiece
make -j
src/spm_train -h
```

spm_train works as expected.

So something is wrong with the way the marian cmake setup compiles sentencepiece.

emjotde:
Fix CMakeLists.txt and src/3rd_party/CMakeLists.txt, bug with overlapping CXX flags 

* Apparently CXX_FLAGS and CXX_FLAGS_RELEASE are concatenated. Hence CXX_FLAGS should not contain flags for optimization. Moved out into CXX_FLAGS_RELEASE.
* If -DCMAKE_BUILD_TYPE is not specified, assume "Release"
* Partially fixes #477. The segfault is a bug in SentencePiece or Protobuf, but does not seem to occur when linking shared libs. Still occurs with static linking, but now warns about the segfault.


emjotde:
Ping @snukky @ugermann ?

PR #481, opened 2019-08-14
Use relative paths in decoder config files

### LINKED ISSUE #460: Model yaml files do not have relative paths
When training a model, the generated yaml files (e.g. `model.npz.decoder.yml`) do not contain relative paths, despite invoking `marian` with `--relative-paths`. The `relative-paths` entry in those yaml files is also `false`.

The version used is the newest master in marian-dev (84063a8475e51c2fedd604a2aecf0e79c528963e). It did not work either with an older version I was using (`v1.7.6 b6d7c56 2018-12-12 17:30:18 -0800`)

Note that `--relative-paths` was observed in the config file generated with `marian --relative-paths --dump-config ...` but not in the model yaml files generated by the training with the previously generated configuration file with `marian -c config.yml --relative-paths ...`.

snukky:
This fixes #460. 

If `--relative-paths` is used, model and vocab paths in `marian.npz.*{decoder,amun}.yml` will be relative.

### DIFF
--- a/src/models/amun.h
+++ b/src/models/amun.h
@@ -185,16 +185,26 @@ private:
   void createAmunConfig(const std::string& name) {
     Config::YamlNode amun;
     auto vocabs = options_->get<std::vector<std::string>>("vocabs");
-    amun["source-vocab"] = vocabs[0];
-    amun["target-vocab"] = vocabs[1];
+
+    if(options_->get<bool>("relative-paths")) {
+      amun["relative-paths"] = true;
+      auto dirPath = filesystem::Path{name}.parentPath();
+      amun["source-vocab"] = filesystem::relative(filesystem::Path{vocabs[0]}, dirPath).string();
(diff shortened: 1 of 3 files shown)


src/models/amun.h
```diff
@@ -185,17 +185,27 @@ class Amun : public EncoderDecoder {
...
+      amun["relative-paths"] = false;
+      amun["source-vocab"] = vocabs[0];
+      amun["target-vocab"] = vocabs[1];
+      amun["scorers"]["F0"]["path"] = name;
+    }
+
     amun["scorers"]["F0"]["type"] = "Nematus";
     amun["weights"]["F0"] = 1.0f;
 
+    amun["devices"] = options_->get<std::vector<size_t>>("devices");
```
emjotde: Let's not set devices. I was thinking about that, but that's not really a model setting.

snukky: Removed.

PR #487, opened 2019-08-15
Early stopping on multiple validation metrics

Proyag:
Following a discussion with @snukky , this should let us early stop on multiple validation metrics.
* For example `--validation-metrics cross-entropy bleu --early-stopping 10 20`
* The previous behaviour is still the default, and specifying fewer arguments for `--early-stopping` will only apply it to the corresponding metrics in order: `--validation-metrics cross-entropy bleu --early-stopping 10` is the same as before.
* We can use `0` to not early stop on one of the metrics: `--validation-metrics cross-entropy bleu translation --early-stopping 10 0 20`

### DIFF
--- a/src/common/config_parser.cpp
+++ b/src/common/config_parser.cpp
@@ -437,9 +437,10 @@ void ConfigParser::addOptionsValidation(cli::CLIWrapper& cli) {
       "Metric to use during validation: cross-entropy, ce-mean-words, perplexity, valid-script, "
       "translation, bleu, bleu-detok. Multiple metrics can be specified",
       {"cross-entropy"});
-  cli.add<size_t>("--early-stopping",
-     "Stop if the first validation metric does not improve for  arg  consecutive validation steps",
-     10);
+  cli.add<std::vector<size_t>>("--early-stopping",
+     "Stop if the corresponding validation metrics do not improve for  arg  consecutive validation steps. "
+     "At most one per metric",
+     {10});
 
   // decoding options
   cli.add<size_t>("--beam-size,-b",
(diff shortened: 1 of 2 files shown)


emjotde:
Hi, Proyag,
Does this break backwards-compatibility of config files?

snukky:
@Proyag The backward-compatibility can be checked with our regression tests. 

I think this feature may also require some modifications in loading/saving of [training state](https://github.com/marian-nmt/marian-dev/blob/master/src/training/training_state.h).

frankseide:
"The backward-compatibility can be checked with our regression tests."

No, this is not good enough. It is unrealistic to expect regression tests to completely cover all use cases. If they fail, then obviously something is wrong, but if they pass, we cannot assume that the other use cases were not affected.

Please analyze the code in how far it affects back compat.

emjotde:
Basically test if a config containing
```
early-stopping: 10
```
works. 

I think I have code in there that upcasts single values to lists if lists are expected. We did that for backwards-compatibility of `--lr-warmup M N`.

Proyag:
I already checked that all regression tests passed.
I also took a random model.npz.yml from the passed tests, changed
```
early-stopping:
    - 10
```
to
```
early-stopping: 10
```
and `marian -c model.npz.yml` still works, and in fact, it converts and displays it like the first case. Looks like you already solved this problem :)

emjotde:
@snukky let's review this.

emjotde:
@snukky ping?

snukky:
We were thinking that this could shorten trainings for which the validation cross-entropy continues improving for a very long time, while BLEU stalles for dozens of epochs. The correct solution was better cleaning/filtering of development and training data, which resulted in better BLEU and shorter training.

I am closing the PR. @Proyag Please feel free to complain and re-open if you think that this is still useful.

PR #489, opened 2019-08-20
filter out beams in trie for paraphrasing

PinzhenChen:
Now supports option "--paraphrase". works well on >=48 beams. @XapaJIaMnu

### DIFF
--- a/src/common/config_parser.cpp
+++ b/src/common/config_parser.cpp
@@ -512,6 +512,9 @@ void ConfigParser::addOptionsTranslation(cli::CLIWrapper& cli) {
   cli.add<std::string>("--trie-pruning-path",
      "Use trie pruning during translation. Provide path to the monolingual corpora.")
     ->implicit_val("");
+  cli.add<std::string>("--paraphrase",
+     "Paraphrase based on trie sentences through NMT as a pivot. Can only be used in trie pruning mode (--trie-pruning-path).")
+    ->implicit_val("true");
 
   addSuboptionsDevices(cli);
   addSuboptionsInputLength(cli);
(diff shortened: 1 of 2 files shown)


src/translator/beam_search.h
```diff
@@ -373,8 +373,11 @@ class BeamSearch {
                      first,
                      batch);
 
-      if (triePrune_) {
-         beams = filterForContinuations(beams, paraphrase_);
+      if (!first || !paraphrase_){ // if first word and paraphrasing, allow words in trie
```
XapaJIaMnu: The comment says "and"  but you' have an or? 

Also you have an inner if statement `if (triePrune_)` that has no else. You should move it to the parent.

src/common/config_parser.cpp
```diff
@@ -512,6 +512,9 @@ void ConfigParser::addOptionsTranslation(cli::CLIWrapper& cli) {
   cli.add<std::string>("--trie-pruning-path",
      "Use trie pruning during translation. Provide path to the monolingual corpora.")
     ->implicit_val("");
+  cli.add<std::string>("--paraphrase",
```
XapaJIaMnu: We should add a check to make sure that there is an error if `paraphrase` is enabled. This could be done around here: https://github.com/marian-nmt/marian-dev/blob/trieme_CPUonly/src/translator/translator.h#L52

But I can add it later. It's not critical since this code ain't production anyways.

src/translator/beam_search.h
PinzhenChen: Comment is the double negation version of the if-statement. I thought using `or` is faster..

Sure I changed that for readability.

kpu:
I don't think this makes much sense.  Are you actually going to have a trie of paraphrases as opposed to a lattice?  And chances are that your n-best list will have at least one entry not found in a pre-generated list of sentences.

src/translator/beam_search.h
XapaJIaMnu: Suuuuure, but the branch predictor would learn those anyways. Also it's not a performance critical path.

PR #505, opened 2019-09-19
Enable multi-source input in marian-server

tomekd:
At the moment, marian-server could not handle multi-source models. This pull request aims to fix this problem. The solution assumes that the source inputs are separated by `\t`.

Additionally, I noticed that marian-server does not follows `max-length` limitations. I fixed it too.


emjotde:
@snukky?

src/common/utils.cpp
```diff
@@ -93,6 +94,26 @@ std::string join(const std::vector<std::string>& words, const std::string& del /
   return ss.str();
 }
 
+std::vector<std::string> tsv2lists(const std::string& inputText, int inputNum) {
+    std::string line_;
+    std::vector<std::vector<std::string>> inputLists(inputNum);
+    std::istringstream inputStream(inputText);
+    while (std::getline(inputStream, line_)) {
+      auto items = marian::utils::splitAny(line_, "\t");
+      std::cerr << "Split into " << items.size() << std::endl;
```
snukky: Please remove the debug.

src/common/utils.cpp
```diff
@@ -93,6 +94,26 @@ std::string join(const std::vector<std::string>& words, const std::string& del /
   return ss.str();
 }
 
+std::vector<std::string> tsv2lists(const std::string& inputText, int inputNum) {
+    std::string line_;
```
snukky: Please remove the trailing underscore from a local variable.

src/common/utils.cpp
```diff
@@ -93,6 +94,26 @@ std::string join(const std::vector<std::string>& words, const std::string& del /
...
+    }
+
+    std::vector<std::string> inputs;
+    for (auto &inputList : inputLists) {
+      inputs.push_back(marian::utils::join(inputList, "\n"));
+    }
+    return inputs;
+}
+
+
```
snukky: Double whitespace.

src/data/text_input.cpp
```diff
@@ -43,6 +46,9 @@ SentenceTuple TextInput::next() {
     std::string line;
     if(io::getline(dummyStream, line)) {
       Words words = vocabs_[i]->encode(line, /*addEOS =*/ true, /*inference =*/ inference_);
+      if(this->maxLengthCrop_ && words.size() > this->maxLength_) {
+        words.resize(maxLength_);
+      }
       if(words.empty())
         words.push_back(Word::ZERO); // @TODO: What is this for? @BUGBUG: addEOS=true, so this can never happen, right?
```
snukky: I think the EOS token needs to be added after cropping the sentence, similar to https://github.com/marian-nmt/marian-dev/blob/master/src/data/corpus_base.cpp#L211.
Then that comment with TODO can be removed.

src/models/model_task.h
```diff
@@ -10,5 +13,9 @@ struct ModelTask {
 
 struct ModelServiceTask {
   virtual std::string run(const std::string&) = 0;
+  virtual std::string run(const std::vector<std::string>&) {
+    ABORT("Not implemented");
```
snukky: Why not a pure virtual function?

src/common/utils.cpp
```diff
@@ -93,6 +94,26 @@ std::string join(const std::vector<std::string>& words, const std::string& del /
   return ss.str();
 }
 
+std::vector<std::string> tsv2lists(const std::string& inputText, int inputNum) {
```
snukky: Please add a comment documenting the function, ideally with an example input and output.

src/common/utils.cpp
```diff
@@ -93,6 +94,26 @@ std::string join(const std::vector<std::string>& words, const std::string& del /
...
+      std::cerr << "Split into " << items.size() << std::endl;
+      for (size_t i = 0; i < items.size(); ++i) {
+        inputLists[i].push_back(items[i]);
+      }
+    }
+
+    std::vector<std::string> inputs;
+    for (auto &inputList : inputLists) {
+      inputs.push_back(marian::utils::join(inputList, "\n"));
+    }
```
snukky: Add a comment why this part is needed.

src/command/marian_server.cpp
```diff
@@ -14,22 +14,29 @@ int main(int argc, char **argv) {
...
     // Translate
     timer::Timer timer;
-    auto outputText = task->run(inputText);
+    std::vector<std::string> inputs;
+    if (inputNum >= 2) {
+      inputs = utils::tsv2lists(inputText, inputNum);
+    } else {
+      inputs = std::vector<std::string>({inputText});
+    }
+    auto outputText = task->run(inputs);
```
snukky: With this logic `std::string run(std::string)` is used nowhere, am I right? So either this version of the function can be removed, or the logic here can be something similar to this:
```
std::string o;
if(inputNum >= 2) {
  o = task->run(utils::tsv2lists(inputText, inputNum));
} else {
  o = task->run(inputText);
}
```
I think the latter is cleaner, but feel free to object.

snukky:
I will revisit this soon after merging #617.

snukky:
It seems that GitHub got confused after merging with the current master. I will open a new pull request that replaces this one.

PR #510, opened 2019-10-08
Support for C++17 and SSSE4.2

alvations:


emjotde:
Sorry, we not going to kill off backward compatibility with this.

alvations:
Totally understand, but just listing some patches for anyone who's using cpp 17. 

But somehow that `make_unique()` checking code and the finding ssse4.2 should work even with cpp11.

alvations:
Or should I make another PR that's backcompatible with the other patches but don't specify the CXX std version?

emjotde:
Maybe we can cherrypick something, but letting it hang here first. As for performance improvments, there is a major update coming that will help a lot in that regard.

alvations:
Ah performance wise, we made changes to the `translator/history.h` using c++17 features but I guess we should patch that after the major update on your end.

emjotde:
We got rid of all locking shared-pointers in performance critical parts. Might just make your stuff go away.

alvations:
BTW, for anyone having a nasty CUBLAS problem during installation with CUDA 10.0 or 10.1 (esp. native docker images), 


```
# Remove any cmake on system.
sudo apt remove -y cmake cmake-data

# Install new cmake.
wget https://github.com/Kitware/CMake/releases/download/v3.14.7/cmake-3.14.7-Linux-x86_64.sh
mv cmake-3.14.7-Linux-x86_64.sh /opt
chmod +x /opt/cmake-3.14.7-Linux-x86_64.sh
sudo bash /opt/cmake-3.14.7-Linux-x86_64.sh

# Create symbolic link for new cmake binaries to /usr/local/bin
ln -s /opt/cmake-3.14.7-Linux-x86_64.sh/bin/* /usr/loca/bin
```

https://github.com/marian-nmt/marian-dev/blob/master/CMakeLists.txt#L1 is seriously outdated ;P

emjotde:
Hm, worked fine for me with cmake 3.5 even. But I only use custom installs of CUDA and cmake. Now using 3.15 works fine. BTW, some of the performance upgrades in current master of marian-dev. Maybe already noticeable?

alvations:
That's like a tonne of commits since 20+ days ago... Migrating our code to latest master is going to take some time but nevertheless the ssse42 patch here and removing the "global" check of if __GNU__ in place for just looking for the make_unique function should work =)

emjotde:
Maybe wait. More's coming. We are getting ready for version 1.9

emjotde:
I'm closing this PR, we cannot really pull it in.

alvations:
https://github.com/marian-nmt/marian-dev/commit/a73a5e404ccdd5a17f09cd357530c7e35d75d9b4 removed the `make_unique` =) 

BTW, when the `make_unique` resurfaced, some bits of the proper cpp std function checks would have been a lot better than  blanket `#ifdef __GNUC__` checks

ZJaume:
Hi,
Regarding the @alvations comment about using a newer version of cmake, I found another way to do it with python and a virtual environment. I think it is way that could bring less problems, or could be useful if the user has no root permissions.
```
python3 -m venv venv
source venv/bin/activate
pip install cmake==3.13.3
```

PR #513, opened 2019-10-21
fix Windows MSVC builds

robberlang:
The changes here fix errors in Windows builds using the MSVC compiler.
The MSVC builds fail because of compiler warnings occuring and being treated as errors, because of CUDA libraries not being found (in the case of `USE_STATIC_LIBS=on`), and Debug builds can fail because of a runtime library mismatch with the NVCC-compiled object files, and finally builds can fail because of SentencePiece build flags being forced.

There are several C4100 ('identifier' : unreferenced formal parameter) warnings which arise from preprocessor directives. The change here is to disable this warning.
The code change in `src/common/options.cpp` is for MSVC warning C4101: 'identifier' : unreferenced local variable.

The changes for CUDA library finds with `USE_STATIC_LIBS` include directories of a CUDA toolkit install on Windows as well as an acceptance of not finding the cuLIBOS library (which doesn't seem to exist in Windows CUDA toolkit installs). An alternative to what I did would be to not look for it at all on Windows builds, but I have not tested this with all supported CUDA versions (only 10.1).

There are issues with building SentencePiece shared libraries on Windows, and these changes make it possible to build and use static SentencePiece libraries even if `USE_STATIC_LIBS` is off (by setting `SPM_ENABLE_SHARED` to `off`).

One more change here is to add an option `USE_MKL` to allow for not using MKL even when it is installed/found.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,6 +22,7 @@ option(COMPILE_EXAMPLES "Compile examples" OFF)
 option(COMPILE_TESTS "Compile tests" OFF)
 option(COMPILE_SERVER "Compile marian-server" ON)
 option(USE_FBGEMM "Use FBGEMM" ON)
+option(USE_MKL "Use MKL" ON)
 
 # Project versioning
 find_package(Git QUIET)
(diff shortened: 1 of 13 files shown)


frankseide:
Hi, thanks for looking into this!

To eliminate a C4100 unreferenced-variable warning, we should not disable the warning. That will bloat the code by at least 10 lines (pragma push/pop, guarded by MSC_VER).

Instead, the easy fix is to reference them, e.g.

x;

Or, if it is a parameter to an overridden function that is just not used in a specific implementations, just put the identifier name in comments in the function definition.

Thanks,

Frank

robberlang:
The change I made initially was to set a compiler flag to disable the warnings, but you are right. They are easy to fix and there are not as many as I first feared! So I've done that and have now also made code changes for the `C4702: unreachable code` warnings which came in the Debug builds, plus I dealt with a few linker warnings:
```
LNK4044: unrecognized option '/Wl,--start-group'; ignored
LNK4044: unrecognized option '/Wl,--end-group'; ignored
LNK4075: ignoring '/INCREMENTAL' due to '/LTCG' specification
LNK4098: defaultlib 'MSVCRTD' conflicts with use of other libs; use /NODEFAULTLIB:library
```

snukky:
Merged with https://github.com/marian-nmt/marian-dev/pull/677

PR #525, opened 2019-10-29
Use ccache for faster compilation if available.

ugermann:
See https://ccache.dev/ for explanations.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,6 +10,12 @@ set(CMAKE_CXX_STANDARD 11)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(BUILD_ARCH native CACHE STRING "Compile for this CPU architecture.")
 
+-# use ccache for faster compilation if available
+-find_program(CCACHE_PROGRAM ccache)
+-if(CCACHE_PROGRAM)
+-  set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE "${CCACHE_PROGRAM}")
+-endif()
+
 # Custom CMake options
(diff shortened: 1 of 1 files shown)


ugermann:
Oops. That was indeed a copy-and-paste artefact from a git diff. Fixed.

emjotde:
Looking at the website, seems to me that should only be explicitly enabled, not implicitly due to availability. There should be no nasty surprises to people. I would also output a message in CMake that it is being used whenever its being used.

emjotde:
Also, any word on nvcc support? That would actually be more useful as that takes longer than normal gcc.

emjotde:
Seems to work for nvcc too, not too bad. Gets tripped up by the custom makefile for NCCL since that one is not using CMake.

emjotde:
@frankseide look at that. This is pretty cool.

frankseide:
I have the largest build times when I switch branches. Does it handle that case correctly?

emjotde:
Looks pretty good. If it wasn't for the NCCL-build, it seems it would do things a lot faster, even after a "make clean" in a new branch.

ugermann:
I've added a cmake option -DUSE_CCACHE (default is off) to allow users to request use of ccache. Ccache is only useful when developers recompile often; the first compile is slower because the cache needs to be built. Also, cmake now issues a mesage when ccache is requested and found / not found.

emjotde:
On a side note, might be worth taking a look if NCCL can be convinced to work with this.

ugermann:
@snukky For the record: ccache should be installed on all Valhalla machines.

PR #546, opened 2019-11-06
Added ssse4.2 support

alvations:
Still even if newer cmake and c++ version isn't supported, ssse4.2 support is still a good thing. 

P/S: Cherry picking from #510


emjotde:
Hi,
Is this resulting in a speed-up? If yes on what hardware?

alvations:
I don't have the exact stats for that but I can try and see if I can share it from the person who has the numbers =)

emjotde:
I am mostly curious where this would help. We either use intrinsics directly or via MKL. Is the compiler doing something smarter then?

kpu:
Marian's default is still `-march=native`, right?  This will pick up SSE4.2 if it's present...

PR #549, opened 2019-11-14
Allow file name templated valid-translation-output files

ugermann:
Allow file name templates for --valid-translation-output. These are essentially format strings that can be interpretted by fmtlib. See config_parser.cpp or --help for valid template parameters. Despite what the git log appears to claim, USE_SENTENCEPIECE is off by default, as in the master branch.


ugermann:
Pinging @snukky.

src/training/validator.cpp
```diff
@@ -353,6 +355,12 @@ float TranslationValidator::validate(const std::vector<Ptr<ExpressionGraph>>& gr
 
   if(options_->hasAndNotEmpty("valid-translation-output")) {
     fileName = options_->get<std::string>("valid-translation-output");
+    fileName = fmt::format(fileName.c_str(),
+			   fmt::arg("E", state.epochs),
+			   fmt::arg("U", state.batches),
+			   fmt::arg("B", state.batchesEpoch),
+			   fmt::arg("W", state.labelsTotal), // alias for T
+			   fmt::arg("T", state.labelsTotal));
```
snukky: Please align all arguments horizontally, i.e. +2 spaces in lines 359-363.

src/training/validator.h
```diff
@@ -37,7 +37,8 @@ class ValidatorBase : public TrainingObserver {
 public:
   ValidatorBase(bool lowerIsBetter) : lowerIsBetter_(lowerIsBetter), lastBest_{initScore()} {}
 
-  virtual float validate(const std::vector<Ptr<ExpressionGraph>>& graphs) = 0;
+  virtual float validate(const std::vector<Ptr<ExpressionGraph>>& graphs,
+                         const TrainingState& state) = 0;
```
snukky: I think we can use `Ptr<TrainingState>` here, following the general convention?

src/training/validator.cpp
snukky: Do we need both "W" and "T"? One of them is not in the help message.

src/common/config_parser.cpp
```diff
@@ -562,7 +562,13 @@ void ConfigParser::addOptionsValidation(cli::CLIWrapper& cli) {
       "Additional args passed to --valid-script-path. These are inserted"
       " between the script path and the output translation-file path");
   cli.add<std::string>("--valid-translation-output",
-     "Path to store the translation");
+     "(Template for) path to store the translation\n"
+     "    template parameters:\n"
```
snukky: Maybe we can start with the simplest usage followed by a more complex one, for example, "Path to store the translation. Can also be a template, e.g. valid-after-{U}-updates-{T}-tokens.out, with parameters: (...)" or similar.

src/common/config_parser.cpp
```diff
@@ -562,7 +562,13 @@ void ConfigParser::addOptionsValidation(cli::CLIWrapper& cli) {
...
+     "    {E} for epoch\n"
+     "    {B} for # of batches within epoch\n"
+     "    {U} for total # of updates\n"
+     "    {T} for total # of tokens seen\n"
+     "    e.g., valid-after-{U}-updates-{T}-tokens.out");
```
snukky: Unfortunately, the CLI library doesn't seem to support multi-line help messages, so everything will be aligned to the left. I think a one-liner is better in that case

src/training/validator.cpp
```diff
@@ -353,6 +355,12 @@ float TranslationValidator::validate(const std::vector<Ptr<ExpressionGraph>>& gr
 
   if(options_->hasAndNotEmpty("valid-translation-output")) {
     fileName = options_->get<std::string>("valid-translation-output");
+    fileName = fmt::format(fileName.c_str(),
```
snukky: I know that fmt is included in spdlog, but maybe we can add an explicit `include` at the beginning of the file for clearness?

src/training/validator.cpp
ugermann: Done.

ugermann: "W" removed.

src/common/config_parser.cpp
ugermann: Done.

src/common/config_parser.cpp
ugermann: Done.

src/training/validator.cpp
ugermann: I put in a comment in the affected function.

src/training/validator.h
ugermann: You mean like in ValidatorBase::actAfterLoaded()? Done.

ugermann:
Changes made as requested.

src/training/validator.cpp
```diff
@@ -353,6 +355,13 @@ float TranslationValidator::validate(const std::vector<Ptr<ExpressionGraph>>& gr
 
   if(options_->hasAndNotEmpty("valid-translation-output")) {
     fileName = options_->get<std::string>("valid-translation-output");
+    // The formatting below uses fmtlib, which is included with spdlog
+    // and is included via the logger.
+    fileName = fmt::format(fileName.c_str(),
+                           fmt::arg("E", state->epochs),
+                           fmt::arg("U", state->batches),
+                           fmt::arg("B", state->batchesEpoch),
+                           fmt::arg("T", state->labelsTotal));
```
snukky: Templating is added to TranslationValidator only, but "valid-translation-output" is also used in BleuValidator. We'd like to have this templating in both.

ugermann: Done.

snukky:
Thanks. I added a regression test for this new feature.

PR #551, opened 2019-11-20
Return exit code 15 (SIGTERM) after SIGTERM.

ugermann:
When marian receives signal SIGTERM and exits gracefully (save model & exit),
it should then exit with a non-zero exit code, to signal to any parent process
that it did not exit "naturally".

### DIFF
--- a/src/command/marian_train.cpp
+++ b/src/command/marian_train.cpp
@@ -68,5 +68,5 @@ int main(int argc, char** argv) {
     }
   }
 
-  return 0;
+  return getSigtermFlag() ? 15 : 0;
 }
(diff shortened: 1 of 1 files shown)


src/command/marian_train.cpp
```diff
@@ -68,5 +68,5 @@ int main(int argc, char** argv) {
     }
   }
 
-  return 0;
+  return getSigtermFlag() ? 15 : 0;
```
emjotde: Maybe add the comment from the PR description. And why 15 specifically?

ugermann: Just added comment and changed the exit code to 128 + SIGTERM. See link in comment for explanation.

PR #553, opened 2019-11-22
Const diligence and thread safety ...

ugermann:
… with respect to Vocab and ShortlistGenerator

- Use pointers to const Vocab instead of Vocab in corpus_base.h and shortlist.h
- Declare some member functions of ShortlistGenerator const
- Use pointers to const ShortlistGenerator instead of ShortlistGenerator in various places.
- Use a thread-local random generator in shortlist.h

The proposed changes allow multiple threads to share pointers / references to ShortlistGenerator instances.


src/data/shortlist.h
```diff
@@ -60,8 +60,8 @@ class SampledShortlistGenerator : public ShortlistGenerator {
   size_t trgIdx_;
   bool shared_{false};
 
-  std::random_device rd_;
-  std::mt19937 gen_;
+  // static thread_local std::random_device rd_;
+  static thread_local std::unique_ptr<std::mt19937> gen_;
```
emjotde: Have you checked this compiles?

emjotde: Ah, this is guarded by #ifdef/#endif, that explains how that would compile despite being wrong. The interface for review on github is really bad.

ugermann:
Undoing that change is guaranteed to break things in the future. The whole point of making some of ShortlistGenerator's member functions (in particular generate) const is that instances can safely be shared across multiple worker threads in the server. If I undo the change in SampledShortlistGenerator, the whole thing falls apart.

In any case SampledShortlistGenerator is currently broken anyway. Look at what its generate() function returns. Won't compile even in the master branch.

frankseide:
My current branch says about SampledShortlistGenerator `// Intended for use during training in the future, currently disabled`, and I don't see it referenced anywhere.

I now also have doubts about a thread-local. That will imply that each worker is permanently tied to a thread. But Marian uses threads a lot in the form of thread pools, right? Thread pools by design do not guarantee on what thread a task runs on. Imagine for one batch, worker 1 runs on thread 1, and for the next batch, worker 1 runs on thread 2. If you ever repeat such a run, it may end up on thread 3 due to non-determinism. Now your trainings are no longer reproducible.

Now, it would still work deterministically if each thread-pool task runs a worker throughout the entire training, or if threads are always allocated in the same order and torn down (that's what we do in our current sync training, I believe). However, that is sort of "luck", since neither the thread-pool class nor the SampledShortlistGenerator (with thread-local static) can guarantee that. In my view, the correct thing is to have the object instance to maintain random state, and then to reference that object instance tied to a worker, not tied to a thread in a thread pool. Then the same sequence of operations on that object instance is guaranteed to have the same outcomes.

frankseide:
BTW, can a method that depends on a random sequence be `const`? The random sequence implies some mutable state somewhere (thread-local or instance-internal).

ugermann:
> BTW, can a method that depends on a random sequence be `const`? The random sequence implies some mutable state somewhere (thread-local or instance-internal).

Yes. All you have to do is make the random generator external to the the method. 

Currently, ShortlistGenerator is not thread-safe. Making certain functions const gives developers assurances that certain uses are thread-safe. Throwing that benefit away for one special case that may or may not be used in the future is overkill in my opinion. Keep in mind that SampledShortlistGenerator is currently not even used.

frankseide:
>Yes. All you have to do is make the random generator external to the the method.

But that is not done in this PR, right?

emjotde:
Hi, my request to undo was more meant to be not be a surprise to me when I ever happen to uncomment it. More likely I am going to remove it.

emjotde:
Regression tests pass? If yes, good to go.

ugermann:
> > Yes. All you have to do is make the random generator external to the the method.
> 
> But that is not done in this PR, right?
> 
> > Currently, ShortlistGenerator is not thread-safe. Making certain functions const gives developers assurances that certain uses are thread-safe. Throwing that benefit away for one special case that may or may not be used in the future is overkill in my opinion. Keep in mind that SampledShortlistGenerator is currently not even used.
> 
> Ah, I see. But how does making it a static fix that?

Only static members can be thread_local. In principle, I'm aware of three options to deal with random generators in a multi-threaded scenario.

1. Make the generator thread_local.
2. Use  a lock. This is probably a bad idea in most scenarios unless you need a random number only very rarely, because locks  slow things down a lot.
3. Pass the random generator as a non-const parameter to the function. Then it's up to the caller to manage the random generator. This would however require changes in the interface and trigger debates about whether it's worth it.

Considering that the SampledShortlistGenerator is currently not used, I'd postpone that discussion to the time when this actually becomes a real issue.

emjotde:
Agreed, I didn't intend to spark any discussion here, since that is dead code.

ugermann:
> Regression tests pass? If yes, good to go.

Currently running.

ugermann:
[regtests-master.log](https://github.com/marian-nmt/marian-dev/files/3882061/regtests-master.log)
[regtests-ug-const-diligence.log](https://github.com/marian-nmt/marian-dev/files/3882062/regtests-ug-const-diligence.log)
same machine, fresh compile of both

emjotde:
I think you compiled without MKL. Should be OK. The failed test should actually be skipped.

PR #561, opened 2019-12-04
Fix compilation on CPUs that don't support AVX and some white space i…

### LINKED ISSUE #559: Marian won't compile on older CPUs (up to and including westmere)
... apparently that's because those architectures don't support -mavx.

@snukky try compiling on hel. Won't work with
```
cmake -DCOMPILE_CUDA=off ..
```

XapaJIaMnu:
This should fix #559 
Note that I didn't actually manage to test whether the code works, as I don't have access to Non AVX system with MKL installed, which is a requirement for float32 decoding. I tested on hel and marian now compiles cleanly and during decoding crashes with `You need to compile with MKL in order to use the CPU version`, which I assume means it managed to go fine through all other operations. Or I may be wrong.


snukky:
On hel, you can now use:

    cmake .. -DMKL_ROOT=/home/romang/intel/mkl

XapaJIaMnu:
Ok I can confirm now that this fixes the issue

emjotde:
Thanks.
I would prefer to have this expressed negatively. Here we have to explicitly set USE_AVX to compile AVX code while we are only handling some old hardware, Rather we should have something like

```
#ifndef NO_AVX
...
#endif
```

and always compile if that is not defined. So in the CMakeLists.txt only set NO_AVX if avx is not available. Otherwise this messes with our Windows code etc.

XapaJIaMnu:
Is this good enough?

emjotde:
Looks good. Let me test a bit.

PR #563, opened 2019-12-05
Update logging messages for training

snukky:
A small update of some logging messages:
* Formatting large numbers with commas
* Setting logging levels of some less informative (for an average user) messages to _debug_.
* Adding a few `[]` labels.

The changes do not require updating regression tests.

### DIFF
--- a/src/layers/generic.cpp
+++ b/src/layers/generic.cpp
@@ -19,7 +19,7 @@ namespace marian {
   // This function assumes that the object holds one or more factor logits.
   // It applies the supplied loss function to each, and then returns the aggregate loss over all factors.
   Expr Logits::applyLossFunction(const Words& labels, const std::function<Expr(Expr/*logits*/, Expr/*indices*/)>& lossFn) const {
-    LOG_ONCE(debug, "[logits] applyLossFunction() for {} factors", logits_.size());
+    LOG_ONCE(info, "[logits] Applying loss function for {} factor(s)", logits_.size());
     ABORT_IF(empty(), "Attempted to read out logits on empty Logits object");
 
     auto firstLogits = logits_.front()->loss();
--- a/src/layers/generic.cpp
+++ b/src/layers/generic.cpp
@@ -19,7 +19,7 @@ namespace marian {
   // This function assumes that the object holds one or more factor logits.
   // It applies the supplied loss function to each, and then returns the aggregate loss over all factors.
   Expr Logits::applyLossFunction(const Words& labels, const std::function<Expr(Expr/*logits*/, Expr/*indices*/)>& lossFn) const {
-    LOG_ONCE(info, "[logits] applyLossFunction() for {} factors", logits_.size());
+    LOG_ONCE(debug, "[logits] applyLossFunction() for {} factors", logits_.size());
     ABORT_IF(empty(), "Attempted to read out logits on empty Logits object");
 
     auto firstLogits = logits_.front()->loss();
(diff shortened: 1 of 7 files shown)


src/layers/generic.cpp
```diff
@@ -19,7 +19,7 @@ namespace marian {
   // This function assumes that the object holds one or more factor logits.
   // It applies the supplied loss function to each, and then returns the aggregate loss over all factors.
   Expr Logits::applyLossFunction(const Words& labels, const std::function<Expr(Expr/*logits*/, Expr/*indices*/)>& lossFn) const {
-    LOG_ONCE(info, "[logits] applyLossFunction() for {} factors", logits_.size());
+    LOG_ONCE(debug, "[logits] applyLossFunction() for {} factors", logits_.size());
```
emjotde: Let's keep "info" for now. This is a relatively new feature and good to monitor it.

src/training/communicator.cpp
```diff
@@ -139,7 +139,7 @@ class FakeMPIWrapper : public IMPIWrapper
 {
 public:
   FakeMPIWrapper(bool) {
-    LOG(warn, "Compiled without MPI support. Falling back to FakeMPIWrapper");
+    LOG(debug, "[comm] Compiled without MPI support. Falling back to FakeMPIWrapper");
```
emjotde: Also, let's keep info.

src/training/communicator_nccl.h
```diff
@@ -191,7 +191,7 @@ class NCCLCommunicator : public ICommunicator {
     groupEnd();
 
     mpiBarrier(); // (synchronize the log messages)
-    LOG(info, "[comm] NCCLCommunicator constructed successfully.");
+    LOG(debug, "[comm] NCCLCommunicator constructed successfully");
```
emjotde: As above, let's keep info. These log messages are not really useful when everything works, but useful if it didn't work. Good to have them in there by default in later error assessment, even when a user files issues.

src/training/communicator.cpp
emjotde: Ah, original had "warn", but I think "info" is OK here.

frankseide:
That MPI message (which I wrote) bugs me as well. I wonder if we can suppress that MPI message unless some configuration indicates that multi-node is expected. Do we have such a condition? Or could we maybe assume that a user will find out quickly from the log that it is running without MPI?

What bugs me most is the "fake" name in the log message. That's an internal class name that is not relevant to the user. Maybe we can simply say
```
[comm] Compiled without MPI support. Running as single process on $HOSTNAME.
```

emjotde:
Roman has the salomonic solution to split this in half between info and debug. That should be fine.

src/layers/generic.cpp
snukky: Done, and changed into a little less confusing "Applying loss function ..."

src/training/communicator.cpp
snukky: I followed Frank's suggestion with "Running as single process on ..."

src/training/communicator_nccl.h
snukky: Done

snukky:
I also added commas in vocabulary sizes reported in logs, and updated some regression tests to work with both formatting options (https://github.com/marian-nmt/marian-regression-tests/commit/549d0b1a3bea12a27d7bd772420dd826062a52ad).

PR #564, opened 2019-12-05
Use __AVX__ compiler define instead of custom and broken NO_AVX flag.  Fixes #559

### LINKED ISSUE #559: Marian won't compile on older CPUs (up to and including westmere)

kpu:
Also, I did chmod -x on the 3rdparty source code.


frankseide:
Please be careful with this. `__AVX__` is a pre-defined symbol. I am quite sure that that is the case for the Windows compiler (https://stackoverflow.com/questions/28939652/how-to-detect-sse-sse2-avx-avx2-avx-512-avx-128-fma-kcvi-availability-at-compile).

kpu:
Yes, it is a predefined symbol.  That is the point of this change.

emjotde:
@frankseide would have been my first question too. :) Looks good then?

emjotde:
Also, what was broken with the original one?

kpu:
@emjotde It was setting the flag based on the host system, not the compiler flags: https://github.com/marian-nmt/marian-dev/issues/559#issuecomment-562282687

emjotde:
Thanks!

PR #575, opened 2020-01-02
update_cycle function

lkfo415579:
Nowaday NMT model are getting larger and deeper, which requires huge gpu memory. Update_cycle is a trick that accumulates N times gradients by average in one step. It can effiectly reduce huge memory consumption. Here is my implementation.


emjotde:
Hi, can you elaborate a bit what this is doing and how it is different from --optimizer-delay N which performs a model update after N forward/backward steps (accumulating gradient sums).

lkfo415579:
> Hi, can you elaborate a bit what this is doing and how it is different from --optimizer-delay N which performs a model update after N forward/backward steps (accumulating gradient sums).

I am sorry, i didn't notice this option at the begining. Actually i did this implement in the earlier version like V1.7.6, there is no --optimizer-delay option. I am happy that you guys added this function.

lkfo415579:
@emjotde Update_cycle will update N forward/backward gradients accumulatly on **each step**. It may be different from the optimizer-delay appoarch, about differnt learning rate. I think update_cycle should be a better appoarch, is my concern correct?

PR #585, opened 2020-01-21
Add a warning message for an incorrect use of output sampling

snukky:
The warning is printed after the config options are printed if `--output-sampling` is used with `--beam-size > 1` or model ensembles.

### DIFF
--- a/src/translator/translator.h
+++ b/src/translator/translator.h
@@ -114,11 +114,11 @@ public:
       if(options_->get<size_t>("beam-size") > 1)
         LOG(warn,
             "[warning] Output sampling and beam search (beam-size > 1) are contradictory methods "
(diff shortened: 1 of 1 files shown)


src/translator/translator.h
```diff
@@ -109,6 +109,17 @@ class Translate : public ModelTask {
 
       threadPool.enqueue(task, device, id++);
     }
+
+    if(options_->get<bool>("output-sampling", false)) {
+      if(options_->get<size_t>("beam-size") > 1)
+        LOG(warn,
+            "[warning] Output sampling and beam search (beam-size > 1) are contradictory methods "
```
emjotde: Maybe, remove "highly" in both messages. Other than that good to go and merge.

snukky: Done

PR #595, opened 2020-02-05
Integrate intgemm into marian

XapaJIaMnu:
Hey Guys,

This is a big one, here's what it does:

- Adds [intgemm](https://github.com/kpu/intgemm/) as a module for Marian. Intgemm is @kpu 's 8/16 bit gemm library with support for architectures from SSE2 to AVX512VNNI
- Removes outdated integer code, related to the `--optimize` option
- Adds a new option `--optimize8`, where the model is converted to 8bit on the fly during decoding. `--optimize` is kept with the old behaviour (16bit on-the-fly conversion)
- Adds `intgemm8` and `intgemm16` architecture agnostic binary format for packed models.

During decoding a packed model, regardless whether you use the `intgemm` or `fbgemm` format, you can get increased performance by providing the `--optimize8` switch, which makes sure that the vocabulary shortlisted matrix is processed by `intgemm` and not by `mkl` or `openblas`. 

Here are some preliminary numbers computed by decoding 2000 lines on Estonian-English student model, prepared by @snukky 

| Configuration  |  AVX2     | AVX512VNNI  |
| ------------------- |:-----------:| -------------------:|
| MKL                                  | 23             | 17.5 |
| FBGEMM + MKL              | 18.9          |  13.4 |
| FBGEMM  + Intgemm16  | 17.9          |  13.2 |
| **FBGEMM  + Intgemm8**   | **17.3**          |  **12.5** |
| Intgemm8                         | 17.7          |  15.4 |
| Intgemm8 + MKL              | 19.7         |   16.2 |
| Intgemm16                       | 21.4         |   18.1 |

Cheers,

Nick


ykim362:
This is great, guys!
Could you also add the size of student model somewhere such as decoder recurrence units, layer tying, state sizes and etc?

XapaJIaMnu:
Hey @ykim362 ,

Turns out gcc 5.4 doesn't like `data<TYPE>()` sometimes. I worked around the issue with `reinterpret_cast`. As for the model:

* Smaller transformer architecture: 256 emb., 1536 FFN, 6-layer encoder, 2-layer decoder with SSRU units (tiny11).
* Vocabulary shortlist was used

It should be similar to the WNGT submission.

Cheers,

Nick

snukky:
The student architecture we use is like "Tiny" from the WNGT submission but with 2-layer decoder. The layers are not tied as we've been training with guided alignment.

ykim362:
Thanks, @XapaJIaMnu and @snukky !
I got a segment fault when I ran fbgemm int8 model with --optimize8 option and no shortlist.
It works fine when I ran fbgemm int8 model with --optimize8 and shortlist.
It's seg fault without much information. Could you try this setting? (no shortlist).

src/command/marian_conv.cpp
```diff
@@ -39,6 +39,10 @@ int main(int argc, char** argv) {
     saveGemmType = Type::packed8avx2;
   } else if(saveGemmTypeStr == "packed8avx512") { // packed8 for AVX512
     saveGemmType = Type::packed8avx512;
+  } else if(saveGemmTypeStr == "intgemm8") { // packed8 for AVX512
```
ykim362: Comments need to be updated.

XapaJIaMnu:
@ykim362 , could you share a input parameters. I just tried this and it worked for me:
```
/home/nbogoych/.local/bin/sacrebleu -t wmt18 -l $SRC-$TRG --echo src > speed/newstest2018.$SRC

test -e model8.bin || $MARIAN/marian-conv -f model.npz -t model8.bin --gemm-type packed8avx512

echo "### Translating wmt18 $SRC-$TRG on CPU"
$MARIAN/marian-decoder $@ \
    --relative-paths -m model8.bin -v vocab.eten.spm vocab.eten.spm \
    -i speed/newstest2018.$SRC -o speed/cpu.newstest2018.packed8.$TRG \
    --beam-size 1 --mini-batch 32 --maxi-batch 100 --maxi-batch-sort src -w 128 \
    --cpu-threads 1 --optimize8 \
    --quiet --quiet-translation --log speed/cpu.newstest2018.packed8.log
```
Cheers,
Nick

ykim362:
@XapaJIaMnu , below is my command.
This crashes.
```
/data/repos/marian-dev-external/build/marian-decoder -m model/model.npz.best-bleu.0203.int8.avx512.bin --cpu-threads 1 --mini-batch 1 --maxi-batch 100 --maxi-batch-sort src -w 768 --beam-size 1 -n --max-length 100 --max-length-crop -i valid_300.src --quiet-translation -v model/DEU.ENU.generalnn.source.vocab model/DEU.ENU.generalnn.target_input.vocab --optimize8
```

This works fine.
```
/data/repos/marian-dev-external/build/marian-decoder -m model/model.npz.best-bleu.0203.int8.avx512.bin --cpu-threads 1 --mini-batch 1 --maxi-batch 100 --maxi-batch-sort src -w 768 --beam-size 1 -n --max-length 100 --max-length-crop -i valid_300.src --quiet-translation -v model/DEU.ENU.generalnn.source.vocab model/DEU.ENU.generalnn.target_input.vocab --optimize8 --shortlist model/lex.s2t 300 200 0
```

ykim362:
@XapaJIaMnu FYI, My machine is a SKX. AVX512 without VNNI.

XapaJIaMnu:
@ykim362 I can't get it to crash, even on AVX512, I suspect it might have something to do with the architecture of the model. Could you share your model or a configuration script so i can try to create one.

As far as I recall the `Wemb` matrix is not part of the fbgemm compressed matrices, so it must be this one that is being peculiar when there is no shortlist, but I can't reproduce it with this model configuration.

kpu:
What's the vocabulary size when it crashes?  @ykim362 Can you compile with `assert` enabled and see if it triggers an assert?

ykim362:
@kpu , @XapaJIaMnu 
One thing maybe related or not. If you comment out, long-term memory from the expression graph, it starts to allocate more and more memories and killed without generating any outputs.

```
    if(node->type() != "param" && node->memoize()) {
      auto it = longterm_->find(hash);
      if(it != longterm_->end()) {
        for(auto found : it->second) {
          return found;
          // @TODO: check why below code does not work for certain nodes and
          // autotuning.
          // if(node->equal(found)) {
          // std::cerr << "found memoized" << std::endl;
          // return found;
          //}
        }
      }
      (*longterm_)[hash].push_back(node);
    }
```


```
[2020-02-06 00:20:00] [memory] Reserving 85 MB, device cpu0
[2020-02-06 00:20:00] [memory] Reserving 21 MB, device cpu0
tcmalloc: large alloc 1073741824 bytes == 0x47712000 @
tcmalloc: large alloc 1207959552 bytes == 0x87712000 @
tcmalloc: large alloc 1476395008 bytes == 0x47712000 @
tcmalloc: large alloc 1610612736 bytes == 0x9f712000 @
tcmalloc: large alloc 1744830464 bytes == 0xff712000 @
tcmalloc: large alloc 2013265920 bytes == 0xb7712000 @
tcmalloc: large alloc 2281701376 bytes == 0x47712000 @
tcmalloc: large alloc 2550136832 bytes == 0x15f712000 @
tcmalloc: large alloc 2818572288 bytes == 0xe7712000 @
tcmalloc: large alloc 3221225472 bytes == 0xff712000 @
tcmalloc: large alloc 3489660928 bytes == 0x47712000 @
tcmalloc: large alloc 4026531840 bytes == 0x12f712000 @
tcmalloc: large alloc 4429185024 bytes == 0x147712000 @
tcmalloc: large alloc 4966055936 bytes == 0x47712000 @
tcmalloc: large alloc 5637144576 bytes == 0x24ff12000 @
tcmalloc: large alloc 6308233216 bytes == 0x47712000 @
tcmalloc: large alloc 7113539584 bytes == 0x47712000 @
tcmalloc: large alloc 8053063680 bytes == 0x3f8712000 @
tcmalloc: large alloc 8589934592 bytes == 0x3f8712000 @
tcmalloc: large alloc 8724152320 bytes == 0x47712000 @
tcmalloc: large alloc 8858370048 bytes == 0x3f8712000 @
tcmalloc: large alloc 8992587776 bytes == 0x768f12000 @
tcmalloc: large alloc 9126805504 bytes == 0x3f8712000 @
tcmalloc: large alloc 9261023232 bytes == 0x981712000 @
tcmalloc: large alloc 9395240960 bytes == 0x3f8712000 @
tcmalloc: large alloc 9529458688 bytes == 0xba9f12000 @
tcmalloc: large alloc 9663676416 bytes == 0x3f8712000 @
tcmalloc: large alloc 9797894144 bytes == 0xde2f12000 @
tcmalloc: large alloc 9932111872 bytes == 0x3f8712000 @
tcmalloc: large alloc 10066329600 bytes == 0x102b712000 @
tcmalloc: large alloc 10200547328 bytes == 0x3f8712000 @
tcmalloc: large alloc 10334765056 bytes == 0x1283f12000 @
tcmalloc: large alloc 10468982784 bytes == 0x3f8712000 @
tcmalloc: large alloc 10603200512 bytes == 0x14ec712000 @
tcmalloc: large alloc 10737418240 bytes == 0x3f8712000 @
tcmalloc: large alloc 10871635968 bytes == 0x1764f12000 @
tcmalloc: large alloc 11005853696 bytes == 0x3f8712000 @
tcmalloc: large alloc 11140071424 bytes == 0x19edf12000 @
tcmalloc: large alloc 11274289152 bytes == 0x3f8712000 @
tcmalloc: large alloc 11408506880 bytes == 0x1c86712000 @
tcmalloc: large alloc 11542724608 bytes == 0x3f8712000 @
tcmalloc: large alloc 11676942336 bytes == 0x1f2ef12000 @
tcmalloc: large alloc 11811160064 bytes == 0x3f8712000 @
tcmalloc: large alloc 11945377792 bytes == 0x21e7f12000 @
tcmalloc: large alloc 12079595520 bytes == 0x3f8712000 @
tcmalloc: large alloc 12213813248 bytes == 0x24b0712000 @
tcmalloc: large alloc 12348030976 bytes == 0x3f8712000 @
tcmalloc: large alloc 12482248704 bytes == 0x2788f12000 @
tcmalloc: large alloc 12616466432 bytes == 0x3f8712000 @
tcmalloc: large alloc 12750684160 bytes == 0x2a71f12000 @
tcmalloc: large alloc 12884901888 bytes == 0x3f8712000 @
tcmalloc: large alloc 13019119616 bytes == 0x2d6a712000 @
tcmalloc: large alloc 13153337344 bytes == 0x3f8712000 @
tcmalloc: large alloc 13287555072 bytes == 0x3073712000 @
tcmalloc: large alloc 13421772800 bytes == 0x3f8712000 @
tcmalloc: large alloc 13555990528 bytes == 0x338bf12000 @
tcmalloc: large alloc 13690208256 bytes == 0x3f8712000 @
tcmalloc: large alloc 13824425984 bytes == 0x36b4f12000 @
tcmalloc: large alloc 13958643712 bytes == 0x3f8712000 @
tcmalloc: large alloc 14092861440 bytes == 0x39edf12000 @
tcmalloc: large alloc 14227079168 bytes == 0x3f8712000 @
tcmalloc: large alloc 14361296896 bytes == 0x3d36712000 @
tcmalloc: large alloc 14495514624 bytes == 0x3f8712000 @
tcmalloc: large alloc 14629732352 bytes == 0x408f712000 @
tcmalloc: large alloc 14763950080 bytes == 0x3f8712000 @
tcmalloc: large alloc 14898167808 bytes == 0x43f8712000 @
tcmalloc: large alloc 15032385536 bytes == 0x4770f12000 @
tcmalloc: large alloc 15166603264 bytes == 0x4af1f12000 @
tcmalloc: large alloc 15300820992 bytes == 0x4e7af12000 @
tcmalloc: large alloc 15435038720 bytes == 0x520bf12000 @
tcmalloc: large alloc 15569256448 bytes == 0x55a4712000 @
tcmalloc: large alloc 15703474176 bytes == 0x5945712000 @
tcmalloc: large alloc 15837691904 bytes == 0x5cee712000 @
tcmalloc: large alloc 15971909632 bytes == 0x609f712000 @
tcmalloc: large alloc 16106127360 bytes == 0x6458712000 @
tcmalloc: large alloc 16240345088 bytes == 0x6819712000 @
tcmalloc: large alloc 16374562816 bytes == 0x6be2712000 @
tcmalloc: large alloc 16508780544 bytes == 0x6fb3712000 @
tcmalloc: large alloc 16642998272 bytes == 0x738bf12000 @
tcmalloc: large alloc 16777216000 bytes == 0x776cf12000 @
Killed
```

XapaJIaMnu:
@ykim362 technically, intgemm only supports matrices that are multiple of 8 (for now) so this vocabulary size will crash it. If you compile marian with a debug build, you'd see some asserts being hit. Of course, a crash is not guaranteed every time, since segfault is never guaranteed.

As for the endless memory allocation, when doing `--optimize` or `--optimize8` on the fly we rely on the cache to make sure that matrices are not double prepared. I suspect that disabling this behaviour will lead to the allocation of more and more matrices. Should we expect to work without long term memory enabled? I will take a look and try to see what's happening.

kpu:
What @XapaJiaMnu meant to say regarding the segfault is he will add an error message or padding for now, while the templated version of intgemm is underway.

ykim362:
Thanks for the clarification, @XapaJIaMnu , @kpu .
For one of my use cases, I should not rely on any caching. (just the model parameter and scratch workspace persist only for one batch). If there's any way to improve the performance for this scenario, that would be great. 
Thanks!

XapaJIaMnu:
@ykim362 for this use could you try intgemm8 binary format without shortlisting (or optimise). I think this should work for now. I will try to find a way to see how to deal with the shortlist caching.

Cheers, 

Nick

XapaJIaMnu:
@ykim362 , I tried the persistent memory code. I noticed that it is commented out in master because it doesn't work for some cases. Could you clarify what it does. For some reason it breaks the hashing of intgemm matrices and they keep getting recreated and reallocated.

ykim362:
Thanks, @XapaJIaMnu .

> could you try intgemm8 binary format without shortlisting (or optimise)
Can we do this for the output projection matrix while it's shared for the input embedding? If so, that would be great. 

> , I tried the persistent memory code. I noticed that it is commented out in master because it doesn't work for some cases

I haven't made any changes in that code in the master. From my understanding, quantized matrices are caches in long term memory and I commented it out for my use case test.

ykim362:
@XapaJIaMnu BTW, is this change tested with factored segmentor vocab models?

XapaJIaMnu:
> Thanks, @XapaJIaMnu .
> 
> > could you try intgemm8 binary format without shortlisting (or optimise)
> 
> Can we do this for the output projection matrix while it's shared for the input embedding? If so, that would be great.
> 
Yeah, it wouldn't be an issue, as long as you have dimensions multiple of 8. (This limitation  will be fixed in future versions). The only thing that is not working for now is intgemm + long term memory, which is disabled in master by default.
> > , I tried the persistent memory code. I noticed that it is commented out in master because it doesn't work for some cases
> 
> I haven't made any changes in that code in the master. From my understanding, quantized matrices are caches in long term memory and I commented it out for my use case test.

Why are those lines of the code:
```
          // @TODO: check why below code does not work for certain nodes and
          // autotuning.
          // if(node->equal(found)) {
          // std::cerr << "found memoized" << std::endl;
          // return found;
          //}
```
Commented out in master? For what cases does it not work?

ykim362:
@XapaJIaMnu 

> Why are those lines of the code:
          // @TODO: check why below code does not work for certain nodes and
          // autotuning.
          // if(node->equal(found)) {
          // std::cerr << "found memoized" << std::endl;
          // return found;
          //}
Commented out in master? For what cases does it not work?

This may be a good question for @emjotde .

XapaJIaMnu:
@ykim362 , the reason long term memory is not working, is because my hashing is not working correctly, and because i create extra node when quantizing the activations matrix, which long term memory attempted to cache. I need to carefully mark the non-memoizeable nodes as memoizeable.

Even then, I have the issue where I have nodes that have the same hash but the `equals()` returns false, which is why long term memory overflows. I am going through the codebase now looking at other node's implementation of `hash()` and `equals()`.  The default ones seem to not be working for the intgemm nodes.

XapaJIaMnu:
@ykim362 , long term memory should be working now. Could you take another look?

Cheers,

Nick

ykim362:
Thanks, @XapaJIaMnu ! I will do another test with my use case this week.

alvations:
Quick question, is the int4 code somewhere here in this branch too?

kpu:
No.  The log4 dot product is at https://github.com/kpu/intgemm/blob/log4-unstable/log4/log4.h and you can read the comments for all the instructions needed to get it to run on a CPU.  Currently a log4 dot product takes at least 2x the time of an 8-bit dot product so there isn't a speedup.

src/layers/generic.cpp
```diff
@@ -258,7 +260,31 @@ namespace marian {
       lazyConstruct(input->shape()[-1]);
 
       if (shortlist_ && !cachedShortWt_) { // shortlisted versions of parameters are cached within one batch, then clear()ed
-        cachedShortWt_ = index_select(Wt_, isLegacyUntransposedW ? -1 : 0, shortlist_->indices());
+        if ((graph_->getBackend()->isOptimized8() || matchType<intgemm8>(Wt_->value_type()) )&& graph_->getDeviceId().type == DeviceType::cpu) {
```
ykim362: @frankseide any opinions?

src/tensors/cpu/expression_graph_packable.h
```diff
@@ -1,10 +1,57 @@
 #pragma once
 
 #include "graph/expression_graph.h"
-#include "packed_gemm.h"
+#include "fbgemm/packed_gemm.h"
+#include "tensors/cpu/integer_common.h"
+
+namespace { //Temporary annonymous transposition, until we figure out how to access the proper one
+inline void transpose4x4_SSE(const float* A,
```
ykim362: Can we move this to integer_common somewhere?

src/tensors/cpu/intgemm_interface.h
```diff
@@ -0,0 +1,340 @@
...
+float quantMult_;
+
+  PrepareBNodeOp(Expr input, Expr quant_mult, float clipValue)
+      : NaryNodeOp({input, quant_mult}, input->shape(), intgemm_<vtype>::intgemmType), clipValue_{clipValue} {
+
+    set_name(input->name());
+    // Check if arguments are not null
+    ABORT_IF(child(0) == nullptr, "A cannot be null");
+    ABORT_IF(child(1) == nullptr, "Quant mult of B cannot be null");
+    ABORT_IF(input->shape()[-1] %8 != 0, "Columns of matrix: " + input->type() + " must be multiple of 8.");
```
ykim362: @XapaJIaMnu It would be nice to have a support for non 8 multiple matrices. We have many models with non 8 multiple vocab sizes.

kpu: I know.  It's coming.  https://github.com/kpu/intgemm/blob/static/tile/dot.inl#L147

src/common/types.cpp
```diff
@@ -26,10 +26,15 @@ size_t requiredBytes(const Shape& shape, Type type) {
       ABORT("Not a supported data type: {}", type);
       return 0;
     }
+  } else if (isIntgemm(type)) {
+  return shape.elements() * sizeOf(type) + sizeOf(Type::float32);
```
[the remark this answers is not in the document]
kpu: Scaling factor.

src/CMakeLists.txt
```diff
@@ -5,6 +5,8 @@ include_directories(3rd_party)
 include_directories(3rd_party/SQLiteCpp/include)
 include_directories(3rd_party/sentencepiece)
 include_directories(3rd_party/fbgemm/include)
+include_directories(3rd_party/intgemm)
+include_directories(${CMAKE_BINARY_DIR}/src/3rd_party/intgemm)
```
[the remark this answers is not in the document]
XapaJIaMnu: No, one of them refers to the source directory and the other one refers to the build directory. This is necessary because we have configuration generated in the build directory.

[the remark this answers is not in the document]
kpu: We do compilation tests to check if the compiler supports AVX512BW and AVX512VNNI.  That generates a header ${CMAKE_BINARY_DIR}/src/3rd_party/intgemm/intgemm_config.h .  

intgemm itself has relative paths.  So if Marian included intgemm by relative path, we could remove `include_directories(3rd_party/intgemm)`.

src/tensors/cpu/expression_graph_packable.h
XapaJIaMnu: This is a shameless code duplication, because I couldn't figure out how to access a transpose outside of the graph. Can we do this in a better way? I took this one from `tensors/cpu/tensor_operators.cpp`

XapaJIaMnu:
Here's some updated benchmarks, since the last ones were many code changes ago. The configuration is the same as before: Estonian-English WMT18 2000 sentences. The configuration is the same as "tiny" from the WNGT paper, except there are 2 layers in the decoder, untied (one of them is for guided alignment).

| Configuration  |  AVX2     | AVX512VNNI  |
| ------------------- |:-----------:| -------------------:|
| MKL                                  | 22.3             | 17 |
| FBGEMM + MKL              | 18.5          |  13 |
| FBGEMM  + Intgemm16  | 17.3          |  12.5 |
| **FBGEMM  + Intgemm8**   | **16.7**          |  **12** |
| Intgemm8                         | 17.4          |  15 |
| Intgemm8 + MKL              | 19         |   16 |
| Intgemm16                       | 20.7        |   16.7 |

src/common/types.cpp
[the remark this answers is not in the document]
XapaJIaMnu: The extra float holds the quantization multiplier for the tensor. Fixed indentation and added the comment.

src/command/marian_conv.cpp
```diff
@@ -52,7 +56,10 @@ int main(int argc, char** argv) {
 
   auto graph = New<ExpressionGraphPackable>();
   graph->setDevice(CPU0);
-  graph->getBackend()->setOptimized(false);
+  if (saveGemmType != Type::intgemm16)
+    graph->getBackend()->setOptimized(false);
```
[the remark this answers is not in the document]
XapaJIaMnu: I just kept the old options unchanged. @emjotde, any opinions here?

XapaJIaMnu:
Implemented optional "shifted" multiplication routine, that seems to be faster on AVX2 and AVX512VNNI (and probably AVX512F). To use it please supply the command line argument `--intgemm-shifted` (in conjunction with `--optimize8`). Updated results:
| Configuration  |  AVX2     | AVX512VNNI  |
| ------------------- |:-----------:| -------------------:|
| MKL                                         | 22.3             | 17 |
| FBGEMM + MKL                     | 18.5          |  13 |
| FBGEMM  + Intgemm16         | 17.3          |  12.5 |
| FBGEMM  + Intgemm8           | 16.7          |  12 |
| **FBGEMM  + Intgemm8Shift**    | **16.5**         |  **11.8** |
| Intgemm8                                | 17.4          |  15 |
| Intgemm8Shift                         | 16.5          |  12 |
| Intgemm8 + MKL                     | 19         |   16 |
| Intgemm16                              | 20.7        |   16.7 |

kpu:
Pinging @emjotde "For me, it looks good. Most of my comments are questions or suggestions. So, feel free to resolve. I will pass it to @emjotde ." - @ykim362

emjotde:
@ykim362 is testing. I will have a good look now in parallel.  All, thanks for all your huge amount of work so far.

src/tensors/cpu/intgemm_interface.h
```diff
@@ -0,0 +1,391 @@
...
+#include "graph/node_operators_unary.h"
+#include "integer_common.h"
+
+namespace marian {
+
+namespace cpu {
+namespace integer {
+
+template<Type vtype>
+struct PrepareANodeOp : public NaryNodeOp {
```
emjotde, line 13: Please add class-level comments to every new node type describing what they do and what their purpose is.

src/tensors/cpu/intgemm_interface.h
```diff
@@ -0,0 +1,391 @@
...
+  NodeOps backwardOps() override {
+    ABORT("Only used for inference");
+    return {NodeOp(0)};
+  }
+
+  const std::string type() override { return "intgemmPrepareA"; }
+};
+
+template<Type vtype>
+struct PrepareBNodeOp : public NaryNodeOp {
```
emjotde, line 60: And here

src/tensors/cpu/intgemm_interface.h
```diff
@@ -0,0 +1,391 @@
...
+  NodeOps backwardOps() override {
+    ABORT("Only used for inference");
+    return {NodeOp(0)};
+  }
+
+  const std::string type() override { return "intgemmPrepareB"; }
+};
+
+template<Type vtype>
+struct SelectColumnsBNodeOp : public UnaryNodeOp {
```
emjotde, line 99: And here.

src/tensors/cpu/intgemm_interface.h
```diff
@@ -0,0 +1,391 @@
...
+    Shape ret = a->shape();
+    ret.set(1, indices.size());
+    return ret;
+  }
+
+  std::vector<uint_least32_t> indices_;
+};
+
+template<Type vtype>
+struct QuantMultNodeOp : public UnaryNodeOp {
```
emjotde, line 165: And here

src/tensors/cpu/intgemm_interface.h
```diff
@@ -0,0 +1,391 @@
...
+
+  const std::string type() override {
+    if (isA_)
+      return "intgemmQuantMultA";
+    else
+      return "intgemmQuantMultB";
+  }
+};
+
+class PrepareBiasForBNodeOp : public NaryNodeOp {
```
emjotde, line 205: And here.

emjotde:
I have mostly requests for comments and the larger question of this being outside Marian now. 

I would prefer to either move that into the marian-nmt organization or into a periodically updated code copy.

src/common/binary.cpp
```diff
@@ -58,12 +59,19 @@ void loadItems(const void* current, std::vector<io::Item>& items, bool mapped) {
...
   for(int i = 0; i < numHeaders; ++i) {
     if(items[i].mapped) { // memory-mapped, hence only set pointer
+      ABORT_IF(isIntgemm(items[i].type), "mmap format not supported for intgemm matrices");
       items[i].ptr = get<char>(current, headers[i].dataLength);
     } else { // reading into item data
       size_t len = headers[i].dataLength;
       items[i].bytes.resize(len);
       const char* ptr = get<char>(current, len);
-      std::copy(ptr, ptr + len, items[i].bytes.begin());
+      if (matchType<intgemm8>(items[i].type)) {
```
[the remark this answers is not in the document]
XapaJIaMnu: This allows us to have architecture agnostic binary format, by doing the matrix reordering at load time.

src/tensors/cpu/expression_graph_packable.h
```diff
@@ -1,10 +1,57 @@
...
+  _mm_store_ps(&B[1 * ldb], row2);
+  _mm_store_ps(&B[2 * ldb], row3);
+  _mm_store_ps(&B[3 * ldb], row4);
+}
+
+// from
+// https://stackoverflow.com/questions/16737298/what-is-the-fastest-way-to-transpose-a-matrix-in-c
+#define ROUND_UP(x, s) (((x) + ((s)-1)) & -(s))
+
+void Transpose10(marian::Tensor out, const marian::Tensor in) {
```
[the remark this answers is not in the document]
XapaJIaMnu: I need to access a transpose operation in the `expression_graph_packable.h`. As far as I could tell, the normal `tranpose` function, requires a `graph->forward()` and I couldn't figure out how to do this just for one tensor and one operation. Is that possible?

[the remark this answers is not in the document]
emjotde: Since this is a copy, why would the copy work and the original wouldn't? Or what am I missing?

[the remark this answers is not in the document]
XapaJIaMnu: Because the original is defined in `tensors/cpu/tensor_operators.cpp` and can't be included in this file. The alternative is to move the transposes to a separate header, but that change was even more invasive. I thought you'd know a better way of doing it.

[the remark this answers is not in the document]
ykim362: How about doing a slow transpose without any vectorized code? By using two for loops. This is always done offline and it might still be within a few seconds even with a plain transpose. @XapaJlaMnu @emjotde

emjotde:
Another problem, since this PR removes code from our repo and puts it in a different repo, this now circumvents our PR reviewing. That's kinda problematic.

emjotde:
@ykim362 FBGEMM hd to go through some kind approval process; do I remember that correctly?

kpu:
> Another problem, since this PR removes code from our repo and puts it in a different repo, this now circumvents our PR reviewing. That's kinda problematic.

Last time we reviewed the matrix multiply code, you said "What dark magic is this?"  

Submodules pin a specific revision.  Marian master is not going to change under your feet.  Updates to the revision will trigger a pull request and you can choose how much to look at the diff.  

The Marian review process, at least for non-Microsoft people, is way too slow.

emjotde:
Luckily we have now more competent people than me like @ykim362  or @frankseide  who can parse code like that. 

I think one other comment got lost that I had (made it on the submodule handle, and apparently that disappeared?). I put some effort into making the int16 code a fallback in cases where no BLAS implementation is available on the CPU side.  That way Marian was standalone. I would prefer to have that back in the main code repo and not in a submodule. Would making a copy of the code and updating KenLM-in-Moses-style work? That way we also get periodic pull-requests whenever you are ready sync.

ykim362:
> FBGEMM had to go through some kind approval process; do I remember that correctly?
@emjotde  Yes, we can do the same thing for the intgemm. I will take care of that.

emjotde:
Or we submodule against a fork of intgemm in the marian-nmt organization like we did with fbgemm. Same principle concerning periodic updates and PRs. Likely better than the copy, since then we can share back to the original repo.

emjotde:
@ykim362 Maybe wait until we have figured out the copy/fork/submodule thing.

snukky:
I think the approach with submodules is better and cleaner than copying, especially for code that can be compiled optionally.
I'd like to note that only marian and marian-dev have restricted write access, so getting changes into any other repo within marian-nmt should not be so slow.

emjotde:
@snukky not so sure about the slowness part though :) If the goal is to circumvent slow PR review then this is not really acceptable. What would help is if more people review code that does not directly concern them.

emjotde:
Since the complaining made me check: @frankseide @snukky @ykim362  and I have have review over 300 PRs, >160 public about 150 internal. That's one every two days.

kpu:
The int16 multiply path exists in intgemm and could be used as a backup for matrix multiply if you want.

It doesn't make sense to lock down intgemm master so even I can't approve changes to it.  There was a 17-day delay in this pull request.  

Besides, I'm working on getting intgemm other customers: https://github.com/kpuatamazon/incubator-mxnet/tree/intgemm/3rdparty https://github.com/kpuatamazon/sockeye/blob/heafield-quantize/sockeye/quantization.py#L170

emjotde:
Of course, and since we are still all working on it it's going to hang here a while longer still than the 17 days. 

I am not requiring you in any way to not maintain your original repo for whatever purposes. I am just gravitating towards forking and locking down the fork with reviewed PRs.

kpu:
You can fork it into marian-nmt.  Please don't use the fork for local Marian changes though, like happened to https://github.com/marian-nmt/FBGEMM .  You can submit those for review, by me, to the main repo.

emjotde:
@ykim362 I am actually not quite sure what the divergence from FBGEMM is? It think we contributed quite a bit back, especially windows related?

emjotde:
Anyway, off-topic. So a fork would be acceptable if we try to make the contributions well-behaved and bidirectional?

src/layers/generic.cpp
```diff
@@ -258,7 +260,31 @@ namespace marian {
...
+          if (!isLegacyUntransposedW) {
+            Wt_ = transpose(Wt_);
+            isLegacyUntransposedW = true;
```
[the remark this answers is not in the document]
XapaJIaMnu: Wouldn't a transpose be memoized  and happen only once?

src/common/binary.cpp
```diff
@@ -58,12 +59,19 @@ void loadItems(const void* current, std::vector<io::Item>& items, bool mapped) {
 
   for(int i = 0; i < numHeaders; ++i) {
     if(items[i].mapped) { // memory-mapped, hence only set pointer
+      ABORT_IF(isIntgemm(items[i].type), "mmap format not supported for intgemm matrices");
```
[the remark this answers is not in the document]
XapaJIaMnu: Mostly, because I couldn't figure out how to test memoized models. How do I do that? I will add support for it then.

XapaJIaMnu:
Hey @emjotde ,

Apologies for the late reply, we were delayed due to WNGT work. I have addressed almost all points of your pull request. For the remaining points.
 
How do we add a fork of intgemm to marian-nmt? I am not authorised to do that, where do I submit a request?

With regards to the code duplication of the transpose, I found no way to access it without invasive code changes. More information in the comments above.

With regards to mmap'd intgemm binary format not being supported, I couldn't figure out how to test this feature so I have not implemented it. Furthermore, as we dynamically prepare the proper matrix layout dependent on the architecture, I believe mmap'd format doesn't make much sense.

Cheers,

Nick

snukky:
> How do we add a fork of intgemm to marian-nmt? I am not authorised to do that, where do I submit a request?

Forked into https://github.com/marian-nmt/intgemm. All members of our organization should already have write access. Let me know if it's not the case.

XapaJIaMnu:
Thank you, @snukky, the submodule now points to marian-nmt's fork.

src/CMakeLists.txt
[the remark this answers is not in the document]
snukky: Whichever the path will be, could you add a comment why it's needed as explained in the response above?

src/common/binary.cpp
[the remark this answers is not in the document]
snukky: Please add a comment.

src/common/config_parser.cpp
```diff
@@ -629,8 +629,12 @@ void ConfigParser::addOptionsTranslation(cli::CLIWrapper& cli) {
   addSuboptionsDevices(cli);
   addSuboptionsBatching(cli);
 
-  cli.add<bool>("--optimize",
-      "Optimize speed aggressively sacrificing memory or precision");
+  cli.add<bool>("--optimize16",
```
snukky, line 632: Does it make sense to have both --optimize16 and --optimize8 together? Which one takes the effect then? If not, maybe use a single option `--optimize 8bit/16bit` or `--optimize intgemm8/intgemm16` with an implicit value?

src/common/config_parser.cpp
```diff
@@ -629,8 +629,12 @@ void ConfigParser::addOptionsTranslation(cli::CLIWrapper& cli) {
...
+      "Optimize speed aggressively sacrificing memory or precision by using 16bit integer GEMM with intgemm instead of floats. Only available on CPU");
```
snukky, line 633: If it's only available on CPU, shall we add these options only to a build with CPU support?

src/common/config_parser.cpp
```diff
@@ -629,8 +629,12 @@ void ConfigParser::addOptionsTranslation(cli::CLIWrapper& cli) {
...
+  cli.add<bool>("--optimize8",
+      "Optimize speed even more aggressively sacrificing memory or precision by using 8bit integer GEMM with intgemm instead of floats. Only available on CPU");
+  cli.add<bool>("--intgemm-shifted",
+      "Use a faster, shifted GEMM implementation. Only available with intgemm8.");
```
snukky, line 637: Does `intgemm8` refer to `--optimize8`? If it applies only to `--optimize8`, could it be collapsed into `--optimize 8bit-shifted` as suggested in my comment above?

src/data/shortlist.h
```diff
@@ -257,6 +257,13 @@ class LexicalShortlistGenerator : public ShortlistGenerator {
       for(auto& it : data_[i])
         indexSet.insert(it.first);
     }
+    // Ensure that the generated vocabulary items from a shortlist are a multiple-of-eight
```
snukky, line 260: Could you also add why this is required/what does require this?

src/common/config_parser.cpp
kpu: I think these should exist in GPU builds with an error message if used, in preparation for 8-bit int tensor core support on Turing and newer arches.

snukky: Good point. We already have `--fp16`, which corresponds to `--precision float16`. Maybe `--optimize8/16` could/should be combined with the `--precision` option? @emjotde, what is the convention?

XapaJIaMnu: I personally like the --precision integer8/integer16/integer8-shifted, but will wait for @emjotde point.

XapaJIaMnu:
@snukky , addressed everything bug the command line options and the conditional compilation. At the moment we have `-DCOMPILE_CPU=OFF`, which disables everything and `-DUSE_FBGEMM=ON` which enables FBGEMM when available. Intgemm compilation is fairly fast, compared to fbgemm, so I guess it can tag along the general CPU path?

As for the `include_directories(${CMAKE_BINARY_DIR}/src/3rd_party/intgemm)`, intgemm generates its configuration over there (what features the compiler supports). Is this acceptable?

XapaJIaMnu:
Merged with master and fixed a rare issue when loading binary models when memory is accidentally unaligned. Any comments?

CHANGELOG.md
```diff
@@ -17,6 +17,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Training and scoring from STDIN
 - Support for reading from TSV files from STDIN and other sources during training
   and translation with options --tsv and --tsv-fields n.
+- Shortlist is now always multiple-of-eight
+- Changed the `--optimize` switch to `--optimize16` and replaced the backend computation to intgemm
+- Added `--optimize8` `--intgemm-shifted` that allows for fast 8bit integer decoding with intgemm
```
snukky, line 22: I think intgemm options has changed. Could you update it here?

src/common/aliases.cpp
```diff
@@ -37,6 +37,30 @@ void ConfigParser::addAliases(cli::CLIWrapper& cli) {
...
+    }
+  });
+
+  cli.alias("int8shift", "true", [&](YAML::Node& config) {
+    if(mode_ == cli::mode::training) {
+      ABORT("int8shift GEMM mode is not supported for training");
+    } else {
+      config["gemm-precision"] = std::string("int8shift");
+    }
+  });
```
snukky, line 62: These options are not created in the training mode, so `marian` exacutable is broken and always produces:
```
Error: Option 'int16' is not defined so alias can not be created
```
or similar. Simply do not define aliases for the trainer, or define those options in all modes.

src/common/config_parser.cpp
```diff
@@ -693,12 +699,18 @@ void ConfigParser::addOptionsScoring(cli::CLIWrapper& cli) {
...
-  cli.add<bool>("--optimize",
-      "Optimize speed aggressively sacrificing memory or precision");
+  cli.add<bool>("--int16",
+      "Optimize speed aggressively sacrificing memory or precision by using 16bit integer GEMM with intgemm instead of floats. Only available on CPU. Corresponds to --gemm-precision int16");
+  cli.add<bool>("--int8",
+      "Optimize speed even more aggressively sacrificing memory or precision by using 8bit integer GEMM with intgemm instead of floats. Only available on CPU. Corresponds to --gemm-precision int8");
+  cli.add<bool>("--int8shift",
+      "Use a faster, shifted 8bit GEMM implementation. Corresponds to --gemm-precision int8shift");
+  cli.add<std::string>("--gemm-precision",
+      "Use lower precision for the GEMM operations only. Supported values: float32, int16, int8, int8shift", "float32");
```
snukky, line 709: Instead of copy-pasting, please extract to a subroutine like the `addSuboptionsBatching(cli)` above, etc. Please keep lines shorter than 100 characters if possible.

CHANGELOG.md
XapaJIaMnu: Updated.

src/common/aliases.cpp
XapaJIaMnu: Sorry about that, I didn't realise exactly the config parser works.

src/common/config_parser.cpp
XapaJIaMnu: Resolved

src/data/shortlist.h
XapaJIaMnu: Changed

src/common/binary.cpp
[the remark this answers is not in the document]
XapaJIaMnu: Further to the previous comment, we reorder the matrices at load time depending on the architecture, so memory mapping doesn't make sense for us.

src/tensors/cpu/backend.h
```diff
@@ -12,6 +12,8 @@ namespace cpu {
 class Backend : public marian::Backend {
 protected:
   bool optimized_{false};
+  bool optimized8_{false};
```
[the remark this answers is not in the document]
XapaJIaMnu: Changed to variables with more descriptive names.

src/tensors/cpu/backend.h
```diff
@@ -21,6 +23,12 @@ class Backend : public marian::Backend {
   // for CPU & inference only, sets to use optimized code for inference. Does nothing for GPU.
   void setOptimized(bool optimize) override { optimized_ = optimize; }
   bool isOptimized() override { return optimized_; }
+
+  void setOptimized8(bool optimize) override { optimized8_ = optimize; }
+  bool isOptimized8() override { return optimized8_; }
+
+  void setShifted(bool shifted) override { shifted_ = shifted; }
```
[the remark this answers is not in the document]
XapaJIaMnu: Some updated comments

XapaJIaMnu:
@snukky I fixed the issues and resolved most review comments. The ones not resolved, I either need more feedback or I am not exactly sure what to do or I am waiting for approval from the one who commented on them.

snukky:
@XapaJIaMnu Since we have a workflow running on Windows, could you update your branch with the current master and check if there is no issues when compiling on Windows? Worflow jobs will start automatically after merge. Thanks!

snukky:
As we mentioned yesterday, please take a look at the compilation errors on Windows. Details inside checks.

kpu:
Achievement unlocked, Microsoft employee help requested.
```
##[error]D:\a\marian-dev\marian-dev\src\3rd_party\intgemm\sse2_gemm.h(54): fatal error C1001: Internal compiler error.
  (compiler file 'msc1.cpp', line 1533)
   To work around this problem, try simplifying or changing the program near the locations listed above.
  If possible please provide a repro here: https://developercommunity.visualstudio.com 
  Please choose the Technical Support command on the Visual C++ 
   Help menu, or open the Technical Support help file for more information
  INTERNAL COMPILER ERROR in 'C:\PROGRA~2\MICROS~1\2019\ENTERP~1\VC\Tools\MSVC\1426~1.288\bin\Hostx64\x64\cl.exe'
      Please choose the Technical Support command on the Visual C++
      Help menu, or open the Technical Support help file for more information
```

kpu:
Almost done with MSVC support. @XapaJIaMnu what is going on here with the duplicate graph `getBackend` code?  

https://github.com/marian-nmt/marian-dev/blob/e2ddc8dfc858593a8bb326a1e41668c3608ceb34/src/command/marian_conv.cpp#L64-L82

kpu:
@ykim362 Can you test I got CPUID right on Windows? Basically print intgemm::kCPU out some windows boxes.  

https://github.com/kpu/intgemm/blob/1ba086e1e380269471b1d023a4f0781afb6bf7ca/intgemm.h#L420

ykim362:
@kpu Sure, I am trying to build intgemm on windows. The integemm branch doesn't work right away for me. Do you have a build setting which works on windows?

kpu:
@ykim362 I'm confused, the continuous integration is reporting that windows is passing for branch `intgemm_reintegrated`.  If there's more for Windows because the environment is different etc, let me know.

emjotde:
@ykim362 We are building for Windows with cmake in the check. Are you using the manually created VSC files?

ykim362:
@kpu @emjotde I was trying to build intgemm stand-alone. I used cmake on intgemm project. I will try to build as a marian package. I am not creating a vcproj file. :)

kpu:
Time for me to setup windows ci directly...

emjotde:
Roman is on vacation this week. He could probably help otherwise as he set up the Windows CI check. I unfortunately do not know what is going on there.

ykim362:
I am seeing lots of various errors with windows cmake. I guess some cmake settings need to be changed for the marian project, too.

kpu:
intgemm and all the tests now compile without warnings on MSVC: https://github.com/kpu/intgemm/runs/972975172?check_suite_focus=true
I'm not sure why the Windows build of intgemm is with debugging given that cmake was configured with
```
if(NOT CMAKE_BUILD_TYPE)
  set(CMAKE_BUILD_TYPE Release)
endif()
```

ykim362:
@kpu I tested on 3 windows machines (AVX2, AVX512BW, AVX512_VNNI). I put a print at the beginning of intgemm benchmark.cc file. All of the three CPU types are correctly detected and printed.

ykim362:
@kpu If I read the benchmark correctly, vnni looks slower than avx512 sometimes. Is it something expected?
Multiply        8       2048    256     Samples=75
     8-bit SSSE3         0.0074873      0.0075339       4.52903e-05
      8-bit AVX2         0.0041518      0.00417263      2.97038e-05
  8-bit AVX512BW          0.000452      0.00045474      7.36317e-07
8-bit AVX512VNNI          0.000974      0.000977904     1.48219e-06
     16-bit SSE2         0.0090992      0.00920425      6.73663e-05
     16-bit AVX2         0.0049246      0.00496485      4.56614e-05
   16-bit AVX512         0.0049536      0.00501178      7.03943e-05

kpu:
@ykim362 There's a rewrite with flexible tile size going on in https://github.com/kpu/intgemm/tree/static/tile .  VNNI lets me get away with more registers for larger tiles that seem to work better.  Also we've noticed VNNI hates the signed * signed (with some sign bit manipulation in the hot loop) but signed * unsigned seems ok, which is what Marian can use in this pull request.

ykim362:
@kpu  Got it. Thanks!

kpu:
It's in MXNet. . .

emjotde:
We will try to get this in this week. I plan to do a review of the options etc. tomorrow.

subbuss:
At the Wikimedia Foundation, some of us have been watching this PR with interest for a while now since this will potentially make it simpler to deploy MarianNMT on standard hardware. No pressure, but do you all have an estimate on when this might get completed and merged?

emjotde:
That's hanging on me at the moment. I am starting to look into it and want to simplify the interface a bit (on the code level). There is also fp16 training coming in a different merge, so I do not want those to things to spread in different directions. I don't have a ETA, but hope to start working on that rather sooner than later.

snukky:
@subbuss FYI. AFAIK, features from this PR have been already tested in multiple scenarios and are used in another project, so if you are only a consumer of Marian (I mean no significant modifications of the code), I think this is quite ready for a deployment. There is even a follow-up work to this PR with more performance improvements and @XapaJIaMnu should be able to provide more information on that. If you do modify Marian's backend in your application though, waiting for merging this PR might be a good idea due to the mentioned upcoming refactorization prior to the merge.

kpu:
> There is also fp16 training coming in a different merge, so I do not want those to things to spread in different directions.

Then you should commit to timely review and merge.  

>  I think this is quite ready for a deployment. 

This is a reason to merge.  

> upcoming refactorization prior to the merge.

The caching as implemented in Marian needs to die.  We all agree on that.  What I don't agree with is blocking contributions for months so that you can get your change in first in order to foist the merging effort on people trying to contribute.  

We in Edinburgh are at the breaking point of running away from Marian or at least abandoning any attempt to push to this repo.

XapaJIaMnu:
@subbuss , if you are interested in followup work, you can check this branch: https://github.com/XapaJIaMnu/marian-dev/tree/8bitgpu

Instructions how to run produce and run the even more optimised intgemm models can be found here: https://github.com/browsermt/students/tree/master/train-student section 5.3 I'm happy to help out and answer questions via github issues/email.

subbuss:
> @subbuss , if you are interested in followup work, you can check this branch: https://github.com/XapaJIaMnu/marian-dev/tree/8bitgpu

Thanks! I'll pass along the info!

emjotde:
Alrighty, back to this one then.

XapaJIaMnu:
I added some comments. The main issue is that as far as I can tell the architecture agnostic format s hidden behind an ABORT statement inside expression_graph_packable, as well the lack of documentation of the options in marian-conv

emjotde:
@XapaJIaMnu I don't see any comments?

XapaJIaMnu:
> @XapaJIaMnu I don't see any comments?

Ugh, forgot the submit review button... how about now?

emjotde:
Replied. 

As for the croaking, we might leave it as is for a couple of days and then unify detection with what FBGEMM is doing? I think that overlaps largely?

kpu:
> As for the croaking, we might leave it as is for a couple of days and then unify detection with what FBGEMM is doing? I think that overlaps largely?

Architecture-independent is the point.  We have no plan to use the architecture-dependent mmap format.

emjotde:
The croaking would be happening if there is an explicit request for a hardware-specific action that the current hardware doesn't support. Since mmapping is now possible, both modes make sense.

emjotde (approved):

CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 
 ### Added
+- Added `intgemm8(ssse3|avx|avx512)?`, `intgemm16(sse2|avx|avx512)?` types to marian-conv with uses intgemm backend. Types intgemm8 and intgemm16 are hardware-agnostic, the other ones hardware-specific.
```
snukky, line 12: avx2 are missing?

XapaJIaMnu: There's no avx actually, it's avx2.

kpu:
> That is a bit wonky, but I am also OK with it. The main problem would be that we would need a way to ask intgemm about the arch it is currently running on in order to be able to change the type to intgemm8avx2.

`extern CPUType kCPU;`

```
// If you want to detect the CPU and dispatch yourself, here's what to use:
enum class CPUType {
  UNSUPPORTED = 0,
  SSE2,
  SSSE3,
  AVX2,
  AVX512BW,
  AVX512VNNI
};
```

XapaJIaMnu:
> > That is a bit wonky, but I am also OK with it. The main problem would be that we would need a way to ask intgemm about the arch it is currently running on in order to be able to change the type to intgemm8avx2.
> 
> In `intgemm.h` there is:
> `extern CPUType kCPU;`
> 
> ```
> // If you want to detect the CPU and dispatch yourself, here's what to use:
> enum class CPUType {
>   UNSUPPORTED = 0,
>   SSE2,
>   SSSE3,
>   AVX2,
>   AVX512BW,
>   AVX512VNNI
> };
> ```

I'll add this tonight

CHANGELOG.md
emjotde: typo

emjotde:
Good to have for the croaks, but not even sure we need it to change the types on-the-fly.

src/common/binary.cpp
```diff
@@ -58,6 +58,11 @@ void loadItems(const void* current, std::vector<io::Item>& items, bool mapped) {
   get<char>(current, offset);
 
   for(int i = 0; i < numHeaders; ++i) {
+    // For intgemm AVX512 and AVX512VNNI have the same arangement, but the VNNI algorithm is faster.
+    // Change the type to the fastest one supported.
+    if (items[i].type == Type::intgemm8avx512) {
```
XapaJIaMnu, line 63: Would this be able to persist after mmap'ing? @emjotde

emjotde:
@snukky @XapaJIaMnu now that EMNLP/WMT is over, how are the regression tests going, etc. Where do we stand here?

snukky:
Regression tests are finished for some time already and wait for testing by @XapaJIaMnu (https://github.com/marian-nmt/marian-regression-tests/pull/56). Other than that, there are a very few unresolved comments (not sure if important), and I think it's good to go.

snukky (approved):
Tests are ready, I think we can merge.

kpu:
Woah is @emjotde gonna merge?

emjotde:
About to catch up with internal and external PRs and this one is first in line.

PR #571, opened 2019-12-17
Remove unused variables

XapaJIaMnu:
graph_->setOptimized/isOptimized is no longer used. This functionality has moved to the Backend, but the methods still exist in the graph_ which always return false
```
grep -R "Optimized("
translator/translator.h:          graph->getBackend()->setOptimized(options_->get<bool>("optimize"));
translator/translator.h:        graph->getBackend()->setOptimized(options_->get<bool>("optimize"));
tests/prod.cpp:        g->getBackend()->setOptimized(false);
tests/prod.cpp:        g->getBackend()->setOptimized(true);
tensors/gpu/backend.h:  // for GPU, this is invalid. for gpu, isOptimized() function always returns false.
tensors/gpu/backend.h:  void setOptimized(bool optimize) override {
tensors/gpu/backend.h:    LOG_ONCE(info, "setOptimized() not supported for GPU_{}", optimize);
tensors/gpu/backend.h:  bool isOptimized() override {
tensors/cpu/backend.h:  void setOptimized(bool optimize) override { optimized_ = optimize; }
tensors/cpu/backend.h:  bool isOptimized() override { return optimized_; }
tensors/backend.h:  // for GPU, this is invalid. for gpu, isOptimized() function always returns false.
tensors/backend.h:  virtual void setOptimized(bool optimize) = 0;
tensors/backend.h:  virtual bool isOptimized() = 0;
rescorer/rescorer.h:        graph->getBackend()->setOptimized(options_->get<bool>("optimize"));
microsoft/quicksand.cpp:  graph->getBackend()->setOptimized(false);
graph/expression_graph.h:  void setOptimized(bool optimized) { optimized_ = optimized; }
graph/expression_graph.h:  bool isOptimized() { return (optimized_ && inferenceOnly_); }
graph/expression_operators.cpp:      if(a->graph()->getBackend()->isOptimized()) {
graph/expression_operators.cpp:      if(a->graph()->getBackend()->isOptimized()) {
command/marian_conv.cpp:  graph->getBackend()->setOptimized(false);
```
This patch removes the unused variable from the graph class.

Cheers,

Nick

PR #598, opened 2020-02-07
Add support for compiling on Mac (and clang)

XapaJIaMnu:
Hey guys,

This patchset adds a support for compiling on mac (marian cpuonly). Decoding was tested to work with OpenBLAS. I don't have a mac capable of running fbgemm (but it compiles). 

Intgemm should also work.

What was necessary:
1) a lot places in the code define `__unix__ ` and `__win32__` bits, which don't capture MacOS. Fixed this particular issue at multiple locations.
2) Pathie-cpp fixes (just applied the upstream pathie-cpp fixes that I did a year ago)
3) Mac has a weird `aligned_alloc`. Used `posix_memalign` instead.
4) Mac defines `size_t` differently from Windows and Unix.
5) Fix detection of Openblas (installed via brew)
6) Clang doesn't support a few of the options that are passed to gcc, so those were disabled.

For Sentencepiece, protobuf was installed via brew.

There are millions of warnings when compiling, but it works. This should also allow compilation with clang on Linux.

Cheers,

Nick


kpu:
This should be a straightfoward accept but it touches a few files so should be prioritized for review.

src/3rd_party/pathie-cpp/src/entry_iterator.cpp
```diff
@@ -31,7 +31,7 @@
 #include "../include/path.hpp"
 #include "../include/errors.hpp"
 
-#if defined(__unix__)
+#if defined(__unix__) || defined(__APPLE__)
```
frankseide: I do not like this define __APPLE__ too much (if we adopt this pattern, it will proliferate if we add more platforms). Is there a common symbol? Why is __unix__ not defined on Mac?

src/3rd_party/pathie-cpp/src/pathie.cpp
```diff
@@ -143,7 +143,7 @@ std::string Pathie::convert_encodings(const char* from_encoding, const char* to_
     errno  = 0;
     errsav = 0;
 
-#ifdef BSD
+#if defined(BSD) && ! defined(__APPLE__) //Since MacOS evolved from BSD, it is captured here but the iconv on macos behaves differently
```
frankseide: Is this distinction needed at all? Can we not always use the version with the const_cast, and leave the comment in to explain why? It seems the other arguments are all the same?

src/3rd_party/pathie-cpp/src/pathie.cpp
```diff
@@ -181,11 +181,10 @@ std::string Pathie::convert_encodings(const char* from_encoding, const char* to_
 std::string Pathie::utf8_to_filename(const std::string& utf8)
 {
   bool fs_encoding_is_utf8 = false;
-
+  char* fsencoding = NULL;
```
frankseide: Why was this moved up?

src/3rd_party/pathie-cpp/src/pathie.cpp
```diff
@@ -206,11 +205,10 @@ std::string Pathie::utf8_to_filename(const std::string& utf8)
 std::string Pathie::filename_to_utf8(const std::string& native_filename)
 {
   bool fs_encoding_is_utf8 = false;
-
+  char* fsencoding = NULL;
```
frankseide: same question here--why moved up? If not needed, please could you move it back; the original seemed more local.

src/3rd_party/pathie-cpp/src/path.cpp
```diff
@@ -902,7 +902,7 @@ Path Path::pwd()
  */
 Path Path::exe()
 {
-#if defined(__linux__)
+#if defined(__linux__) || defined(__APPLE__)
```
frankseide: BTW these changes are in an external library. Is there possibly an update that already does this? If not, are we going to contribute back?

src/common/fastopt.cpp
```diff
@@ -88,6 +88,7 @@ template struct As<std::vector<int>>;
 // So, we need an explicit definition for uint64_t. Otherwise, there's a linking error on windows.
 // https://software.intel.com/en-us/articles/size-of-long-integer-type-on-different-architecture-and-os/
 template struct As<std::vector<uint64_t>>;
+template struct As<std::vector<size_t>>; //MacOS support
```
frankseide: Why is that not already covered? Is size_t != uint64_t on MacOS? Please explain this in the line-end comment.

Could you also please have a space after // like everywhere else?

src/common/fastopt.h
```diff
@@ -367,12 +367,17 @@ class FastOpt {
   }
 
   const FastOpt& operator[](const char* const key) const {
+#ifdef __APPLE__
```
frankseide: Why is this change needed? And if it is needed, can we always use the __APPLE__ path? It seems identical.

src/common/utils.cpp
```diff
@@ -8,12 +8,22 @@
...
 #include <string>
 #include <set>
-#ifdef __unix__
+#if defined(__unix__) || defined(__APPLE__)
 #include <unistd.h>
 #endif
 #include <codecvt>
 #include <cwctype>
 
+//MACOS lacks HOST_NAME_MAX
```
frankseide: space after //

src/tensors/cpu/device.cpp
```diff
@@ -17,6 +17,15 @@ namespace cpu {
 // functions which tests for alignment. 
 #ifdef _WIN32
 #define MALLOC(size) _aligned_malloc(size, alignment_)
+#elif __APPLE__
+//On macos, aligned_alloc is available only on c++17
+//Furthermore, it requires that the memory requested is an exact multiple of the alignment, otherwise it fails
+void * apple_aligned_alloc(size_t alignment_, size_t size) {
```
frankseide: Is posix_memalign the same as aligned_alloc on Linux? Then we should use it in both cases, and condition on the correct #define (something POSIX I presume)

src/3rd_party/pathie-cpp/src/pathie.cpp
XapaJIaMnu: `fsencoding` is defined only in the `else` part of the `ifdef`, resulting in a compilation error when this preprocessor path is taken. It's an oversight by the author, I had the same discussion with him when I proposed this change upstream.

src/3rd_party/pathie-cpp/src/entry_iterator.cpp
XapaJIaMnu: I have no idea. If you have a mac lying around, you can try to do `gcc -dM -E - </dev/null`
Neither clang, nor gcc define `unix` on mac. I am open to suggestions. Looking through fbgemm code, they use `__linux__` and `__APPLE__` defines.

src/3rd_party/pathie-cpp/src/pathie.cpp
XapaJIaMnu: Same as the above, it's defined inside of the `else` branch of an `ifdef` so the code doesn't compile if the preprocessor takes the other branch.

src/3rd_party/pathie-cpp/src/path.cpp
XapaJIaMnu: I have already contributed this upstream about a year ago: https://github.com/Quintus/pathie-cpp/pull/11

It seems that the marian version has diverged quite a bit since then...

src/3rd_party/pathie-cpp/src/pathie.cpp
frankseide: Thanks

src/3rd_party/pathie-cpp/src/pathie.cpp
frankseide: Thanks

src/tensors/cpu/device.cpp
XapaJIaMnu: `aligned_alloc` is the same as `posix_memalign` if you are using the standard C++ libraries. I believe, if you use `libc++` it might be different. At any rate, `aligned_alloc` is stricter than `posix_memalign` and you can definitely use it on linux. If you prefer, I can merge those two codepaths. I do find `aligned_alloc` more C++ style.

frankseide:
"There are millions of warnings when compiling, but it works. This should also allow compilation with clang on Linux."

I am not a fan of warnings. I would request that all warnings in our own code are resolved, while it is OK to disable warnings in 3rd-party code after checking them once.

Warnings should be fixed by minimizing changes. I.e. not just adding `#ifdef __APPLE__` around it. My experience is that most warnings can be fixed cleanly (I have done that each time I try to build a latest version of Marian on the Windows, which has some different ideas about warnings compared to gcc...)

src/3rd_party/pathie-cpp/src/pathie.cpp
XapaJIaMnu: I have not tried it, maybe. I will try tomorrow.

src/common/fastopt.cpp
XapaJIaMnu: I have no idea, to be honest. Both of the fastopt changes are pretty cryptic to me. Without this change, the code fails during linking, because of undefined reference to a certain type. Inspecting the `libmarian.a` revealed that it exported 127 symbols of this type vs a 130 on my linux box. Adding `size_t` made the number of symbols consistent.  Normally, they should be identical, but when compiling marian they are not.

src/common/fastopt.h
XapaJIaMnu: Same as the above issue. I do believe we can use the linux codepath everywhere.

XapaJIaMnu:
> "There are millions of warnings when compiling, but it works. This should also allow compilation with clang on Linux."
> 
> I am not a fan of warnings. I would request that all warnings in our own code are resolved, while it is OK to disable warnings in 3rd-party code after checking them once.
> 

Me neither, but it is really too many of them. We can start addressing them one by one, but it will take a while. Here is a a [compile log](https://nbogoychev.com//files/marian_compile_errors_mac.gz) on mac. Most of the errors complain about unused variables and fields, as delete called on a non-final method.

src/common/fastopt.cpp
kpu: "The type size_t is an implementation-defined unsigned integer type that is large enough to contain the size in bytes of any object." https://stackoverflow.com/questions/32021860/c-should-you-size-t-with-a-regular-array

kpu:
> "There are millions of warnings when compiling, but it works. This should also allow compilation with clang on Linux."
> 
> I am not a fan of warnings. I would request that all warnings in our own code are resolved, while it is OK to disable warnings in 3rd-party code after checking them once.
> 
> Warnings should be fixed by minimizing changes. I.e. not just adding `#ifdef __APPLE__` around it. My experience is that most warnings can be fixed cleanly (I have done that each time I try to build a latest version of Marian on the Windows, which has some different ideas about warnings compared to gcc...)

In principle yes fix warnings.  But these already existed in the code; clang is just surfacing them by complaining about different things.  Mac support is a net improvement even with warnings and we shouldn't block this change on it.

frankseide:
"But these already existed in the code; clang is just surfacing them by complaining about different things."

I kindly request to fix them anyways (in original Marian code). I understand this is a voluntary contribution, but for other ports, we also required a clean build. For example, I did this when I first built it for Windows, and I have redone it each time we added a larger chunk of new code. In my experience, it does not actually take too much time, since many warnings are duplicates.

We should not fix warnings in third-party code. For that, we should instead configure the compiler to not warn.

* For unused parameters, please enclose the name in `/* */`. We follow this convention in many places already.
* `delete called on non-final ...` is probably a legitimate issue that we should fix
* `private field not used` is something where clang helped us to find an unused field, and is easy to fix
* Invalid linker command lines should be fixed (unless they are caused by cmake; then we can disable the warning).
* I am not sure what to do about `expression result unused`. This comes from referencing a variable, e.g. `v;`. This is sometimes needed, e.g. when that variable is only used in the debug build. Here, I would be OK to explicitly disable this specific warning whole-sale.

src/3rd_party/pathie-cpp/src/pathie.cpp
XapaJIaMnu: No, the BSD codepath doesn't compile on Linux. We can't merge the two, and since BSD captures MAC, we have to add the exception for APPLE in the ifdef

src/tensors/cpu/device.cpp
kpu: @frankseide Take a look at what I just pushed.

emjotde:
As for warnings, we don't accept any PRs that result in warnings. Neither MS-internal nor external. If they slip through it's the first thing we fix. As Frank said, either by addressing the issue, or if 3rd-party code or compilers being unreasonable by disabling them. Disabling has to be well motivated though.

kpu:
@emjotde The current Marian code generates a lot of warnings when compiled on a mac.  This PR fixes the errors and should not introduce new warnings (if it does we'll fix them).  

Fixing the existing warnings is a larger project that will touch a lot of files.  Demanding Nick fix all your existing compiler warnings (as reported by clang) is IMHO unreasonable and violates the principle that changes should be small, focused, and frequently merged.  

If you want to get started on the project of fixing clang warnings, merge $571.

emjotde:
I would be OK if the code produces no new warnings on Windows or Linux with g++. So we are talking only about warning on Mac, correct?

frankseide:
How about we fix the clang warnings first, and then merge this PR? While we cannot compile on a Mac, we can compile with clang on Linux.

kpu:
Correct.  Nick is pointing out that existing Marian code produces warnings on Mac, including those generated by #571.  If there are any new warnings on Linux or Windows, feel free to object!

frankseide:
Marcin will make the decision. However, I still disagree, because I now who will have to fix the warnings in the end (me). This PR consists of two things. Compile with clang, and compile on Mac. If the first step is not ready for merging, then the second should not be merged either.

kpu:
@frankseide I want to get Mozilla to setup continuous integration on Mac including integration tests.  We can reach integration tests faster by fixing errors then fixing Mac warnings...

kpu:
@frankseide Can you accept #571 then?  It fixes some warnings on mac...

emjotde:
I am tending towards breaking this up into Clang on Linux PR and Mac-specific PR. Then we can add a Jenkins build for Clang. 

I am taking a look at #571 just now. I think it actually ready for merging.

frankseide:
I would be OK to accept this PR (assuming all other comments are addressed), and working on clang on Linux in parallel, if you guys could commit to submit a second PR that fixes any remaining Mac-specific warnings after the clang/Linux fixes. Would that work for you (Nick, Kenneth, Marcin)?

emjotde:
#571 merged.

XapaJIaMnu:
This pull request does not introduce new warnings on Linux (with GCC) and does the bare minimum (the `CMakeLists.txt` ) changes to allow compilation with clang. As far as I can tell, the vast majority of the warnings appear because we are suddenly compiling with clang, which has superior static code analysis, not because there is any mac specific issue. They do definitely occur on Linux as well (as tested today).

src/common/utils.cpp
kpu: Fixed in current version.

src/3rd_party/pathie-cpp/src/entry_iterator.cpp
frankseide: Can we use `BSD` instead of `__APPLE__`?

src/3rd_party/pathie-cpp/src/pathie.cpp
frankseide: Can't one always pass a non-const pointer where a const pointer is expected?

src/common/fastopt.cpp
frankseide: Let's remove the //MacOs support comment, and then Resolve this.

frankseide:
Does this allow to build with clang on Linux?

src/common/fastopt.cpp
XapaJIaMnu: I have updated this and explained. Having both `<size_t>` and  `<uint64_t>` break linux compilation, so I have wrapped this in an `__APPLE__` tag.

XapaJIaMnu:
> Does this allow to build with clang on Linux?

Yes. You only need the `CMakeLists.txt` changes for linux clang build. Tested it to work with clang6 (for the CPU only) and clang9 for CPU and GPU. I am not sure if clang6 can compile NCCL successfully.

frankseide:
Since size_t and uint64_t are not guaranteed to be the same, it is natural that both should be defined. We don’t actually need any comment here, and I suggest to delete the comment to avoid confusion and keep the code regular.

src/3rd_party/pathie-cpp/src/pathie.cpp
XapaJIaMnu: I don't have access to a BSD machine to test if this codepath is absolutely necessary for compilation in BSD. The upstream author of `pathie-cpp` deemed it was necessary and coded it in this way. This change was accepted by him when I added mac support upstream. I don't see a reason to deviate from it. If you want, I can remve BSD support altogether.

src/3rd_party/pathie-cpp/src/entry_iterator.cpp
XapaJIaMnu: In this particular case, we can. Pathie was designed to not compile unless the operating system was supported. There are multiple (non consistent!) defines throughtout the codebase that check for windows and have an else case `#error Unsupported system.` that doesn't compile. I can change this code to:
```c++
#ifdef _WIN32
#include <Windows.h>
#else
#include <sys/types.h>
#include <dirent.h>
#include <errno.h>
#include <stdexcept>
#endif
```
But this further deviates from upstream.

kpu:
In light of #606 are you happy to let this in then?

src/tensors/cpu/device.cpp
XapaJIaMnu: We also need to update: https://github.com/marian-nmt/FBGEMM/blob/84e66a976046180187724aff60a236c5378fde7c/include/fbgemm/Utils.h#L14 
as this won't work on clang with libc++. @ykim362 , why was the allocator changed for fbgemm? Is it just for Windows support?

kpu:
@frankseide In light of #606 are you happy to let this in then?

src/common/fastopt.cpp
```diff
@@ -87,7 +87,12 @@ template struct As<std::vector<int>>;
 // Windows and Unix based OS have different type definitions for 'unsigned long'.
 // So, we need an explicit definition for uint64_t. Otherwise, there's a linking error on windows.
 // https://software.intel.com/en-us/articles/size-of-long-integer-type-on-different-architecture-and-os/
+// We need an explicit size_t definition as well, as it is implementation defined and on MacOS we experience linker issues without it. More: https://stackoverflow.com/questions/32021860/c-should-you-size-t-with-a-regular-array
+// However, since on other platforms/implemenations size_t == uint64_t this means we get a duplicate init, so we need to wrap it in __APPLE__
 template struct As<std::vector<uint64_t>>;
+#ifdef __APPLE__
```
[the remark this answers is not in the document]
frankseide: Depending on the test, either delete it, or keep it as-is. Then please Resolve this comment.

src/common/fastopt.cpp
frankseide: Ah, got an answer from our C++ people inside. The solution is easier:

Just add a template specialization for the type that size_t expands to on Mac. Then you can eliminate the `#ifdef`.
```
template struct As<std::vector<uint32_t>>;
```

XapaJIaMnu:
@frankseide ,

Just having `size_t` works on linux and on mac, However according to the comments `uint64_t` is there for Windows compilation:
```
// Windows and Unix based OS have different type definitions for 'unsigned long'.
// So, we need an explicit definition for uint64_t. Otherwise, there's a linking error on windows.
```
 Can you guys check on your side if it compiles with just `size_t` and without `uint64_t`

Cheers,

Nick

src/common/fastopt.cpp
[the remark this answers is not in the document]
XapaJIaMnu: Added question in the comments: https://github.com/marian-nmt/marian-dev/pull/598#issuecomment-591164165

frankseide:
Windows unsigned long is uint32_t, AFAIK. So how about we just add uint64_t and uint32_t, and no size_t. I think that should cover the bases.

XapaJIaMnu:
> Windows unsigned long is uint32_t, AFAIK. So how about we just add uint64_t and uint32_t, and no size_t. I think that should cover the bases.

I couldn't get the mac version to compile without `size_t`. Adding `uint32_t` and `uint64_t` didn't do the job. How about if we have `size_t` and `uint32_t`. Would that work on Windows? It works on mac and Linux.

src/common/fastopt.cpp
XapaJIaMnu: The problem is that `uint64_t` on mac expands to `unsigned long long`, whereas on linux it expands to `unsigned long`. `size_t` on both linux and mac expands to `unsigned long`. As such `unsigned long` and `uint64_t` can't coexist on linux. A solution that would work is to make it:
```
template struct As<std::vector<unsigned long>>;
```
is this acceptable on all OSs? `uint32_t` evaluates to `unsinged int` on mac and linux.

XapaJIaMnu:
@frankseide I have merged #606 in this branch, it compiles on linux, mac (Apple Clang 10), gcc 5.4 9.2 and clang 6-8, no warnings.

XapaJIaMnu:
@frankseide , Do you have any more objections? Can we merge this?

src/common/fastopt.cpp
frankseide: If it builds, yes.

src/common/fastopt.h
frankseide: If the same code path works everywhere, then let's do that.

src/tensors/cpu/device.cpp
frankseide: I prefer the unified code path over the C++ style. Please add a comment aht posix_memalign is used instead of aligned_alloc for clang compatibility

frankseide:
I have approved it. @Marcin, I hope that’s OK with you.

Thanks,

Frank

snukky:
@XapaJIaMnu All regression tests passed.

Issue #597, opened 2020-02-07
Marian hangs on multiple GPUs, GPU utilization remains at constant 100%


rihardsk:
Tried to run Marian on a new system. When running on a single GPU, everything's fine. But when using 2 or more GPUs, the training process stalls after or during the memory reservation phase and the GPU utilization remains stuck at 100% indefinitely. This happens on the current `master`, when i try to reproduce the issue on another system with 2 GPUs, everything's fine.

To investigate, i compiled Marian in debug mode, ran it with `gdb`, and paused the execution during the stall. Here's the stacktrace i got
```
[2020-02-07 14:47:54] [comm] NCCLCommunicator constructed successfully.
[2020-02-07 14:47:54] [training] Using 2 GPUs
[Thread 0x7fffb07fc700 (LWP 24955) exited]
[2020-02-07 14:47:54] Training started
[2020-02-07 14:47:54] [data] Shuffling data
[2020-02-07 14:48:04] [data] Done reading 12469417 sentences
[2020-02-07 14:49:07] [data] Done shuffling 12469417 sentences to temp files
[2020-02-07 14:49:12] [training] Batches are processed as 1 process(es) x 2 devices/process
[2020-02-07 14:49:12] [memory] Reserving 217 MB, device gpu3
[2020-02-07 14:49:12] [memory] Reserving 217 MB, device gpu2
[2020-02-07 14:49:12] [memory] Reserving 217 MB, device gpu3
[2020-02-07 14:49:12] [memory] Reserving 217 MB, device gpu2
[2020-02-07 14:49:13] [memory] Reserving 108 MB, device gpu2
[2020-02-07 14:49:13] [memory] Reserving 108 MB, device gpu3
^C
Thread 1 "marian" received signal SIGINT, Interrupt.
0x00007ffff7ffab62 in clock_gettime ()
(gdb) where
#0  0x00007ffff7ffab62 in clock_gettime ()
#1  0x00007fffe73c1ea6 in __GI___clock_gettime (clock_id=4, tp=0x7fffffffc620)
    at ../sysdeps/unix/clock_gettime.c:115
#2  0x00007fffe443e37e in ?? () from /usr/lib/x86_64-linux-gnu/libcuda.so.1
#3  0x00007fffe45024f7 in ?? () from /usr/lib/x86_64-linux-gnu/libcuda.so.1
#4  0x00007fffe44264ac in ?? () from /usr/lib/x86_64-linux-gnu/libcuda.so.1
[...]

PR #599, opened 2020-02-11
clang: Disable unused-function warnings for 3rd-party NCCL library

kpu:
Actually a lot of warnings from clang are unused functions inside NCCL.  Not all of them, but it's a start.

### DIFF
--- a/src/3rd_party/CMakeLists.txt
+++ b/src/3rd_party/CMakeLists.txt
@@ -86,6 +86,12 @@ if(CUDA_FOUND)
     if(COMPILE_CUDA_SM70)
       set(GENCODE "${GENCODE} -gencode=arch=compute_70,code=sm_70")
     endif(COMPILE_CUDA_SM70)
+    # clang generates a lot of unused function warnings for NCCL
+    if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+      set(NCCL_CXXFLAGS "-Wno-unused-function")
+    else()
+      set(NCCL_CXXFLAGS "")
+    endif()
 
     # install nccl in ${CMAKE_BINARY_DIR}/local similar to /usr/local linux installation
     ExternalProject_Add(nccl_install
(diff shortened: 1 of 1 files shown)


kpu:
If you're going to express opinions, at least act on them.  

"I am not a fan of warnings. I would request that all warnings in our own code are resolved, while it is OK to disable warnings in 3rd-party code after checking them once." --@frankseide
"We should not fix warnings in third-party code. For that, we should instead configure the compiler to not warn." --@frankseide
"As Frank said, either by addressing the issue, or if 3rd-party code or compilers being unreasonable by disabling them. Disabling has to be well motivated though." --@emjotde
"I would be OK to accept this PR (assuming all other comments are addressed), and working on clang on Linux in parallel, if you guys could commit to submit a second PR that fixes any remaining Mac-specific warnings after the clang on Linux has been made warning-free. Would that work for you (Nick, Kenneth, Marcin)?" --@frankseide

emjotde:
Relax.

frankseide:
Sorry, did not realize that this was meant for me. Looking now.

kpu:
Thanks @frankseide

emjotde:
Reverting for now as this seems to break NCCL compilation with CUDA with g++

emjotde:
Also confirmed on my local machine, I think the CXXFLAGS overwrite is removing some options?

kpu:
Feel free to revert; I'll try again on that machine. Funny thing is I tried passing CXXFLAGS through too and that's still broken??

emjotde:
Could it be it's removing options from the Make? Like Fpic?

kpu:
Ugh https://github.com/marian-nmt/nccl/blob/b56650c7f59b8cd40d18809784a6d6be38ef8acb/makefiles/common.mk#L46

kpu:
Actually you just need to update your NCCL: https://github.com/NVIDIA/nccl/blob/master/makefiles/common.mk

emjotde:
Downgraded on purpose, higher version have a bug for models above 2GB, unfortunately. Mentioned this in Rikhards issue. 

We may need to make a special branch in our fork?

kpu:
Ugh yeah I just re-read #597.  Is a special branch of NCCL overkill for compiler warnings on clang?

PR #601, opened 2020-02-14
combine two for-loops in nth_element.cpp on CPU

PinzhenChen:
Original `nth_element.cpp` fills a vector with (accumulative) beam sizes and vocab indices for different batches/sentences. Then it uses a different for loop to read them and sort n-best scores. Since beam size is fixed, this commit combines the two for-loops to improve efficiency.

### DIFF
--- a/src/translator/nth_element.cpp
+++ b/src/translator/nth_element.cpp
@@ -36,25 +36,29 @@ public:
     size_t maxSize = N * dimBatch;
     h_res.resize(maxSize);
     h_res_idx.resize(maxSize);
+    size_t pos = 0; // iterates through h_res and h_res_idx
 
-    std::vector<int> idxs(dimBatch * inputN * vocabSize);
+    size_t batchOffset = inputN * vocabSize;
+    std::vector<int> idxs(batchOffset); // re-used for each batch
     std::iota(idxs.begin(), idxs.end(), 0);
 
     for(size_t batchIdx = 0; batchIdx < dimBatch; ++batchIdx) {
 
-      std::vector<int>::iterator begin = idxs.begin() + batchIdx * inputN * vocabSize;
-      std::vector<int>::iterator middle = begin + N;
-      std::vector<int>::iterator end = idxs.begin() + (batchIdx + 1) * inputN * vocabSize;
       std::partial_sort(
-          begin, middle, end, [&](int a, int b) { return scoresData[a] > scoresData[b]; });
+        idxs.begin(),
+        idxs.begin() + N,
+        idxs.end(),
+        [&](int a, int b) { return scoresData[a] > scoresData[b]; }
+      );
 
-      size_t pos = batchIdx * N;
-      while(begin != middle) {
-        int idx = *begin++;
-        h_res_idx[pos] = idx;
+      for(int temp = 0; temp < N; ++temp) {
+        int idx = idxs[temp];
+        h_res_idx[pos] = idx + batchIdx * batchOffset;
         h_res[pos] = scoresData[idx];
         ++pos;
       }
+
+      scoresData += batchOffset;
     }
     getPairs(/*cumulativeBeamSizes.back(),*/ outKeys, outPathScores);
   }
--- a/src/translator/nth_element.cpp
+++ b/src/translator/nth_element.cpp
@@ -8,7 +8,6 @@
 #include <iterator>
 #include <limits>
 #include <numeric>
-#include <iostream>
 
 namespace marian {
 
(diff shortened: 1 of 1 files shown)


src/translator/nth_element.cpp
```diff
@@ -59,23 +31,31 @@ class NthElementCPU {
...
     h_res_idx.resize(maxSize);
 
-    selectNBest(scores->data(), batchFirstElementIdxs, cumulativeBeamSizes);
+    std::vector<int> idxs(dimBatch * inputN * vocabSize);
+    std::iota(idxs.begin(), idxs.end(), 0);
+
+    for(size_t batchIdx = 0; batchIdx < dimBatch; ++batchIdx) {
+
+      std::vector<int>::iterator begin = idxs.begin() + batchIdx * inputN * vocabSize;
+      std::vector<int>::iterator middle = begin + (int)N;
```
kpu: Remove (int) here.  N is size_t which is the right type for adding to begin.

src/translator/nth_element.cpp
```diff
@@ -59,23 +31,31 @@ class NthElementCPU {
...
+      std::vector<int>::iterator end = idxs.begin() + (batchIdx + 1) * inputN * vocabSize;
+      std::partial_sort(
+          begin, middle, end, [&](int a, int b) { return scoresData[a] > scoresData[b]; });
+
+      int pos = batchIdx * (int)N;
```
kpu: Let's change this to size_t pos and also remove the (int).

src/translator/nth_element.cpp
```diff
@@ -59,23 +31,31 @@ class NthElementCPU {
...
-    ABORT_IF(cumulativeBeamSizes.back() != dimBatch * N, "cumulativeBeamSizes.back() wrong??");
+    const float* scoresData = scores->data();
 
     size_t maxSize = N * dimBatch;
     h_res.resize(maxSize);
     h_res_idx.resize(maxSize);
 
-    selectNBest(scores->data(), batchFirstElementIdxs, cumulativeBeamSizes);
+    std::vector<int> idxs(dimBatch * inputN * vocabSize);
+    std::iota(idxs.begin(), idxs.end(), 0);
```
kpu: I know this was in the pre-existing code, but I think it can be optimized further.  
```
std::vector<int> idxs(inputN * vocabSize);
std::iota(idxs.begin(), idxs.end(), 0);
```
Then in the loop reuse the same indices for each batch.  
```
const float *batchScores = scoresData + batchIdx * inputN * vocabSize;
std::partial_sort(idxs.begin(), idxs.begin() + N, idxs.end(), [&](int a, int b) { return batchScores[a] > batchScores[b]; }
```
If hypotheses have ties this could make tie-breaking batch dependent but it wasn't guaranteed to be anything consistent anyway.  
```

**Comment by kpu** [comment]
@snukky Can you look at this for inclusion?

File: src/translator/nth_element.cpp [hunk 4]
```diff
@@ -59,23 +31,35 @@ class NthElementCPU {
...
+    size_t pos = 0; // iterates through h_res and h_res_idx
+
+    size_t batchOffset = inputN * vocabSize;
+    std::vector<int> idxs(batchOffset); // re-used for each batch
+    std::iota(idxs.begin(), idxs.end(), 0);
+
+    for(size_t batchIdx = 0; batchIdx < dimBatch; ++batchIdx) {
 
-    selectNBest(scores->data(), batchFirstElementIdxs, cumulativeBeamSizes);
+      std::partial_sort(
```
snukky: Could you add a short in-line comment for the sort?

src/translator/nth_element.cpp
```diff
@@ -59,23 +31,35 @@ class NthElementCPU {
...
+        idxs.begin(),
+        idxs.begin() + N,
+        idxs.end(),
+        [&](int a, int b) { return scoresData[a] > scoresData[b]; }
+      );
+
+      for(int temp = 0; temp < N; ++temp) {
```
snukky: I would suggest not using the name `temp` if it's just for iteration. `temp` suggests that it will be used to temporary store an intermediate value during some computation, but here we just have an int iterator.

Also, it can be size_t instead of int.

src/translator/nth_element.cpp
```diff
@@ -59,23 +31,35 @@ class NthElementCPU {
...
+        int idx = idxs[temp];
+        h_res_idx[pos] = idx + batchIdx * batchOffset;
+        h_res[pos] = scoresData[idx];
+        ++pos;
+      }
```
snukky: Could you add a comment for this one or the outer for-loop explaining the rough idea?

PinzhenChen: Hi thanks roman. I added comments as you asked.

snukky: Thanks!

PR #606, opened 2020-02-17
Make things compile with clang-8 with fewer warnings.

ugermann:
This PR aims to resolve some of the discussion related to PR #598, specifically reducing the number of warnings when compiling with clang.

Current status:
* no warnings when compiling with GCC on Linux with 
  ```bash
  cmake -DUSE_SENTENCEPIECE=on -DUSE_CCACHE=on /path/to/source/code
  make -j
  ```
* a few warnings related to the compilation of sentencepiece when compiling with clang-8 on Linux. **clang compilation with CUDA currently doesn't work yet, but this PR is a first step in the right direction.**
  ```bash
  export CXX=clang++-8
  export CC=clang-8
  cmake -DUSE_SENTENCEPIECE=on -DUSE_CCACHE=on -DCOMPILE_CUDA=off /path/to/source/code
  make -j
  ```
(Obviously, I run these two compilations in two different build dirs)
With Clang, there are still a few warnings when compiling sentencepiece, but these are related to flags set in src/3rd_party/sentencepiece/src/CMakeLists.txt and I don't know how to 'unset' flags in emacs. Generally, I would propose setting compile flags outside of submodules rather than making changes there.

Clang warnings mostly related to the following:
* classes that have virtual member functions but no virtual destructor. Clang doesn't like that.
* unused parameter names in functions
* compiler flags that Clang doesn't know about.

Also, please take note that __GNUC__ is defined when using Clang, so `#ifdef __GNUC__` is not a valid check for using gcc.

@XapaJIaMnu: feel free to pull this branch into yours and see if this works on MacOS. I've tested this only with clang-8 on linux.


XapaJIaMnu:
Compilation without MKL is broken here:
```
[ 43%] Building CXX object src/CMakeFiles/marian.dir/tensors/cpu/tensor_operators.cpp.o
/Users/xapajiamnu/marian-dev/src/tensors/cpu/tensor_operators.cpp:226:32: error: unused parameter 'out' [-Werror,-Wunused-parameter]
void TransposeFirst3In4(Tensor out, Tensor in, const std::vector<int>& vAxis) {
                               ^
/Users/xapajiamnu/marian-dev/src/tensors/cpu/tensor_operators.cpp:226:44: error: unused parameter 'in' [-Werror,-Wunused-parameter]
void TransposeFirst3In4(Tensor out, Tensor in, const std::vector<int>& vAxis) {
```

ugermann:
@XapaJIaMnu: pull and try again. Should now also work with -DCOMPILE_EXAMPLES=on.

XapaJIaMnu:
Some tests fail to compile, now
```
/Users/xapajiamnu/marian-dev/src/tests/sqlite.cpp:10:14: error: unused
      parameter 'argc' [-Werror,-Wunused-parameter]
int main(int argc, char** argv) {
```
Try 
`-DCOMPILE_TESTS=on`

ugermann:
Fixed. Pull & retry.

XapaJIaMnu:
That works now together with my mac_compile branch.

Could you try compiling `fbgemm` (`-DUSE_FBGEMM=on`) on your clang version. I can't compile it on the mac, but that's (only partly) due `aligned_alloc`.

src/common/intrusive_ptr.h
```diff
@@ -169,12 +169,12 @@ inline bool operator!=(const IntrusivePtr<T>& a, std::nullptr_t) {
 
 template<class T>
 inline bool operator==(T* a, const IntrusivePtr<T>& b) {
-  return b.get();
+  return a == b.get(); // used to say: return b.get(); That cannot be right. [UG]
```
frankseide: Thanks for finding this!

src/data/shortlist.h
```diff
@@ -183,13 +184,12 @@ class LexicalShortlistGenerator : public ShortlistGenerator {
                             Ptr<const Vocab> srcVocab,
                             Ptr<const Vocab> trgVocab,
                             size_t srcIdx = 0,
-                            size_t trgIdx = 1,
+                            size_t /*trgIdx*/ = 1, // not used in this class
```
frankseide: (no need for the comment I think. Not important.)

src/tensors/cpu/tensor_operators.cpp
```diff
@@ -275,11 +276,8 @@ void TransposeFirst3In4(Tensor out, Tensor in, const std::vector<int>& vAxis) {
       }
     }
   }
-#else
```
frankseide: Is there a reason to remove this? It seems unrelated to clang?

src/tensors/tensor_operators.h
```diff
@@ -24,15 +24,20 @@
...
+void copy(Ptr<Backend>& backend, const InIt beg, const InIt end, OutIt it) {
   if(backend->getDeviceId().type == DeviceType::gpu)
     gpu::copy(backend, beg, end, it);
   else
-#endif
     std::copy(beg, end, it);
 }
+#else
+template <typename InIt, typename OutIt>
+void copy(Ptr<Backend>& , const InIt beg, const InIt end, OutIt it) {
```
frankseide: Couldn't you just say

backend;

in an #else branch?

src/tensors/tensor_operators.h
```diff
@@ -200,13 +206,26 @@ static inline void LayerNormalizationGrad(
...
                             float eps) {
-#ifdef CUDA_FOUND
   if(gradX->getBackend()->getDeviceId().type == DeviceType::gpu)
     gpu::LayerNormalizationGrad(allocator, gradX, gradGamma, gradBeta, adj, y, x, gamma, beta, eps);
   else
-#endif
     cpu::LayerNormalizationGrad(gradX, gradGamma, gradBeta, adj, y, x, gamma, beta, eps);
 }
+#else
+static inline void LayerNormalizationGrad(
```
frankseide: This feels like duplicating too much. What is the problem with the oruiginal? Can you solve it by saying `var;` to reference the unreferenced variables?

src/training/gradient_dropping/sparse_tensor.h
```diff
@@ -118,7 +118,11 @@ class SparseTensorBase : public std::enable_shared_from_this<SparseTensorBase> {
   }
 
   // Convert a tensor into a sparse tensor format
+#ifdef CUDA_FOUND
   void fromDense(Tensor t) {
+#else
```
frankseide: Same comment here: Just say `t;`

src/tensors/cpu/tensor_operators.cpp
ugermann: There's exactly one call to that function in the Marian code, and that's farther down in this file. That call is already guarded by `#if MKL_FOUND`. Should anyone ever decide to call the function from somewhere else without using MKL, compilation should fail, not running the program at some random point. This change is related to clang, because otherwise we get an unused variable warning.

src/data/shortlist.h
ugermann: Comment removed.

src/tensors/tensor_operators.h
ugermann: I don't like no-op code just to deal with compiler warnings, especially since I'm not sure if that's actually a no-op, or whether the compiled code will still access memory (and potentially screw up caching). However, I agree that code duplication is undesirable. The proper way (on those compilers that support it, is through `__attribute__(unused)` (or `[[maybe_unused]]`). See macro MAYBE_UNUSED in src/common/definitions.h.

src/training/gradient_dropping/sparse_tensor.h
ugermann: See my comment above. Resolve conversation when satisfied.

src/tensors/tensor_operators.h
ugermann: See my comment above. Resolve conversation when satisfied.

frankseide: I like the MAYBE_UNUSED approach, as it seems low-overhead.

src/tensors/tensor_operators.h
frankseide: We do do this at other places though, so it wouldn't be unusual in this codebase. MAYBE_UNUSED is good.

ugermann:
@XapaJIaMnu: now compiles with -DUSE_FBGEMM and clang-8 on Ubuntu 16.04.

ugermann:
```
cmake -DUSE_SENTENCEPIECE=on -DUSE_CCACHE=on -DCOMPILE_CUDA=off -DUSE_FBGEMM=on -DCOMPILE_TESTS=on -DCOMPILE_EXAMPLES=on /path/to/repository 
make -j
```
now runs without warning with clang 8,9,10 and gcc 5.4 on linux (including with `-DCOMPILE_CUDA=on` for gcc..

snukky:
I added a job on Jenkins that compiles CPU-only Marian with Clang-8 (on gna)
http://vali.inf.ed.ac.uk/jenkins/view/marian/job/marian-dev-cpu-clang-8/
with the following command: 

    CC=/usr/bin/clang-8 CXX=/usr/bin/clang++-8 cmake -DUSE_SENTENCEPIECE=ON -DCOMPILE_CPU=ON -DCOMPILE_CUDA=OFF -DCOMPILE_TESTS=ON -DCOMPILE_EXAMPLES=ON ..

ugermann:
Nicked merged this into #598, which got merged into master.

Issue #620, opened 2020-04-09
Update python scripts to python3


kdavis-mozilla:
### Bug description
Please add a clear and concise description of the bug, including observed and if possible expected behavior.

The tests under `regression-tests` use python3 as a result of the use of `pip3` in the `pip` target of `regression-tests/Makefile`. Thus the python scripts `scripts/*/*.py` should also use python3. However, not all of them do.

A patch to change this is as follows
```bash
diff --git a/scripts/checkpoints/average.py b/scripts/checkpoints/average.py
index 53bff186..da1ca252 100755
--- a/scripts/checkpoints/average.py
+++ b/scripts/checkpoints/average.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 """
 This script takes multiple Marian *.npz model files and outputs an elementwise average of the model,
 meant to do check-point averaging from: 
diff --git a/scripts/contrib/inject_ctt.py b/scripts/contrib/inject_ctt.py
index 751ee1c6..620c3152 100755
--- a/scripts/contrib/inject_ctt.py
+++ b/scripts/contrib/inject_ctt.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 from __future__ import print_function
 
diff --git a/scripts/contrib/inject_model_params.py b/scripts/contrib/inject_model_params.py
index 46096eb8..a0e637a3 100755
--- a/scripts/contrib/inject_model_params.py
+++ b/scripts/contrib/inject_model_params.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 from __future__ import print_function
 
diff --git a/scripts/embeddings/prepare_corpus.py b/scripts/embeddings/prepare_corpus.py
index 98326218..0c54be5d 100755
--- a/scripts/embeddings/prepare_corpus.py
+++ b/scripts/embeddings/prepare_corpus.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 # -*- coding: utf-8 -*-
 
 from __future__ import print_function
diff --git a/scripts/embeddings/process_word2vec.py b/scripts/embeddings/process_word2vec.py
index 4f5ba493..685f8d23 100755
--- a/scripts/embeddings/process_word2vec.py
+++ b/scripts/embeddings/process_word2vec.py
@@ -1,4 +1,4 @@
[...]

PR #621, opened 2020-04-09
python3 shebang from #620

kpu:
### Description

This changes shebangs from `python` to the more explicit `python3` according to the patch in #620.  

The regression tests use `pip3` so `python3` is consistent with that.  

https://github.com/marian-nmt/marian-regression-tests/blob/6a08849b23f6c14eefbe12f4eb73dc638b962587/Makefile#L25


snukky:
Thanks! Have you or @kdavis-mozilla tested these scripts with python3, especially scripts that are not used in regression tests? Have you checked if there are no scripts in regression tests that are mistakenly called via `python script.py` instead of just `./script.py`? If you already did it, I will not go over this again.

kdavis-mozilla:
I'll run a set of tests later today to test this.

kpu:
Also, I think the client works in python2.7 so we shouldn't force python3?

kdavis-mozilla:
Python 3 is already required from the [Makefile](https://github.com/marian-nmt/marian-regression-tests/blob/master/Makefile#L25)

snukky:
Yes, but there may be direct executions with python2 somewhere in the regression tests, and also not all scripts are used in regression tests. I will check.

kpu:
The client script is intended to be usable outside the regression tests.  And many users don't care about running the regression tests.  So would a better way to be to change the tests (which assume `pip3` as @kdavis-mozilla points out) to invoke with `python3` rather than use the shebang?

kdavis-mozilla:
I guess I'm just saying if you install the dependencies with pip3 for the tests they won't be available for python 2. So the tests have to use python 3 or they won't have the correct dependencies unless you get lucky.

kpu:
Yeah I think the tests should always call python3. Which could be accomplished by always writing `python3 client_example.py` in the regression tests.

snukky:
I think the issue is related rather to marian-regression-tests than marian-dev. I see `python xxx/model_info.py` in some regression tests, and if I remember correctly, this script generates different outputs for python2 and python3. I will take a look and unify this in the marian-dev/marian-regression-tests repo.

emjotde:
@snukky I would put it on you to decide here.

snukky:
I will clean up script executions in regression tests and then check if the rest of scripts works properly with python3. I am keeping the PR open.

snukky:
I cleaned up calls to python in regression tests making sure that python3 is used everywhere (https://github.com/marian-nmt/marian-regression-tests/pull/51). This solved the issue reported by @kdavis-mozilla being motivation for reporting #620. 

The following scripts are used in regression tests and works fine with python3:
- contrib/model_info.py
- embeddings/export_embeddings.py
- server/client_example.py

Other scripts:
- checkpoints/average.py works with python3 
- scripts in embeddings/ seem to work too
- contrib/inject_ctt.py works
- I have not tested contrib/inject_model_params.py, but it should work. This script very unlikely is still compatible with the recent version of Nematus
- contrib/fix_hard.py has no shebang, but is definitely python2

Python 2 is no longer a default Python version in recent Ubuntu distros. It seems that `python` will not be available by default in Ubuntu 20.04 LTS, so it would be nice to update shebangs in our scripts.

kpu:
> Following my last comment, I think it is good to have these updates. Could you add a note about the change in CHANGELOG.md?

Done.

PR #630, opened 2020-04-11
Fix 0 * nan behavior due to using -O3 instead of -OFast

emjotde:
### Description
Address a number of issues resulting from the change of -OFast to -O3 in nan handling. Fixes unit tests and one CPU-side operator. We switch from Approx.epsilon to Approx.margin to correctly use absolute tolerance not relative tolerance as was done before incorrectly.


src/tensors/cpu/tensor_operators.cpp
```diff
@@ -136,15 +138,14 @@ void Split1(std::vector<Tensor>& outputs, const Tensor in) {
...
-    cpu::gInsertCols(out->data(),
-                     in->data(),
-                     rows,
-                     cols_out,
-                     cols_out,
-                     cols_in,
-                     0,
-                     offset,
-                     1);
+    cpu::gInsertCols<true>(out->data(),
```
frankseide, line 141: consider adding a comment, e.g. `</*add=*/true>` for better readability (also consider adding such fields for the non-obvious 0 argument here and above)

PR #660, opened 2020-06-17
Improved handling of SIGTERM

ugermann:
### Description
- moved signal handling out of scheduler code (was already global there, but still in the files)
- pay attention to the signal in the threads that prepare batches in the background, which can take quite a while and prevented Marian from shutting down quickly after receiving SIGTERM in some circumstances

### How to test

start training, then send SIGTERM to process, it should save the current state of affairs and exit

### Checklist


src/training/training.h
```diff
@@ -77,6 +77,12 @@ class Train : public ModelTask {
     bool restored = !options_->get<bool>("no-restore-corpus")
                     && batchGenerator->restore(trainState);
 
+    // Install custom handler for SIGTERM, to allow for a graceful
+    // shutdown that saves the current state of training before exiting.
+    // This signal handler simply sets a flag that can be checked from
+    // everywhere (getSignalFLAG(SIGTERM); #include common/signal_handling.h)
+    signal(SIGTERM,setSignalFlag);
```
emjotde, line 84: nit: space after comma.

src/data/batch_generator.h
```diff
@@ -132,8 +133,14 @@ class BatchGenerator : public RNGEngine {
       if(current_ != data_->end())
         ++current_;
     }
+
+    std::deque<BatchPtr> tempBatches;
+
     size_t sets = 0;
     while(current_ != data_->end() && maxiBatch->size() < maxSize) { // loop over data
+      if (getSignalFlag(SIGTERM)) { // received SIGTERM, abandon ship ...
+        return tempBatches;
```
emjotde, line 142: If I understand that correctly that will return a set of batches that have been produced so far. Are we actually expecting that these will be processed in anyway? If not would it make more sense to maybe return an empty collection?

src/training/training.h
ugermann: Fixed.

src/data/batch_generator.h
ugermann: Since the return value is ignored after a SIGTERM, it shouldn't matter whether it's empty or not. Even though it's marginal here, returning an empty collection involves another call to the constructor of the return value. I have no strong feelings about this.

snukky:
All checks have passed, so no new compilation issues on Windows or MacOS with this PR.

emjotde:
Ah, there are running for PRs now? Neat!

ugermann:
> Looks alright to me. @frankseide can you take a look at that as well? Especially as far as the batch prefetching is concerned.

Pinging @frankseide.

src/data/batch_generator.h
```diff
@@ -242,13 +253,13 @@ class BatchGenerator : public RNGEngine {
       ABORT_IF(!futureBufferedBatches_.valid(), "Attempted to wait for futureBufferedBatches_ when none pending.\n"
           "This error often occurs when Marian tries to restore the training data iterator, but the corpus has been changed or replaced.\n"
           "If you have changed the training corpus, add --no-restore-corpus to the training command and run it again.");
+
       bufferedBatches_ = std::move(futureBufferedBatches_.get());
-      // if bg thread returns an empty swath, we hit the end of the epoch
-      if (bufferedBatches_.empty()) {
+      if (bufferedBatches_.empty() // i.e., end of Epoch
+          || getSignalFlag(SIGTERM)) { // process received SIGTERM, abandon ship ...
```
frankseide, line 259: I see the flag tested at three places. Are these mutually exclusive code paths, or does one check lead to returning dummy batches, and then later on, it is checked again to know to ignore the dummy values?

src/training/scheduler.h
```diff
@@ -154,11 +152,10 @@ class Scheduler : public TrainingObserver {
       : options_(options), state_(state) {
     ABORT_IF(state_->factor != 1, "state.factor unexpectedly not 1 at this point??");
     updateLearningRate(*state);
-    installSignalHandlers();
   }
 
   bool keepGoing() {
-    if(getSigtermFlag()) // received signal SIGERM => exit gracefully
+    if(getSignalFlag(SIGTERM)) // received signal SIGERM => exit gracefully
```
frankseide, line 158: I realize that there is too much duplication of knowledge on what it means to signal this job to terminate. E.g. you duplicate the knowledge that it is SIGTERM at four places up to here.

I suggest to abstract this. Have a function `terminationRequested()` or so, which encapsulates the mechanism. One can also imagine much less risky mechanisms, e.g. creating a file whose presence indicates a termination request. It is also not clear to a reader whether reading out the signal clears it or not.

Signals are risky in that they interfere with UNIX' idiotic EINTR protocol, which is, for example, not implemented correctly in some versions of NCCL, and likely not implemented correctly in any of our external file-reading/writing libraries (e.g. the NPZ writer quite surely does not follow it, and I wouldn't put it beyond Protobufs either).

src/training/training.h
```diff
@@ -77,6 +77,12 @@ class Train : public ModelTask {
     bool restored = !options_->get<bool>("no-restore-corpus")
                     && batchGenerator->restore(trainState);
 
+    // Install custom handler for SIGTERM, to allow for a graceful
+    // shutdown that saves the current state of training before exiting.
+    // This signal handler simply sets a flag that can be checked from
+    // everywhere (getSignalFLAG(SIGTERM); #include common/signal_handling.h)
+    signal(SIGTERM, setSignalFlag);
```
frankseide, line 84: Please abstract this together with the `terminationRequested()` function.

src/data/batch_generator.h
ugermann: It's a nested loop that you want to break out of as soon as possible, so you need to check the conditions for the inner loops and the outer loop. Keep in mind that SIGTERM may occur at any time. The default signal handler for SIGTERM kills all threads in the process immediately. We don't want that, so that we can save the current state of training before exiting. That means that we have to check for SIGTERM in strategic locations where it is convenient for us to check, preferably in the middle of expensive loops. You also have no guarantee that bufferedBatches is empty after SIGTERM -- it may occur between the assignment and and the check.

src/training/scheduler.h
ugermann: There is nothing to abstract here. We want to react to one very specific scenario: The process receives a request to save its current state and exit a.s.a.p. That's called a graceful shutdown and exactly what SIGTERM and signal handling are meant for. There are other termination requests where we do NOT want to save the state of training, e.g. after SIGINT. 

In fact, on linux, (glibc or later), if I read [this](https://man7.org/linux/man-pages/man2/signal.2.html) correctly, installing a signal handler through the gcc signal function (which wraps around sigaction) will by default restarted interruptible system calls that are interrupted by a signal. Except data transfer functions that do not return EINTR but instead the number of bytes transferred, so chances are the NPZ writer won't stumble over interrupted system calls.

src/training/training.h
ugermann: There is nothing to abstract here. We want to handle a specific case: SIGTERM, not just any termination request. For SIGINT, for example, we want the default behaviour.

src/data/batch_generator.h
ugermann: Now returning empty deque. It doesn't make a difference, because we need to check for SIGTERM later anyway, because SIGTERM can occur at any point, so we have to check in strategic locations, preferably in the middle of potentially long  loops (e.g., reading a million sentences from disk to create the next minibatch).

src/command/marian_train.cpp
```diff
@@ -47,9 +48,9 @@ int main(int argc, char** argv) {
   // for bash in http://tldp.org/LDP/abs/html/exitcodes.html. This allows parent
   // scripts to determine if training terminated naturally or via SIGTERM.
   // Whith this approach we can accommodate additional signals in the future.
-  // An alternative would be to return 124, which is what the timeout command
+  // An alternative would be to exit with code 124, which is what the timeout command
   // returns for timeout -s SIGTERM <seconds> ...., because exiting after SIGTERM
```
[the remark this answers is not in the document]
ugermann: How could it possibly be? The NCCL issue was probably caused by compiling with -pg, which installs the SIGPROF signal handler, and it probably happened during profiling, when the program is hammered with SIGPROF signals.

src/common/signal_handling.h
```diff
@@ -0,0 +1,27 @@
+#pragma once
+#include <csignal>
+
+// SIGNAL HANDLING
+
+// The Marian signal handlers set global flags that thread can
```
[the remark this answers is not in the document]
ugermann: This is a generic and general mechanism so that people who write executables based on Marian as a library can easily implement custom behavior in response to signals received. The explanation of what it is "actually used for" is immediately below in the rest of the comment.

src/command/marian_train.cpp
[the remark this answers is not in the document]
ugermann: Besides, NCCL issue No. 137 was filed on 3 October 2018. The original pull request for graceful shutdown was merged on 27 October 2019. So I doubt that SIGTERM handling caused the NCCL issue.

src/data/batch_generator.h
```diff
@@ -149,16 +156,20 @@ class BatchGenerator : public RNGEngine {
...
     // @TODO: we could just use a vector and do a sort() here; would make the cost more explicit
     const size_t mbWords = options_->get<size_t>("mini-batch-words", 0);
     const bool useDynamicBatching = options_->has("mini-batch-fit");
     BatchStats::const_iterator cachedStatsIter;
     if (stats_)
       cachedStatsIter = stats_->begin();
+
     while(!maxiBatch->empty()) { // while there are sentences in the queue
+
+      if (getSignalFlag(SIGTERM)) { // received SIGTERM, abandon ship ...
```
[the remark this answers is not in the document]
ugermann: Fixed.

ugermann:
replaced by #703

emjotde:
Reopened to continue review here rather than in #703

src/common/config_parser.cpp
```diff
@@ -324,13 +324,23 @@ void ConfigParser::addOptionsModel(cli::CLIWrapper& cli) {
         "Dropout for transformer attention (0 = no dropout)");
     cli.add<float>("--transformer-dropout-ffn",
         "Dropout for transformer filter (0 = no dropout)");
+
```
emjotde, line 327: Superfluous empty line.

src/common/config_parser.cpp
```diff
@@ -324,13 +324,23 @@ void ConfigParser::addOptionsModel(cli::CLIWrapper& cli) {
...
 void ConfigParser::addOptionsTraining(cli::CLIWrapper& cli) {
-  auto previous_group = cli.switchGroup("Training options");
+  auto previous_group = cli.switchGroup("Signal Handling");
+  // --sigterm is deliberately not a boolean, to allow for a consistent
+  // pattern of specifying custom signal handling in the future.
+  // (e.g., dump model but continue training upon SIGUSR1, or report current
+  // training status upon SIGINFO.)
+  cli.add<std::string>("--sigterm",
+                       "What to do with SIGTERM: 'graceful' => save and exit (default); "
+                       "'immediate' => exit immediately.", "graceful");
```
emjotde, line 341: Please put the default value in the next line. We might not always do that, but let's aim for that.

src/common/config_parser.cpp
```diff
@@ -324,13 +324,23 @@ void ConfigParser::addOptionsModel(cli::CLIWrapper& cli) {
...
         "Dropout for transformer filter (0 = no dropout)");
+
   }
   cli.switchGroup(previous_group);
   // clang-format on
 }
 
 void ConfigParser::addOptionsTraining(cli::CLIWrapper& cli) {
-  auto previous_group = cli.switchGroup("Training options");
+  auto previous_group = cli.switchGroup("Signal Handling");
```
emjotde, line 334: @snukky can you comment if this is the right place to put the handler or if you would like it be located somewhere else? @ugermann is signal handling currently only specific to training or are the other things like translation and scoring reacting as well?

src/common/signal_handling.cpp
```diff
@@ -7,15 +7,27 @@
...
   // sig_atomic_t has 32 bits. We don't accommodate signals beyond that.
-  ABORT_IF(sig >= 32, "Signal {} out of range (must be < 32).", sig);
+  ABORT_IF(sig >= 32, "Signal out of range (must be < 32, is {}).", sig);
   return sigflags_ & (1<<sig);
 }
 
+void requestGracefulExit(int sig) {
+  setSignalFlag(sig);         // keep track of triggering signal
+  gracefulExitRequested_ = 1; // set flag to exit gracefully
+  LOG(debug, "Graceful exit requested via signal {}.", sig);
```
emjotde, line 21: let's go for `info` instead of `debug`. I would say actually forcing an exit should always be logged.

src/data/batch_generator.h
```diff
@@ -252,18 +246,26 @@ class BatchGenerator : public RNGEngine {
...
-      // We only get here if the future has been scheduled to run; it must be valid.
-      ABORT_IF(!futureBufferedBatches_.valid(), "Attempted to wait for futureBufferedBatches_ when none pending.\n"
-          "This error often occurs when Marian tries to restore the training data iterator, but the corpus has been changed or replaced.\n"
-          "If you have changed the training corpus, add --no-restore-corpus to the training command and run it again.");
-
-      bufferedBatches_ = std::move(futureBufferedBatches_.get());
-      if (bufferedBatches_.empty() // i.e., end of Epoch
-          || getSignalFlag(SIGTERM)) { // process received SIGTERM, abandon ship ...
-        return nullptr;
+      if(runAsync_) { // by default we will run in asynchronous mode
```
emjotde: Comment to self and @frankseide: this is git being bad at doing a three-way diff. The `runAsync_` part is from the `master` branch.

src/data/batch_generator.h
```diff
@@ -252,18 +246,26 @@ class BatchGenerator : public RNGEngine {
...
+        // out of data: need to get next batch from background thread
+        // We only get here if the future has been scheduled to run; it must be valid.
+        ABORT_IF(!futureBufferedBatches_.valid(), "Attempted to wait for futureBufferedBatches_ when none pending.\n"
+            "This error often occurs when Marian tries to restore the training data iterator, but the corpus has been changed or replaced.\n"
+            "If you have changed the training corpus, add --no-restore-corpus to the training command and run it again.");
+        bufferedBatches_ = std::move(futureBufferedBatches_.get());
+        // if bg thread returns an empty swath, we hit the end of the epoch
+        if (bufferedBatches_.empty() || gracefulExitRequested()) {
```
emjotde, line 257: only this call to `gracefulExitRequested()` is new.

src/data/batch_generator.h
```diff
@@ -252,18 +246,26 @@ class BatchGenerator : public RNGEngine {
...
+          return nullptr;
+        }
+        // and kick off the next bg operation
+        fetchBatchesAsync();
+      } else { // don't spawn any threads, i.e. batch fetching is blocking.
+        bufferedBatches_ = fetchBatches();
+        // if bufferedBatches is empty we hit the end of the epoch
+        if (bufferedBatches_.empty() || gracefulExitRequested()) {
```
emjotde, line 265: And this call to `gracefulExitRequested`

src/data/batch_generator.h
emjotde: @ugermann I see what you meant with the 3-way diff.

emjotde:
@frankseide this looks good to me now and ready for another look from you.

src/common/config_parser.cpp
ugermann: Currently it's only specific to training. The other option where it might be useful is a translation server, where we want to might shut down a server for load balancing / migration to a different node but want to complete pending requests before the particular server instance goes offline.

emjotde: Right, it seems to be that signal handling is a little bit similar to things like `--log-level` or even `--help`. Sort of a meta-option.  Let's @snukky have the last word here.

src/common/signal_handling.cpp
ugermann: Technically, the graceful shutdown is being logged anyway (see Scheduler::finished()). I have no strong opinions about this, though. Also technically, the only thing that should happen in a signal handler is setting flags. as it's entirely unpredictable when a signal is handled. I don't know if the logger involves a lock, (and whether that lock has a timeout), but theoretically: if the signal handler is triggered while the logger is logging something else in the same thread, we might end up with a deadlock. Happy to take the logging out here (or comment it out should we ever need to debug again). Or we might guard this to be included only in the Debug build type but not in Release.

src/common/config_parser.cpp
ugermann: Fixed.

src/common/config_parser.cpp
ugermann: Fixed.

src/common/config_parser.cpp
snukky: If potentially the option can be used outside the training too, I would add it under 'General options'; only for the trainer. Creating a separate section for a single option will also make the help even more longer.

src/common/signal_handling.cpp
emjotde: I think the logging is thread safe.

src/common/config_parser.cpp
```diff
@@ -330,7 +330,17 @@ void ConfigParser::addOptionsModel(cli::CLIWrapper& cli) {
...
+  // --sigterm is deliberately not a boolean, to allow for a consistent
+  // pattern of specifying custom signal handling in the future.
+  // (e.g., dump model but continue training upon SIGUSR1, or report current
+  // training status upon SIGINFO.)
+  cli.add<std::string>("--sigterm",
+                       "What to do with SIGTERM: 'graceful' => save and exit (default); "
+                       "'immediate' => exit immediately.", 
+                       "graceful");
```
snukky, line 341: Everywhere else we start help messages in a new line with four whitespaces at the beginning. Could we follow the same pattern here? Maybe you will save one line.

src/common/signal_handling.cpp
kpu: I'm with @ugermann here, do as little as possible in a signal handler.  You're limited to these functions https://man7.org/linux/man-pages/man7/signal-safety.7.html and would need to show that the logger only uses them as well as being reentrant in the same thread, not just thread safe.

src/common/config_parser.cpp
ugermann: Now in 'General options' as suggested, without separate section.

src/common/signal_handling.cpp
emjotde: OK, if we just remove it then there is no question about the logger to use :)

ugermann: @emjotde: signal handling INTERRUPTS one thread (kernel decides which one among those that don't explicitly block the signal) to execute the signal handling in that thread, so even threat safety might not prevent a deadlock. The timing would have to be very unfortunate, but it is conceivable. Better not to mess with it. Logging removed from signal handler.

src/common/config_parser.cpp
ugermann: Formatting fixed.

src/common/config_parser.cpp
```diff
@@ -143,6 +143,15 @@ void ConfigParser::addOptionsGeneral(cli::CLIWrapper& cli) {
   cli.add<std::string>("--dump-config",
     "Dump current (modified) configuration to stdout and exit. Possible values: full, minimal, expand")
     ->implicit_val("full");
+  if(mode_ == cli::mode::training) {
+    // --sigterm is deliberately not a boolean, to allow for a consistent
+    // pattern of specifying custom signal handling in the future.
+    // (e.g., dump model but continue training upon SIGUSR1, or report current
+    // training status upon SIGINFO.)
+    cli.add<std::string>("--sigterm",
+      "What to do with SIGTERM: 'graceful' => save and exit (default); 'immediate' => exit immediately.",
```
frankseide, line 152: "graceful" is not descriptive. Please say in the option string what it does, e.g. "save-and-exit"

"immediate" could be "terminate"

src/common/signal_handling.cpp
```diff
@@ -0,0 +1,32 @@
...
+// https://stackoverflow.com/questions/231912/what-is-the-difference-between-sigaction-and-signal).
+
+namespace marian{
+volatile std::sig_atomic_t sigflags_{0};
+volatile std::sig_atomic_t gracefulExitRequested_{0};
+
+bool getSignalFlag(const int sig) {
+  // sig_atomic_t has 32 bits. We don't accommodate signals beyond that.
+  ABORT_IF(sig >= 32, "Signal out of range (must be < 32, is {}).", sig);
+  return sigflags_ & (1<<sig);
```
frankseide, line 15: `&` is an integer operation, but you return `bool`.

Can you please also explain what this function does? It seems it does two things: It converts the signal, but then also masks it with a class member. I cannot understand the meaning of the function without a comment.

src/common/signal_handling.cpp
```diff
@@ -0,0 +1,32 @@
...
+// using sigaction, which apparently is not available for Windows (cf.
+// https://stackoverflow.com/questions/231912/what-is-the-difference-between-sigaction-and-signal).
+
+namespace marian{
+volatile std::sig_atomic_t sigflags_{0};
+volatile std::sig_atomic_t gracefulExitRequested_{0};
+
+bool getSignalFlag(const int sig) {
+  // sig_atomic_t has 32 bits. We don't accommodate signals beyond that.
+  ABORT_IF(sig >= 32, "Signal out of range (must be < 32, is {}).", sig);
```
frankseide, line 14: Do we really need this check? I suggest to remove both the comment and the ABORT, since `sig_atomic_t` is, by its name, clearly meant for signals, and therefore defines the valid signal ranges. Marian is not an operating system, we don't need to hedge against every possible invalid argument. You can also remove the check and comment in setSignalFlag

src/common/signal_handling.cpp
```diff
@@ -0,0 +1,32 @@
...
+  return sigflags_ & (1<<sig);
+}
+
+void requestGracefulExit(int sig) {
+  setSignalFlag(sig);         // keep track of triggering signal
+  gracefulExitRequested_ = 1; // set flag to exit gracefully
```
frankseide, line 20: gracefulExitRequested() returns a bool, but here you set it to an integer. Please choose one (`bool` if possible)

src/common/signal_handling.h
```diff
@@ -0,0 +1,39 @@
...
+
+// SIGNAL HANDLING
+
+// The signal handlers (and checkers) here are implemented in line with with the recommendations
+// for signal handling in the SEI CERT C Coding Standard, specifically
+//
+// - SIG30-C:
+//   https://wiki.sei.cmu.edu/confluence/display/c/SIG30-C.+Call+only+asynchronous-safe+functions+within+signal+handlers
+//
+// - SIG31-C:
```
frankseide, line 13: Does this mean that sig=31 is a valid value? If so, please use unsigned types in the bit masks, e.g. 1U << sig

src/common/signal_handling.h
```diff
@@ -0,0 +1,39 @@
...
+
+/// Request graceful exit (signal handler)
+void requestGracefulExit(const int sig);
+
+/// Check if graceful exit was requested.
+bool gracefulExitRequested();
+
+/// General purpose signal handler that simply sets a flag when a signal is received.
+//  (only for SIGNAL No. < 32).
+void setSignalFlag(const int sig);  // custom handler (set flag) for sig
```
frankseide, line 34: I suggest to not use `const` in these function signatures. It exposes an internal behavior of the function to the caller. And the functions are so simple that adding `const` does not make them easier to udnerstand or verify.

src/training/training.h
```diff
@@ -16,6 +16,7 @@ template <class ModelWrapper>
 class Train : public ModelTask {
 private:
   Ptr<Options> options_;
+  void installCustomSignalHandlers();
```
frankseide, line 19: Is it installing multiple signal handlers?

emjotde:
@ugermann what was the default behavior before the PR until now; would it save the checkpoint on Ctrl+C? I cannot remember ever seeing that happen actually.

src/training/training.h
ugermann: Potentially in the  future. I prefer to keep API changes rare.

src/common/config_parser.cpp
ugermann: Strings changed, but SIGTERM is widely considered a request for a graceful shutdown.
See here: https://qph.fs.quoracdn.net/main-qimg-1180ef2465c309928b02481f02580c6a
and here: https://www.booleanworld.com/kill-process-linux/#:~:text=By%20default%2C%20all%20the%20process,gracefully%22.
and here: https://github.com/python-trio/snekomatic/issues/6

frankseide: In all of these examples, graceful shutdown means to leave the world in a well-defined state (e.g. finish in-flight requests to a service). The examples do not suggest saving state, and note that the PR saves state in a way that is inconsistent with configured behavior, because the configuration says to save every `save-freq` updates, and not at a time of receiving SIGTERM, which is a random update.

In my mind, you don't want a "graceful shutdown" as given in these examples, but more like an "early exit" from the training process.

Alternatively, you may be wanting "preemption," i.e. save state that allows to restart the training. Your PR accomplishes that as well (right?), and such feature would be useful in our internal server farm as well when running jobs in low-priority preemptable mode.

src/common/signal_handling.cpp
ugermann: sig_atomic_t is meant for signals in the sense that certain operations on it are guaranteed to be atomic even during signal handling. Essentially it is an int type that I'm using here as a bit field, so that we have a generic signal handler that we can install for arbitrary signals (lower than 30). That check is a check against programming errors. Signals can be higher than 30, but this code does not accommodate installing setSignalFlag for such signals.

src/common/signal_handling.cpp
ugermann: This relies on implicit type conversion from int types to bool. sig_atomic_t is specifically meant to be used in signal handlers. This is run-of-the-mill signal handling, see "Compliant Solution (Writing volatile sig_atomic_t)" here: https://wiki.sei.cmu.edu/confluence/display/c/SIG31-C.+Do+not+access+shared+objects+in+signal+handlers.

src/common/signal_handling.cpp
ugermann: Explanation added in code.

src/common/signal_handling.h
ugermann: 1. No. SIG31-C is just the number of a rule 
2. Signals go up to 64, but 32 and above (with 32 and 33 actually reserved for thread management) are part of the POSIX real time extensions and not of interest to us. Neither is 30 (system lost power) or 31 (wrong argument provided to system call). Since sig_atomic_t  is a signed it, I've limited the range to 1-30. No. 29 (SIGIO) might be of interest at some point (e.g. in a server that uses sockets).

src/common/signal_handling.h
ugermann: Const removed.

src/common/config_parser.cpp
ugermann: Saving state to resume later is what Microsoft incidentally calls 'graceful' preemption, as opposed to 'immediate' preemption: https://docs.microsoft.com/en-us/powershell/high-performance-computing/understanding-policy-configuration?view=hpc19-ps#:~:text=Immediate%20preemption%20(Default)%3A%20Take,allocated%20to%20another%20job%20immediately.&text=Graceful%20preemption%3A%20Take%20resources%20from,that%20work%20is%20not%20lost.

ugermann:
@frankseide I think I have now addressed all your concerns.

CHANGELOG.md
```diff
@@ -39,6 +39,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Properly record cmake variables in the cmake build directory instead of the source tree.
 - Added default "none" for option shuffle in BatchGenerator, so that it works in executables where shuffle is not an option.
 - Added a few missing header files in shortlist.h and beam_search.h.
+- Improved handling for graceful shutdown upon receiving SIGTERM. SIGTERM now also interrupts batch prefetching, which runs in a separate thread. Graceful shutdown can be disabled with --sigterm 'immediate'.
```
frankseide, line 42: Please say what graceful shutdown means here.

src/common/signal_handling.cpp
```diff
@@ -0,0 +1,62 @@
...
+  return (sigflags_ & (1<<sig)) != 0;
+}
+
+void requestGracefulExit(int sig) {
+  setSignalFlag(sig);         // keep track of triggering signal
+  gracefulExitRequested_ = 1; // set flag to exit gracefully
+}
+
+bool gracefulExitRequested() {
+  return gracefulExitRequested_ == 1;
```
frankseide, line 59: I like this better. Thanks!

(You can close this comment once read)

src/training/scheduler.h
```diff
@@ -192,13 +189,12 @@ class Scheduler : public TrainingObserver {
 
   void started() { LOG(info, "Training started"); }
   void finished() {
-    if (getSigtermFlag())
-      LOG(info, "Training interrupted (SIGTERM).");
+    if (gracefulExitRequested())
```
frankseide, line 192: I don't see the name change. `earlyExitRequested`?

src/training/training.h
frankseide: Please rename to `installEarlyExitSignalHandler()`

ugermann:
@frankseide renamed 'graceful exit' to 'save and exit'.

src/training/training.h
ugermann: I strongly disagree on this one. installCustomSignalHandlers() is like parseOptions() a generic concept: install whatever custom signal handlers may be specified in the options. Either my pull request goes in as-is now, or I'll withdraw it. @emjotde

CHANGELOG.md
ugermann: This has been updated to reflect the current state of affairs.

src/training/scheduler.h
ugermann: Changed to saveAndExit().

emjotde:
@frankseide can take another look if you concerns are satisfied? If yes, please approve, I will merge then.

src/training/scheduler.h
```diff
@@ -223,9 +219,10 @@ class Scheduler : public TrainingObserver {
 
   void validate(const std::vector<Ptr<ExpressionGraph>>& graphs,
                 bool isFinal = false) {
-    // Do not validate if already validated (for instance, after the model is
-    // loaded) or if validation is scheduled for another update, or when signal SIGTERM was received
-    if(getSigtermFlag() // SIGTERM was received
+    // Do not validate if already validated (for instance, after the model is loaded)
+    // or if validation is scheduled for another update, or when a graceful shutdown
+    // was requested.
+    if(saveAndExit()
```
frankseide, line 225: Please rename to `saveAndExitRequested()`

emjotde:
I will rename this quickly and pull it in.

emjotde:
@snukky I love the new continuous integration :)

emjotde:
Aren't the checks a necessary condition before merging? Now it allows me to merge despite the checks not having finished yet after my update? Should that not be a blocker?

emjotde:
@ugermann merged, thanks.

Issue #668, opened 2020-07-08
set_property can not be used on an ALIAS target


kdavis-mozilla:
### Bug description
Compiling on macOS fails on lines 71 and 73 of src/3rd_party/CMakeLists.txt due to set_property in  being called on an ALIAS target.

### How to reproduce
``` bash
git clone https://github.com/marian-nmt/marian-dev
cd marian-dev
mkdir build
cmake -DCMAKE_BUILD_TYPE=Release -DUSE_STATIC_LIBS=on -DUSE_SENTENCEPIECE=on -DCOMPILE_CUDA=off ..
```

### Context
* Marian version: Commit 5b93079 of marian-dev
* CMake command: cmake command is above and I don't think think the --build-info all info is required.
* Log file: NA

Add any other information about the problem here.

snukky:
This is not a commit from the master branch. Does the issue occur in master too? If so, and you have a solution, please feel free to open a PR to master.

kdavis-mozilla:
It's in master too [CMakeLists.txt#L71](https://github.com/marian-nmt/marian-dev/blob/master/src/3rd_party/CMakeLists.txt#L71) and [CMakeLists.txt#L73](https://github.com/marian-nmt/marian-dev/blob/master/src/3rd_party/CMakeLists.txt#L73)

I'm currently testing a PR.

snukky:
Closing, PR was merged into master. Thanks @kdavis-mozilla

PR #669, opened 2020-07-08
rm aliases, added corresponding targets, and removed duplicate

kdavis-mozilla:
### Description
rm aliases, added corresponding targets, and removed duplicate

This PR fixes a bug It is related to issue #668

List of changes:
- Removed aliases in foreach
- Added corresponding targets
- Removed duplicate, and alias and its target were in the foreach

Added dependencies: none

### How to test
Run cmake on macOS

I tested this by running cmake on macOS. Previous to this commit it failed as a result of this commit it did not fail

### Checklist

- [X] I have tested _running cmake manually on macOS_

### DIFF
--- a/src/3rd_party/CMakeLists.txt
+++ b/src/3rd_party/CMakeLists.txt
@@ -66,7 +66,7 @@ if(USE_SENTENCEPIECE)
     PROPERTIES RUNTIME_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}")
 
   if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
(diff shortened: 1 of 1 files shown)


snukky:
Commit diff shows fewer changes that it is suggested in the description. Is it a mistake or an issue with GitHub?

kdavis-mozilla:
No it doesn't. Here are the details...

- The ALIAS `sentencepiece` was removed
- The TARGET `sentencepiece-static` corresponding to the ALIAS `sentencepiece` was added
- The ALIAS `sentencepiece_train` corresponding to the TARGET `sentencepiece_train-static` was removed as the TARGET `sentencepiece_train-static` was already in the foreach and thus the ALIAS `sentencepiece_train` was a duplicate

PR #676, opened 2020-07-14
Add initial GitHub workflows

snukky:
### Description
This PR adds initial CI on Ubuntu/Windows/MacOS via GitHub Actions and fixes CMake-based compilation on Windows. 

At the moment, only GitHub-hosted runners (virtual machines) are used.  Jobs will trigger with every push and pull request to master. Jobs compile a CPU-only version of Marian with SentencePiece and FBGEMM, and run unit tests. On Windows and MacOS, the compilation does not include MKL yet. It will be added later.

List of changes:
- Adding workflows running on Ubuntu, Windows and MacOS that build and test CPU-only versions of Marian
- Fixing CMake-based compilation on Windows (with Ninja):
  - Escape strings generated for build-info option
  - C4702 is disabled globally in CMake as it is disabled in the VS project file
  - Disabling warnings 3rd party zlib properly
  - Compile unit tests on Windows
  - Disabling building cpuinfo-* tools on Windows due to linker errors

Please note that these changes should not affect Windows compilation using the VS solution file.

Added dependencies: none

Future work:
- Running CPU regression tests in Ubuntu workflow
- Installing MKL in Windows/MacOS workflows
- Compiling and testing with CUDA on self-hosted runners

### How to test
See https://github.com/marian-nmt/marian-dev/actions 

### Checklist


CMakeLists.txt
```diff
@@ -65,7 +65,7 @@ if(MSVC)
   # Disabled bogus warnings for CPU intrinsics:
   # C4310: cast truncates constant value
   # C4324: 'marian::cpu::int16::`anonymous-namespace'::ScatterPut': structure was padded due to alignment specifier
-  set(DISABLE_GLOBALLY "/wd\"4310\" /wd\"4324\"")
+  set(DISABLE_GLOBALLY "/wd\"4310\" /wd\"4324\" /wd\"4702\"")
```
emjotde, line 68: Please add a comment what kind of warning this is.

src/3rd_party/phf/phf.cc
```diff
@@ -679,7 +679,7 @@ PHF_PUBLIC void PHF::compact(struct phf *phf) {
 	}
 
 	/* simply keep old array if realloc fails */
-	if ((tmp = realloc(phf->g, phf->r * size)))
+	if ((tmp = realloc(phf->g, phf->r * size)) != 0)
```
emjotde, line 682: 3rd party code, but maybe let's turn this into two lines. I am allergic to assignments inside if-conditions.

src/CMakeLists.txt
```diff
@@ -220,7 +220,12 @@ endif(USE_STATIC_LIBS)
 if(COMPILE_SERVER)
   add_executable(marian_server command/marian_server.cpp)
   set_target_properties(marian_server PROPERTIES OUTPUT_NAME marian-server)
-  target_compile_options(marian_server PUBLIC ${ALL_WARNINGS})
+  if(MSVC)
+    target_compile_options(marian_server PUBLIC ${ALL_WARNINGS})
+  else(MSVC)
+      # -Wno-suggest-override disables warnings from Boost 1.69+
```
emjotde, line 226: Maybe add a short sentence that this only disables the warning for marian-server, will warn everywhere else (I hope?)

src/tests/units/rnn_tests.cpp
```diff
@@ -181,7 +181,7 @@ void tests(DeviceType type, Type floatType = Type::float32) {
 
       auto context = concatenate({rnnFw.construct(graph)->transduce(input, mask),
                                   rnnBw.construct(graph)->transduce(input, mask)},
-                                  /*axis =*/ input->shape().size() - 1);
+                                  /*axis =*/ int(input->shape().size() - 1));
```
emjotde, line 184: Nitpicking: We usually use explicit conversion `(int)value` instead of the constructor `int(value)`.

src/tests/units/operator_tests.cpp
```diff
@@ -399,8 +410,8 @@ void tests(DeviceType device, Type floatType = Type::float32) {
       std::vector<float> SV;    // create CSR version of S
       std::vector<IndexType> SI, SO;
       SO.push_back((IndexType)SI.size());
-      for (IndexType i = 0; i < S->shape()[0]; i++) {
-        for (IndexType j = 0; j < S->shape()[1]; j++) {
+      for (IndexType i = 0; size_t(i) < S->shape()[0]; i++) {
```
emjotde, line 413: Shape has `int` value type currently, so would be better to have the comparison look like this: `i < (IndexType)S->shape()[0]`

src/tests/units/operator_tests.cpp
```diff
@@ -399,8 +410,8 @@ void tests(DeviceType device, Type floatType = Type::float32) {
...
+        for (IndexType j = 0; size_t(j) < S->shape()[1]; j++) {
```
emjotde, line 414: Same here.

CMakeLists.txt
snukky: Comment added. I tried disabling this only in the Debug build, but bogus warnings show up for the Release build too so I keep it disabled globally (as it is in the VS project file).

src/3rd_party/phf/phf.cc
snukky: Changed.

src/CMakeLists.txt
snukky: Done. Yes, it keeps treating other warnings as errors.

src/tests/units/operator_tests.cpp
snukky: Fixed.

src/tests/units/operator_tests.cpp
snukky: Fixed.

src/tests/units/rnn_tests.cpp
snukky: Changed.

PR #688, opened 2020-07-29
Minor fixes

ugermann:
### Description
Please add a clear and concise description of the changes.

- adds default "none" when retrieving "shuffle" in Batch generator, because that fails when the executable doesn't include the shuffle option.
- another minor fix in determining the MARIAN_GIT_DIRECTORY in src/CMakeList.txt
- adds a few missing header files in shortlist.h and batch_generator.h 
- adds an optional argument in n-best list generation that allows the caller to request that empty hypotheses are skipped
- removed trailing whitespace in a few files

### How to test

```
cmake -DUSE_SENTENCEPIECE=on -DUSE_CCACHE=on /path/to/source/dir
```
on Ubuntu 16.04

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Training and scoring from STDIN
 - Support for reading from TSV files from STDIN and other sources during training
   and translation with options --tsv and --tsv-fields n.
-- interal optional parameter in n-best list generation that skips empty hypotheses
+- internal optional parameter in n-best list generation that skips empty hypotheses
 
 ### Fixed
 - Fix compilation without BLAS installed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   absolute and not relative. We assumed incorrectly that epsilon is absolute tolerance.
 - Fixed bug in finding .git/logs/HEAD when Marian is a submodule in another project.
 - Properly record cmake variables in the cmake build directory instead of the source tree.
+- Added default "none" for option shuffle in BatchGenerator, so that it works in executables where shuffle is not an option.
 
 ### Changed
 - Move Simple-WebSocket-Server to submodule
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Training and scoring from STDIN
 - Support for reading from TSV files from STDIN and other sources during training
   and translation with options --tsv and --tsv-fields n.
+- interal optional parameter in n-best list generation that skips empty hypotheses
 
 ### Fixed
 - Fix compilation without BLAS installed
@@ -35,6 +36,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed bug in finding .git/logs/HEAD when Marian is a submodule in another project.
 - Properly record cmake variables in the cmake build directory instead of the source tree.
 - Added default "none" for option shuffle in BatchGenerator, so that it works in executables where shuffle is not an option.
+- Added a few missing header files in shortlist.h and beam_search.h.
 
 ### Changed
 - Move Simple-WebSocket-Server to submodule
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Training and scoring from STDIN
 - Support for reading from TSV files from STDIN and other sources during training
   and translation with options --tsv and --tsv-fields n.
-- internal optional parameter in n-best list generation that skips empty hypotheses
+- Internal optional parameter in n-best list generation that skips empty hypotheses.
 
 ### Fixed
 - Fix compilation without BLAS installed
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Training and scoring from STDIN
 - Support for reading from TSV files from STDIN and other sources during training
   and translation with options --tsv and --tsv-fields n.
+- internal optional parameter in n-best list generation that skips empty hypotheses
```
snukky, line 22: "Internal" ;)

src/CMakeLists.txt
```diff
@@ -129,7 +129,9 @@ set(MARIAN_GIT_DIR ${CMAKE_CURRENT_SOURCE_DIR}/../.git)
 if(NOT IS_DIRECTORY ${MARIAN_GIT_DIR}) # i.e., it's a submodule
   file(READ ${MARIAN_GIT_DIR} MARIAN_GIT_DIR)
   string(REGEX REPLACE "gitdir: (.*)\n" "\\1" MARIAN_GIT_DIR ${MARIAN_GIT_DIR})
-  get_filename_component(MARIAN_GIT_DIR "${CMAKE_CURRENT_SOURCE_DIR}/../${MARIAN_GIT_DIR}" ABSOLUTE)
+  if(NOT IS_ABSOLUTE ${MARIAN_GIT_DIR})
+    get_filename_component(MARIAN_GIT_DIR "${CMAKE_CURRENT_SOURCE_DIR}/../${MARIAN_GIT_DIR}" ABSOLUTE)
```
snukky, line 133: A minor question: why are we making sure that it is an absolute path? Is it required to work in the upper module or safer? Add a comment if you think it is worth it.

src/data/batch_generator.h
```diff
@@ -261,7 +261,7 @@ class BatchGenerator : public RNGEngine {
                  Ptr<Options> options,
                  Ptr<BatchStats> stats = nullptr)
       : data_(data), options_(options), stats_(stats), threadPool_(1) {
-    auto shuffle = options_->get<std::string>("shuffle");
+    auto shuffle = options_->get<std::string>("shuffle","none");
```
snukky, line 264: Missing space after comma.

src/data/shortlist.h
```diff
@@ -3,9 +3,12 @@
 #include "common/config.h"
 #include "common/definitions.h"
 #include "common/file_stream.h"
+#include "data/types.h"
+#include "data/corpus_base.h"
```
snukky, line 7: Almost sorted alphabetically ;)

src/translator/history.h
```diff
@@ -37,7 +37,7 @@ class History {
 
   size_t size() const { return history_.size(); } // number of time steps
 
-  NBestList nBest(size_t n) const {
+  NBestList nBest(size_t n, bool skip_empty = false) const {
```
snukky, line 40: Can `skip_empty = true` produce an n-best list with fewer than n elements and there is no longer a guarantee that all n-best lists are of the same size? If so, this requires a comment or even a warning. Could you add a docstring to this method?

emjotde:
Hi, 
Empty hypotheses as in consisting only of EOS?

src/CMakeLists.txt
ugermann: Notice the NOT. It's a bit complicated, but for submodules, git keeps the git information in the root directory and stores the path to the relevant directory in a file named .git in the submodule. The path can be either relative or absolute, as I found out the hard way. I can't remember all the details, but it may have also involved running things in host-mounted directories in compiling in Docker containers for dockerization. What I do remember is that at some point, things didn't work and that fixed it. For all practical purposes, is has no effect on run-of-the-mill marian compilation, because then .git is always a directory.

CHANGELOG.md
ugermann: Fixed.

src/data/batch_generator.h
ugermann: Fixed.

src/translator/history.h
ugermann: Yes, if n equals the beam size. I've added a docstring with a note to that effect. I think a warning would be overkill, as skipEmpty is not the default behaviour. The typical use case for this new option is calling nBest(1, true)[0] with a beam size > 1.

ugermann:
> Hi,
> Empty hypotheses as in consisting only of EOS?

Yes. See https://arxiv.org/abs/1908.10090 and my own experience.

src/CMakeLists.txt
snukky: Thanks!

src/translator/history.h
snukky: Yes, sorry, I meant a warning in the comment, your note serves the purpose.

emjotde:
Where is the skipEmpty thing actually used? I don't see it anywhere in the current code.

ugermann:
> @ugermann Where is the skipEmpty thing actually used? I don't see it anywhere in the current code.

In the REST server, which currently lives in a different repo to have its own issue tracker etc.
https://github.com/browsermt/mts/blob/master/src/service/common/translation_job.cpp

emjotde:
I am wondering if that shouldn't at least broadcast a message to the `LOG(debug, ...)` channel.

emjotde:
A request for the future, if you happen to do the merge yourself, do a squash merge, please,

ugermann:
Is there a way to do that from the web interface? And how can I do a squash merge into master if I don't have write permission to master? New branch and new pull request?

snukky:
Yes, once a pull request has at least one approval, a button "squash and merge" should become available (at the bottom of the page) for the author or maybe even for all people who have committed to the PR. From the command line, I think it is possible to squash commits in a local branch with PR, push to the remote branch with PR, then merge into master.

PR #689, opened 2020-07-30
Optimize CPU LayerNormalization 6x with -ffast-math and lifting branches

kpu:
### Description

LayerNormalization on CPU was taking 9% of the time in an enes student, so @XapaJIaMnu set out to optimize it.  

The kernel performs a lot of sums that, pedantically, have to be done serially due to the C++ standard.  Since Marian does not compile with OpenMP by default, the `#pragma omp simd` is also ignored.  Enabling `-ffast-math` allows the compiler to assume associative math (pedantically this could have been `-fassociative-math` in gcc and something else in clang that's hard to use).  This is only enabled at the function level.  But that wasn't enough because the compiler was having trouble vectorizing all the branches for strides.  

Measured based on `intgemm_reintegrated_computestats` branch 1de262e28 with enes.student.tiny11`.  Time is as reported by gprof for condition cols = 256, alphaStride = 1, betaStride = 1 on var.

| Time(s) | % | Condition |
|--|--|--|
| 3.49 | 9.08% | Baseline |
| 1.74 | 4.81% | `-ffast-math` |
| 0.57 | 1.65% | + lift conditionals to templates |

Added dependencies: none

### How to test
```
cmake .. -DCOMPILE_CUDA=OFF -DUSE_FBGEMM=ON -DUSE_SENTENCEPIECE=ON -DCOMPILE_TESTS=ON
```
Ran unit tests, ran student model, timing above

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 
 ### Added
+- Optimize LayerNormalization on CPU by 6x
 - Decoding multi-source models in marian-server with --tsv
 - GitHub workflows on Ubuntu, Windows, and MacOS
 - LSH indexing to replace short list
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 
 ### Added
-- Optimize LayerNormalization on CPU by 6x
+- Optimize LayerNormalization on CPU by 6x through vectorization (ffast-math) and fixing performance regression introduced with strides in 77a420
 - Decoding multi-source models in marian-server with --tsv
 - GitHub workflows on Ubuntu, Windows, and MacOS
 - LSH indexing to replace short list
--- a/src/common/definitions.h
+++ b/src/common/definitions.h
@@ -22,6 +22,26 @@
 #define DONT_OPTIMIZE // silently ignore on Visual Studio, where this is less of a problem
 #endif
 
+// Use these macros to enable faster math.  Put them around one or more
+// functions.  Usage:
+//
+// MARIAN_FFAST_MATH_BEGIN
+// void LayerNormalization(float *arg) { *arg += 1.0; }
+// void SomethingElse() {}
+// MARIAN_FFAST_MATH_END
+//
+#if defined(_MSC_VER)
+#define MARIAN_FFAST_MATH_BEGIN __pragma(float_control(precise, off, push))
+#define MARIAN_FFAST_MATH_END __pragma(float_control(pop))
+#elif defined(__clang__)
+#define MARIAN_FFAST_MATH_BEGIN _Pragma("float_control(precise, off, push)")
+#define MARIAN_FFAST_MATH_END _Pragma("float_control(pop)")
+#elif defined(__GNUC__)
+// Also available as __attribute__((optimize("-fassociative-math"))) but done as pragmas for consistency
+#define MARIAN_FFAST_MATH_BEGIN _Pragma("GCC push_options") _Pragma("GCC optimize(\"-ffast-math\")")
+#define MARIAN_FFAST_MATH_END _Pragma("GCC pop_options")
+#endif
+
 namespace marian {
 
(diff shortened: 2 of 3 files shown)


emjotde:
In general great, however fast math has been disabled on purpose since it messes with isnan and isinf on the CPU side which then is a problem for training and error detection.

emjotde:
Ah, this is local only?

kpu:
This is local only.  I turned it on for `LayerNormalizationGrad` too.  

I'd like to note my current version passed all three builds on the first try despite needing compiler-specific pragmas to enable fast math locally.

kpu:
> since it messes with isnan and isinf on the CPU side which then is a problem for training and error detection.

In this case it is more numerically stable because the compiler can use associativity.  In practice, the compiler is vectorizing addition to sum in 4-16 buckets then horizontally adding the results.  So yes it "messes" with isnan and isinf because the order of addition is different and you might/might not reach an infinity as a result of the ordering being different.  But vector arithmetic is still perfectly capable of generating NaNs.  

Note gcc has `-fassociative-math` and clang >= 11 has `#pragma clang fp reassociate` for more fine-grained control.  But I don't think MSVC has that fine-grained of control: https://docs.microsoft.com/en-us/cpp/build/reference/fp-specify-floating-point-behavior?view=vs-2019

src/common/definitions.h
```diff
@@ -22,6 +22,26 @@
 #define DONT_OPTIMIZE // silently ignore on Visual Studio, where this is less of a problem
 #endif
 
+// Use these macros to enable faster math.  Put them around one or more
```
frankseide, line 25: I don't think we use the double-space-after-period convention anywhere else in the code. Let's be consistent.

src/tensors/cpu/tensor_operators.cpp
```diff
@@ -1071,15 +1065,55 @@ void LayerNormalization(Tensor out_,
 #pragma omp simd
     for(int i = 0; i < cols; ++i) {
       float t = alpha[alphaStride * i] * ((sp[i] - mean) / sigma);
-      if(beta != nullptr) {
+      if(hasBeta) {
         t += beta[betaStride * i];
```
frankseide, line 1069: Consider removing the curlies here, to keep the code vertically more compact

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 
 ### Added
+- Optimize LayerNormalization on CPU by 6x
```
frankseide, line 12: can you add what you did to get that?
...by enabling fast math

src/common/definitions.h
```diff
@@ -22,6 +22,26 @@
...
+// functions.  Usage:
+//
```
frankseide, line 27: Could you add to the comment what the trade-offs are? I.e. what are we (potentially) losing by switching to faster math and why is that OK?

src/tensors/cpu/tensor_operators.cpp
kpu: I tracked down where that brace came from.  It was @emjotde in d9d66f41 .  Then @frankseide touched the next line in 77a420740c79782d251bfe43a10279e3cfb10781 but left the brace as is.  Just following the style you set.

emjotde:
Regression tests look good too. Once the comments are addressed this is good to be merged. In case this is not done by me or @snukky, please do a squash merge.

emjotde:
And the comment you made above about isnan and the buckets probably deserves to be put in the code as well. Good information.

src/common/definitions.h
kpu: 3078927

src/tensors/cpu/tensor_operators.cpp
kpu: 3078927

src/common/definitions.h
kpu: Long comment. 3078927

CHANGELOG.md
kpu: 3078927

emjotde:
Looks like GitHub requires the reviewer who requested changes to sign off on them. In DevOps we have "Approve with suggestions" options which I assume would have been Frank's first choice here. But let's stick to the GitHub way for now.

frankseide:
Looks like the trick to go through the previous comments is to search for "show conversations'...

PR #696, opened 2020-08-09
Partial CUDA11 Support

XapaJIaMnu:
CUDA11 has removed `cusparseScsr2csc`, `cusparseScsrmm` and deprecated `cusparseSgemmi`.

I have implemented the replacement for `cusparseScsr2csc` and wrapped the other two calls in `#IFDEF` guards, with `ABORT("Not implemented");` calls. 

Since those functions are only needed for the `CSRProd` implementation (And the interface has change substantially, to the point that `CSRProd` should probably be completely rewritten to use the newer implementations) and `CSRProd` is not even tested in the regression-tests (At least as far as I can tell), I think we can stub it for now and let CUDA 11 users still use marian.

### Checklist


snukky:
The current master compiles with CUDA 11 after 83a0af2, so closing this PR. We will add a build with CUDA 11 to GitHub workflows with #704.

frankseide:
AFAIK it builds but does not actually work (the deprecated functions were replaced by stubs that ABORT, like Nick did for 2 of them). I ran into this this week, and had to downgrade to CUDA 10, to be able to decode with factored models.

Let's leave this Issue open until it is actually implemented fully.

It would also be good if we could pull in Nick's fix for one of the 3 functions.

Do we have tests for these functions?

emjotde:
This is a PR thought that doesn't really do anything?

frankseide:
I understand it includes an implementation of at least one of the three needed functions, `cusparseScsr2csc()`. Can we merge that one? But it would be great to have a test case for it.

emjotde:
Oh, indeed. I misunderstood the description and though these are only ifdefs around aborts.

emjotde:
@XapaJIaMnu just so I understand correctly, why did re-implemented that one function specifically? Just because it was easier than the others or is there a specific reason?

XapaJIaMnu:
@emjotde yes, it was the easiest to implement with relatively straight forward one to one equivalent and requiring only one additional argument. I spent some time looking at the rest, but it was going to be a lot more work, as the matrix format conversion is done differently now. (I think you're supposed to create a matrix struct.). Also, I think that with the cuda11 functionality a shorter version using less cublas calls can be written.

I didn't find any tests in the codebase for those, unfortunately.

emjotde:
How about https://github.com/marian-nmt/marian-dev/blob/e9d7de4499f1e171c4f1f544f620d8ee30414878/src/tests/units/operator_tests.cpp#L397

@frankseide does this contain the required tests?

frankseide:
This looks like it is testing all combinations. Why does the test not fail with CUDA 11?

We should also see to what degree the test succeeds for CPU. At least the combinations needed for inference are implemented for CPU.

emjotde:
It probably fails but needs to be compiled and actually run. 
```
cmake -DCOMPILE_TESTS=on ..
make -j
./src/tests/units/run_operator_tests
```

snukky:
> This looks like it is testing all combinations. Why does the test not fail with CUDA 11?

They fail. I have checked.

CUDA unit tests are not run on GitHub checks, because GitHub-hosted VMs do not have GPUs.

snukky:
Or just:
```
cmake -DCOMPILE_TESTS=on ..
make -j
make test
```

emjotde:
OK, so with this PR one less test should fail, then. @XapaJIaMnu can you check?

XapaJIaMnu:
@emjotde there isn't a test for the specific CUSPARSE function. All three functions are needed to implement the CSRProd and the tests test just the CSRProd, not the CUSPARSE subroutines. As such with the existing test suites, I can't tell if my implementation is correct or not until the other two functions are (re)implemented.

src/tensors/gpu/prod.cpp
```diff
@@ -363,6 +363,9 @@ static cusparseSgemmiEx(cusparseHandle_t handle, int m,
   const float *cscValB, const int *cscColPtrB, const int *cscRowIndB, const float *beta,
   float *C, int ldc)
 {
+#if CUDA_VERSION >= 11000
+  ABORT("cusparseSgemmi is deprecated in CUDA VERSION >= 11.");
```
frankseide, line 367: deprecated --> no longer available

src/tensors/gpu/prod.cpp
```diff
@@ -426,6 +430,44 @@ void CSRProd(marian::Tensor C,
     St_indices = allocator->alloc<int>(numValues);
     St_offsets = allocator->alloc<int>(colsS + 1);
     // transpose the second argument
+#if CUDA_VERSION >= 11000
+    size_t buffer_size;
```
frankseide, line 434: Since we cannot test this, but I would like to not lose this contribution, how about we add an ABORT at the end of this block, saying "This code is untested. Please remove this ABORT once tests exist and pass."

XapaJIaMnu:
@frankseide I have addressed the comments and merged with master. Now, both your and my workarounds to CUDA11 are in the code. I'm not sure that makes sense or not.

frankseide:
No, please merge the code in a meaningful way.

XapaJIaMnu:
I removed your ifdefs at the beginning of the file.
Is that acceptable?

src/tensors/gpu/prod.cpp
```diff
@@ -495,12 +481,16 @@ void CSRProd(marian::Tensor C,
...
         /*idxBase=*/ CUSPARSE_INDEX_BASE_ZERO));
+#endif
     std::swap(rowsS, colsS); // these variables now represent the dims of the explicitly transposed object
   }
   if (swapOperands) {
     // C = D x S for row-major matrices
     // Implemented via cusparse as C' = S' x D' ("csrmm") where C' and D' are column-major,
     // and S' is CSR (if not transS then we make a transposed copy).
+#if CUDA_VERSION >= 11000
+    ABORT("CSRProd is not yet implemented for CUDA VERSION >= 11");
```
emjotde, line 492: Is this error message correct? This seems to only be true for the transpose?

src/tensors/gpu/prod.cpp
```diff
@@ -483,6 +430,45 @@ void CSRProd(marian::Tensor C,
...
+                                          /*cscRowInd=*/ St_indices->data<int>(),
+                                          /*valType=*/ CUDA_R_32F,
+                                          /*copyValues=*/ CUSPARSE_ACTION_NUMERIC,
+                                          /*idxBase=*/ CUSPARSE_INDEX_BASE_ZERO,
+                                          /*alg=*/ CUSPARSE_CSR2CSC_ALG1,
+                                          /*buffer=*/ buffer->data<uint8_t>()));
+
+    if (buffer)
+      allocator->free(buffer);
+    ABORT("This code is untested. Please remove this ABORT once tests exist and pass.");
```
emjotde, line 470: How about the tests mentioned in the comments to the PR in `operator_tests.cpp`?

emjotde: Ah, saw the comment concerning the tests.

src/tensors/gpu/prod.cpp
emjotde: Too minor to block merging.

emjotde:
@XapaJIaMnu merged, thanks!

PR #698, opened 2020-08-18
Enable compute75 when using cuda10 

XapaJIaMnu:
# Description
At the moment we're compiling with all Compute architectures that CUDA 9.0 supports. This has the major shortcoming that we're not compiling for turing, missing out on the performance for the most desired training GPU - the 2080ti. At the same time, we're passing a bunch of targets that are deprecated in CUDA11, leaving us with tons of compile warnings. 

This leads to a small increase in training speed on a 2x2080ti 38361 -> 38720 and in a nice increase in translation speed on my desktop 2060: Translating WMT19 on ende student.tiny11(fp32) went down from 4.46s to 3.76 seconds.

What we should do instead is look at the CUDA version found and enable the appropriate compute architectures
List of changes:
-  Enable appropriate compute architectures in GPU compilation based on the CUDA version found.
Added dependencies: none

### Checklist


CMakeLists.txt
```diff
@@ -235,6 +231,29 @@ if(CUDA_FOUND)
       message(WARNING "On some Unix systems CUDA 10.0+ requires CMake 3.12.2+; you use CMake ${CMAKE_VERSION}")
   endif()
 
+  # We want to compile as many targets as possible but different CUDA versions support different targets.
+  # Let's instead enable options based on what cuda version we have.
+  if((CUDA_VERSION VERSION_EQUAL "9.0" OR CUDA_VERSION VERSION_GREATER "9.0") AND CUDA_VERSION VERSION_LESS "11.0")
+    option(COMPILE_CUDA_SM35 "Compile GPU version with SM35 support" ON)
```
snukky, line 237: This creates different sets of options, which are not known from the beginning. I can see potential issues here. It might be better to only *set values* of those options here instead of creating them; see the very first comment.

CMakeLists.txt
```diff
@@ -235,6 +231,29 @@ if(CUDA_FOUND)
...
+    option(COMPILE_CUDA_SM70 "Compile GPU version with SM70 support" ON)
+  endif()
+  if((CUDA_VERSION VERSION_EQUAL "10.0" OR CUDA_VERSION VERSION_GREATER "10.0") AND CUDA_VERSION VERSION_LESS "11.0")
+    option(COMPILE_CUDA_SM35 "Compile GPU version with SM35 support" ON)
+    option(COMPILE_CUDA_SM50 "Compile GPU version with SM50 support" ON)
+    option(COMPILE_CUDA_SM60 "Compile GPU version with SM60 support" ON)
+    option(COMPILE_CUDA_SM70 "Compile GPU version with SM70 support" ON)
+    option(COMPILE_CUDA_SM75 "Compile GPU version with SM75 support" ON)
+  endif()
+  if((CUDA_VERSION VERSION_EQUAL "11.0" OR CUDA_VERSION VERSION_GREATER "11.0") AND CUDA_VERSION VERSION_LESS "12.0")
```
snukky, line 249: Can we skip `CUDA_VERSION VERSION_LESS "12.0"`?

snukky (changes requested):
Thanks Nick for the pull request! I think it might be clearer to keep the set of options always the same and manipulate specific options depending on the available CUDA version.

CMakeLists.txt
```diff
@@ -235,6 +231,29 @@ if(CUDA_FOUND)
       message(WARNING "On some Unix systems CUDA 10.0+ requires CMake 3.12.2+; you use CMake ${CMAKE_VERSION}")
   endif()
 
+  # We want to compile as many targets as possible but different CUDA versions support different targets.
+  # Let's instead enable options based on what cuda version we have.
```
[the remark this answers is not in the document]
XapaJIaMnu: No. I have specifically removed options not supported by the CUDA version. The user will only be able to set options supported by the CUDA version they are using. Otherwise the compilation will either fail, or produce warnings about the use of deprecated architecture.

CMakeLists.txt
XapaJIaMnu: Yes, we should so that when CUDA12 gets out ppl can get reasonable binaries  in most cases.

CMakeLists.txt
```diff
@@ -13,10 +13,6 @@ set(BUILD_ARCH native CACHE STRING "Compile for this CPU architecture.")
 # Custom CMake options
 option(COMPILE_CPU "Compile CPU version" ON)
 option(COMPILE_CUDA "Compile GPU version" ON)
-option(COMPILE_CUDA_SM35 "Compile GPU version with SM35 support" ON)
-option(COMPILE_CUDA_SM50 "Compile GPU version with SM50 support" ON)
-option(COMPILE_CUDA_SM60 "Compile GPU version with SM60 support" ON)
-option(COMPILE_CUDA_SM70 "Compile GPU version with SM70 support" ON)
```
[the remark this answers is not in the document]
XapaJIaMnu: I considered this and opted against it because this solution seemed hacky for me: https://stackoverflow.com/questions/3766740/overriding-a-default-option-value-in-cmake-from-a-parent-cmakelists-txt
eg, what happens if we disable the option but the user specifically enables it? Then the compilation might fail due to unsupported compute architecture?  It also wasn't clear for me what takes precedence? The `CMakeLists.txt` option or the user command line input when invoking `cmake` but I can test that easily. Still, I don't think it's a better solution.

CMakeLists.txt
XapaJIaMnu: See my first comment. How do you deal with overriding user input? Doesn't this get even more confusing?

XapaJIaMnu:
More on why we should be using the proper compute flags from the CUDA documentation:  https://docs.nvidia.com/cuda/turing-tuning-guide/index.html#tensor-operations

> Any binary compiled for Volta will run on Turing, but Volta binaries using Tensor Cores will only be able to reach half of Turing's Tensor Core peak performance. Recompiling the binary specifically for Turing would allow it to reach the peak performance. See the Turing Compatibility Guide for more information.

XapaJIaMnu:
@snukky I re-enabled the deprecated architectures with CUDA11. Now the compiler will generate lots (but harmless) warnings.

CMakeLists.txt
```diff
@@ -243,6 +239,31 @@ if(CUDA_FOUND)
       message(WARNING "On some Unix systems CUDA 10.0+ requires CMake 3.12.2+; you use CMake ${CMAKE_VERSION}")
   endif()
 
+  # We want to compile as many targets as possible but different CUDA versions support different targets.
+  # Let's instead enable options based on what cuda version we have.
```
snukky, line 243: With the very last changes, does this comment need to be updated?

CMakeLists.txt
```diff
@@ -243,6 +239,31 @@ if(CUDA_FOUND)
...
+  if(CUDA_VERSION VERSION_EQUAL "11.0" OR CUDA_VERSION VERSION_GREATER "11.0")
+    option(COMPILE_CUDA_SM35 "Compile GPU version with SM35 support" ON)
+    option(COMPILE_CUDA_SM50 "Compile GPU version with SM50 support" ON)
+    option(COMPILE_CUDA_SM60 "Compile GPU version with SM60 support" ON)
+    option(COMPILE_CUDA_SM70 "Compile GPU version with SM70 support" ON)
+    option(COMPILE_CUDA_SM75 "Compile GPU version with SM75 support" ON)
+    option(COMPILE_CUDA_SM80 "Compile GPU version with SM80 support" ON)
+  endif()
+
+
```
snukky, line 266: Please remove one superfluous blank line.

CMakeLists.txt
XapaJIaMnu: No, because Cuda 9.0 enables `35 50, 60 and 70`, while CUDA10 enables `75` on top of it and CUDA11 enables `80` on top of 10. 
When CUDA12 is released, we'd have to drop `35 and 50`.

emjotde:
@snukky have your concerns been addressed?

snukky:
There was a question if disabling older architectures would make Marian compiled with new CUDA not runnable on older GPUs that we used to support, but since recent changes enable all previous architectures regardless of CUDA 11 warnings, this PR is mostly adding support for new compute architectures and we can merge it.

snukky (approved):

XapaJIaMnu:
I have missed it, accidentally. I will add it later today.

Is there any reason why there are separate NVCC flags here?

snukky:
I don't know. I think that's a question for @emjotde .

PR #703, opened 2020-08-23
Squashed commit of fixes to graceful shutdown via SIGTERM

ugermann:
### Description

This pull request replaces #660. I believe it fixes all the issues raised by @frankseide with respect to that. 
I created a new PR to get github to trigger a new diff to facilitate final code review.

Currently (master) SIGTERM will already trigger a graceful shutdown/exit (i.e., Marian will save models before exiting upon SIGTERM), but batch generation in the background will not pay attention to it. This can mean considerable delays in shutdown. This PR fixes that by inserting checks for graceful exit requests in the batch generation loop.

In addition, the graceful exit behavior upon SIGTERM can now be disabled (set to the default of just terminating everything without saving or any other cleanup) on the command line with `--sigterm immediate`.
Please add a clear and concise description of the changes.

This PR fixes a bug/adds a new feature/refactorizes the code/does something else.
It is related to issues: #998, #999, ...

List of changes:
- addresses concerns by @frankseide. These are mostly cosmetic, such as removing unneccesary braces; plus having a generic flag for graceful exist rather than tying this to a specific signal flag.
- graceful exit via SIGTERM can now be disabled via a configuration switch (see above).
Added dependencies: none

### How to test

1. Start training, wait for the first update, then send termination signal (i.e., `kill <process id>`). Marian should save the model before exiting.

 2. Start training with `--sigterm immediate`, wait for the first update, then send termination signal. Marian should exit immediately.
 
### Checklist


emjotde:
Can you clean up the description please?
Was it necessary to abandon the old PR, it's now hard to compare against old comment? Or should be we just treat that as completely new?

emjotde:
From a quick look at the code, it looks rather right you mostly re-did it. That's OK then.

ugermann:
@emjotde Description updated. What I did was branch from the latest master, did a squashed merge with ug-graceful-shutdown, and committed. If anyone can tell me how I can get github to update the three-way diff after responding to review comments, I'd be very grateful.

emjotde:
Is this different from https://github.com/marian-nmt/marian-dev/pull/660 now? (I reopened it to check).

emjotde:
There are only tiny diffs from what I can see (two cosmetic differences), so it seems to me we should keep working on #660 then?

Could it be you misunderstood my comment a while ago about "squashing"? I meant that PRs into master should be squashed via the Github interface if it happens that someone other than Roman or me does the merging. Didn't really apply to anything else, that might have been misleading on my part.

ugermann:
Functionally only in the sense that I've added a configuration option to disable graceful shutdown. All other changes are cosmetic:
- I've removed unneccessary braces
- I've introduced a generic global flag (plus checking function) for shutdown instead of checking specifically for the SIGTERM flag (as requested by Frank), and introduced a generic signal handler for requesting a graceful shutdown.

I created a new branch + PR mostly because I was not able to get github to update the diff for code review to the current state of the code with the old PR. It seemed the shortest wayt to getting the changes into master, both on my side and on the reviewer's side. It's not many changes to review, and I think they are straightforward.

emjotde:
Let me see what git does there. Internally we have the luxury to be able to work with CodeFlow which makes those diffs between commits etc. easy. I think generally it's preferable to stay within one PR to have a continues history of comments and reviews, but if the interface doesn't play well with that, that's a different story.

emjotde:
So, from what I see is this: git allows you to review changes from "your last review" or from a specific commit. So both, @frankseide  and me, can look at your changes after the reviews there. Would it then make sense to close this PR and rather finish #660 if that is ready to be looked at again in your opinion?

ugermann:
Whatever you prefer, the code is currently the same.

emjotde:
OK, then let's move back over to #660, thanks. I am closing this one for now.

PR #705, opened 2020-08-28
Print message that marian-server is listening on port X after it is accepting connections

### LINKED ISSUE #701: Server isn't actually listening on port when log message says it is
The log message saying "Server is listening on port" appears before the server is started:

https://github.com/marian-nmt/marian-dev/blob/a21e48fb4f0beec0f83987112ff61b23a0140080/src/command/marian_server.cpp#L55-L56

It's a race condition because one might look for that message to start the client.  

It's also misleading because e.g. the message will still be printed if the port is already in use or some other error happens in starting the server.

snukky:
### Description
Fixes #701 adding a callback function to `server.start()`: https://github.com/marian-nmt/Simple-WebSocket-Server/blob/master/server_ws.hpp#L370.

Added dependencies: none

### How to test
Tested manually and run regression tests for marian-server, e.g. `./run_mrt.sh '#server'`.

### Checklist


src/command/marian_server.cpp
```diff
@@ -52,8 +52,7 @@ int main(int argc, char **argv) {
 
   // Start server thread
   std::thread serverThread([&server]() {
-    LOG(info, "Server is listening on port {}", server.config.port);
-    server.start();
+    server.start([](unsigned short port) { LOG(info, "Server is listening on port {}", port); });
```
kpu, line 55: That's a lot of action for one line of code, maybe newlines would help here.

kpu (approved):
The code appears correct, just suggesting formatting changes.

emjotde:
@snukky do you want to change anything here or good to go?

Issue #663, opened 2020-06-24
Batched GEMM produces different output compared to the normal GEMM


ykim362:
### Bug description
Batched GEMM generates different output compared to the normal GEMM.
https://github.com/marian-nmt/marian-dev/blob/master/src/tensors/cpu/prod.cpp#L99
Due to the impact on the Mircrosoft production systems, this needed to be disabled in the MS' internal repo.

### How to reproduce
One of the German-English systems produces around 30 different outputs among 3004 sentences from WMT 2017 test data.

### Context
* Marian version: any version with Batched GEMM is implemented and enabled
* CMake command: cmake -DUSE_FBGEMM=ON -DCOMPILE_CUDA=OFF ..
* Log file: Attach your training/decoding logs

Add any other information about the problem here.

emjotde:
Can you take a look at the n-best lists? Are the same hyps/scores in there? I could imagine that floating point computation order might have an influence?

ykim362:
@XapaJIaMnu Could you take a look at this? I had to disable this in Microsoft's internal repo for the prod model mismatch.

XapaJIaMnu:
@ykim362 , as far as I can tell, the code is correct. I suspect that the batched function produces slightly different floating point scores in some cases. Could you post the n-best list (or path scores) for one or two of the offending test cases? I'll reach out to my contacts at Intel and see what they have to say about `cblas_sgemm_batched`.

frankseide:
Do we have an operator test for batched GEMM? Can we try to reproduce the difference and create a test that exhibits that difference?

Thanks,

Frank

emjotde:
It's passing regression tests for a transformer model for translation that uses batched gemm and that was created with the previous function. We would need to catch that specific case.

ykim362:
I will try to provide the problem case.

XapaJIaMnu:
@ykim362 any progress on that? I personally think it's safe to re-enable it, or at least put it in as a command line option, as opposed to hidden behind an `#if 0`

ykim362:
@XapaJIaMnu I have a FSM model which can be shared. I will get back to you with repro. Thanks.

ykim362:
@XapaJIaMnu hmm.. It's not reproducible in the environment I have. Should we just move this back in?

XapaJIaMnu:
I think so. I couldn't reproduce this in the regression tests and neither could @emjotde.

Cheers,

Nick

ykim362:
[...]

PR #710, opened 2020-09-02
Re-enable mkl based batched GEMM

ykim362:
### Description
We disabled mkl based batched GEMM due to MS prod accuracy related issue. In the latest branch the issue is not reproducible. So, we are re-able this.

This PR is related to #663 .

List of changes:
- Enable MKL's batched GEMM

Added dependencies: none

### How to test
Run any transformer based model

### Checklist


XapaJIaMnu (approved):
Looks good to me. The old codepath will be used with OpenBLAS and the likes which don't have `cblas_sgemm_batched`.

emjotde:
@snukky not related to the PR, but since I am seeing it here, are we keeping the failing ubuntu 20.04 build? It looks a bit misleading, as if the error was due to PR and not to the FBGEMM situation.

Issue #709, opened 2020-09-02
Marian's FBGEMM does not compile on CPUs with VNNI using GCC 9.3+


snukky:
### Bug description
It seems that Marian can not be compiled with FBGEMM on machines with new CPUs supporting VNNI using GCC 9.3 or 10.0. It produces a long list of errors like these:

    /tmp/ccWTYViL.s: Assembler messages:
    /tmp/ccWTYViL.s:49532: Error: invalid use of register

We see this occasionally on GitHub CI, e.g. [this build](https://github.com/marian-nmt/marian-dev/runs/1055063509?check_suite_focus=true), and I can reproduce it on my laptop on Ubuntu 20.04 WSL. Marian+FBGEMM compiles without issues on 1) a machine without VNNI using GCC >=9.3, or 2) on a machine with VNNI using GCC <9, tested with GCC 8.4, 7.5 and 5.5.

GCC 9.3 is the default on Ubuntu 20.04, so the issue may become more prominent over time.

`marian-nmt/fbgemm` can be compiled with any tested GCC version, however, some unit tests in PackedRequantizeAcc16Test fail. They work if compiling the latest version of `pytorch/fbgemm`. Thanks to @XapaJIaMnu for suggesting this.

@ykim362 Have you experienced this? Maybe we should try updating FBGEMM used in Marian?

### How to reproduce
Compile Marian on a machine with CPU that has `avx512_vnni` (on Ubuntu it can be checked by `cat /proc/cpuinfo | grep avx`)

    CC=/usr/bin/gcc-9 CXX=/usr/bin/g++-9 \
    cmake .. -DUSE_SENTENCEPIECE=on -DUSE_FBGEMM=on -DCOMPILE_CUDA=off -DCOMPILE_CPU=on

### Context
* Marian version: I think any recent version of marian-dev has this issue
* CMake command: see above
* Log file: [compilation logs from GitHub checks](https://github.com/marian-nmt/marian-dev/runs/1055063509?check_suite_focus=true)

CMake 3.16.3, MKL 2020.0.1/2020.0.0, GCC 9.3.0.
Tested on Ubuntu 20.04 WSL and confirmed on non-virtual Ubuntu 18.04.

XapaJIaMnu:
Related, #607

emjotde:
We have seen this on some of the randomly assigned machines for CI. If upstream FBGEMM works, would be good to merge. @ykim362

ykim362:
[...]

PR #717, opened 2020-09-06
Disable GitHub check with Ubuntu 20.04 and GCC 9

snukky:
### Description
The GitHub check with Ubuntu 20.04, GCC 9 and FBGEMM fails more often than expected, which is misleading, e.g. https://github.com/marian-nmt/marian-dev/pull/710#issuecomment-687710304

It could be enabled again after fixing #709.

Added dependencies: none

### DIFF
--- a/.github/workflows/ubuntu-gpu.yml
+++ b/.github/workflows/ubuntu-gpu.yml
@@ -16,11 +16,11 @@ jobs:
       matrix:
         include:
           # Ubuntu 20.04 supports CUDA 11+
-          - os: ubuntu-20.04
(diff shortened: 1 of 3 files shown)


emjotde (approved):

emjotde:
Merge whenever you feel like it.

emjotde:
To make a point, this PR is failing the check it's meant to fix :)

snukky:
Replaced Invoke-WebRequest with a wget command (it seems to be available on Windows runners) as it retry downloading by default, so hopefully jobs will be more stable. Wget is also faster.

PR #746, opened 2020-10-16
Fixes bug for certain reductions

rhenry-nv:
### Description
This PR fixes a bug for reductions when all of the following conditions are met:

1) We are reducing into a scalar variable
2) There are <= 32 elements in the input tensor
3) The identity of the reduction operator is not 0

This bug *may* only be present on Volta architectures and newer. I have not checked older architectures. An example of this is performing a max reduction into a scalar. This will only show up if all of the values in the input are negative due to the 3rd requirement.

List of changes:
- Changes the size of the final cta tile to be the min(32, blockSize) instead of just 32. This caused 0s to flow in from the warp shuffle instructions

### How to test
Run the unit tests. Added a new unit test under operator tests that fails on the original master branch and passes with this commit.

OS and CMake command:

OS: Ubuntu 18.04.3 LTS
gcc 7.5.0
nvcc 10.1.243

Ran with:
cmake .. -DCOMPILE_TESTS=on -DCOMPILE_CPU=on -DCOMPILE_CUDA=on -DUSE_SENTENCEPIECE=on -DUSE_STATIC_LIBS=off -DCOMPILE_SERVER=off -DUSE_FBGEMM=on -DCOMPILE_CUDA_SM35=off -DCOMPILE_CUDA_SM50=off -DCOMPILE_CUDA_SM60=off -DCOMPILE_CUDA_SM70=on 

### Checklist

### DIFF
--- a/src/3rd_party/reduce_all.h
+++ b/src/3rd_party/reduce_all.h
@@ -129,15 +129,17 @@ __global__ void reduceSinglePass(Functor functor, AccType aggInit, AggFunctor ag
 
   cg::sync(cta);
 
-  cg::thread_block_tile<32> tile32 = cg::tiled_partition<32>(cta);
+  // leverage that blockSize is always pow of 2 so no special logic needed in reduction loop.
+  constexpr int partitionSize = blockSize > 32? 32 : blockSize;
+  cg::thread_block_tile<partitionSize> tile = cg::tiled_partition<partitionSize>(cta);
 
   if (cta.thread_rank() < 32) {
     // Fetch final intermediate sum from 2nd warp
     if (blockSize >= 64) 
       mySum = aggFunctor(mySum, sdata[tid + 32]);
     // reduce final warp using shuffle
-    for (int offset = tile32.size() / 2; offset > 0; offset /= 2) {
-      mySum = aggFunctor(mySum, tile32.shfl_down(mySum, offset));
+    for (int offset = tile.size() / 2; offset > 0; offset /= 2) {
+      mySum = aggFunctor(mySum, tile.shfl_down(mySum, offset));
     }
   }
 
--- a/src/3rd_party/reduce_all.h
+++ b/src/3rd_party/reduce_all.h
@@ -130,7 +130,7 @@ __global__ void reduceSinglePass(Functor functor, AccType aggInit, AggFunctor ag
   cg::sync(cta);
 
   // leverage that blockSize is always pow of 2 so no special logic needed in reduction loop.
-  constexpr int partitionSize = blockSize > 32? 32 : blockSize;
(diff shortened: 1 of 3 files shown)


src/3rd_party/reduce_all.h
```diff
@@ -129,15 +129,17 @@ __global__ void reduceSinglePass(Functor functor, AccType aggInit, AggFunctor ag
 
   cg::sync(cta);
 
-  cg::thread_block_tile<32> tile32 = cg::tiled_partition<32>(cta);
+  // leverage that blockSize is always pow of 2 so no special logic needed in reduction loop.
+  constexpr int partitionSize = blockSize > 32? 32 : blockSize;
```
ykim362, line 133: Please add a space before '?'.

src/tests/units/operator_tests.cpp
```diff
@@ -297,13 +328,14 @@ void tests(DeviceType device, Type floatType = Type::float32) {
     s2->val()->get(values); CHECK(values == vS2);
 
     CHECK(m3->val()->scalar() == 9);
+    auto floatApproxLocal = [](T x, T y) -> bool { return x == Approx(y).margin(0.004); };
```
ykim362, line 331: Just curious. Why 0.004?

src/tests/units/operator_tests.cpp
```diff
@@ -66,6 +70,33 @@ void tests(DeviceType device, Type floatType = Type::float32) {
...
+      
+      float reduced = vA[0];
+      for(const auto val : vA) {
+        reduced = op(reduced, val);
+      }
+      return floatApprox(reduced, val);
+    };
+
+    // @TODO: add all operators here for completeness
+    auto maxReduce = max(a, -1);
```
ykim362, line 93: Could you add /*axis=*/ before -1?
And, it would be great to add a few more reduction operators.

src/tests/units/operator_tests.cpp
rhenry-nv: Ah - this test was failing on my Titan V so I just increased the margin so that it passed. Should I revert this?

ykim362: Got it. No need to revert. Thanks!

snukky: Could you please add a short comment to the code with that explanation?

rhenry-nv: Done

src/tests/units/operator_tests.cpp
rhenry-nv: Done

emjotde:
Awesome, thank you.

emjotde (approved):

Issue #683, opened 2020-07-22
Error in decoding using packed binary model (packed8avx512)


robberlang:
### Bug description
A fatal error occurs when decoding with a model that was converted to `packed8avx512` GEMM type from a .npz formatted model using `marian-conv`. I have a few models where this happens, and I also have models where it does not.
The error message is `Error: Actual pathScore (-inf) is lower than INVALID_PATH_SCORE (-3.40282e+38)??` when the beam size is 2 or 3, and is `Error: No hypotheses in n-best list??` when the beam size is 1.

### How to reproduce
marian-conv -f model.npz -t model.bin -g packed8avx512
echo 'test' | marian-decoder -b `<beam-size>` --cpu-threads 1 -m model.bin -v vocab.src.spm vocab.trg.spm

### Context
* Marian version: v1.9.28; b28905a2 2020-07-21 11:32:08 +0100
* CMake command: `cmake -DUSE_SENTENCEPIECE:BOOL=ON -DCOMPILE_CPU:BOOL=ON -DCOMPILE_CUDA:BOOL=OFF -DUSE_CUDNN:BOOL=OFF -DUSE_FBGEMM:BOOL=ON -DUSE_NCCL:BOOL=OFF ..`
Output of `--build-info all`:
[marian-build-info-all.txt](https://github.com/marian-nmt/marian-dev/files/4960501/marian-build-info-all.txt)

* Log file: 
[decode-b1.log](https://github.com/marian-nmt/marian-dev/files/4960493/decode-b1.log)
[decode-b2.log](https://github.com/marian-nmt/marian-dev/files/4960494/decode-b2.log)

No problems when using a model converted to the `float32` type with `marian-conv`.

emjotde:
Whoa. That's really odd. Can you share the model by any chance?

robberlang:
I sent you an email with info for getting the model and vocab files. Thanks.

emjotde:
Got the e-mail, thanks.

XapaJIaMnu:
A bit late, but.. Could it be that you have an `avx2` machine? Does `packed8avx2` work for you?

robberlang:
[...]

PR #747, opened 2020-10-29
No quantization of parameter with name ending in _Wt

robberlang:
### Description
marian-conv does not work properly with models trained with `tied-embeddings` and `tied-embeddings-all` both set to false.
This PR removes quantization of parameters with name ending in _Wt, which is used for the vocabs when `tied-embeddings` and `tied-embeddings-all` are both set to false (see the logic around `tiedParam_` in `mlp::Output::lazyConstruct()` in `src/layers/generic.cpp`, and for transformer models, see `DecoderTransformer::lazyCreateOutputLayer()` in `src/models/transformer.h`): if either `tied-embeddings` and `tied-embeddings-all` are set to true then the parameter name for the vocab is `Wemb` or ends in `_Wemb`. These parameters are not quantized.

This PR fixes a bug/adds a new feature/refactorizes the code/does something else.
It is related to issues: #683

List of changes:
Parameters with names ending in _Wt added to logic of those to not quantize.

Added dependencies: none

### How to test
`marian-conv -f model.npz -t model.bin -g packed8avx512`
`echo 'test' | marian-decoder -b <beam-size> --cpu-threads 1 -m model.bin -v vocab.src.spm vocab.trg.spm`
The error message is `Error: Actual pathScore (-inf) is lower than INVALID_PATH_SCORE (-3.40282e+38)??` when the beam size is 2 or 3, and is `Error: No hypotheses in n-best list??` when the beam size is 1. With this PR, normal translation occurs.

Also compare results with and without PR of `marian-conv -f model.npz -t model.bin -g packed8avx2` and `marian-conv -f model.npz -t model.bin -g packed16`

Linux
`cmake -DUSE_SENTENCEPIECE:BOOL=ON -DCOMPILE_CPU:BOOL=ON -DUSE_FBGEMM:BOOL=ON ..`

### Checklist

### DIFF
--- a/src/tensors/cpu/fbgemm/expression_graph_packable.h
+++ b/src/tensors/cpu/fbgemm/expression_graph_packable.h
@@ -39,7 +39,8 @@ public:
       // int8 - all the weights used for affine op and dot op
       // fp16 - all the weights used for affine op
       if ((gemmElementType == Type::packed8avx2 || gemmElementType == Type::packed8avx512)
-        && (pName.find("_W") == pName.length() - 3 || pName.find("_W") == pName.length() - 2)) {
+         && ((pName.find("_W") == pName.length() - 3 && pName.back() != 't')
+             || pName.find("_W") == pName.length() - 2)) {
 #if USE_FBGEMM
         using namespace marian::cpu::variant;
         // packing information - size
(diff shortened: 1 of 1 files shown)


ykim362:
@emjotde @snukky This fix would resolve the issue. But, in general, this converter would be a temporary solution as it depends on the string patterns in the weight names. We may want to have an input file to list the weights? In any case, I will not block this PR itself.

ykim362 (commented):
Looks fine for me as a temporary fix. But, I will let @emjotde decide.

PR #748, opened 2020-11-01
Factor concatenation improvements and documentation

kpu:
### Description

Adds options for handling factors in Marian including concatenation.  This is primarily written by @pedrodiascoelho.   

### How to test
Marian continuous integration tests are passing.  Unbabel has also been doing quality testing.  

### Checklist

- [ ] I have tested the code manually


emjotde:
The added doc is a nice touch, thank you for that. @pedrodiascoelho anything specific we should pay attention to? Is worth to work with @snukky to add regression/unit tests?

emjotde:
And a first remark, can please add comments in the github interface where you added new functionality and where you made re-factoring changes? I don't know this part of the code very well and will need a few hints. 

@snukky I will give you an old production model with factors later this week to make sure things work and there are no regressions for our factored models. We will add that to the regression tests then.

snukky:
OK, creating a regression test for the existing code is a good idea.

src/common/config_parser.cpp
```diff
@@ -194,6 +194,14 @@ void ConfigParser::addOptionsModel(cli::CLIWrapper& cli) {
...
       "Size of embedding vector",
       512);
+  cli.add<int>("--factors-dim-emb",
+      "Embedding dimension of the factors. Only used if concat is selected as factors combining form");
+  cli.add<std::string>("--factors-combine", 
+    "How to combine the factors and lemma embeddings. Options available: sum, concat",
+    "sum");   
+  cli.add<std::string>("--factor-predictor",
+      "Method to use when predicting target factors. Options: soft-transformer-layer, hard-transformer-layer, lemma-dependent-bias, re-embedding",
+      "soft-transformer-layer");
```
pedrodiascoelho, line 204: Since concatenation was implemented, two new options were created. One (-factors-combine) that controls if we want to embed the lemmas and factors by summing them (the default option. It follows the implementation written by Frank), or apply concatenation.
In case you chose concatenation a dimension of that embedding must be specified with --factors-dim-emb.

src/common/config_parser.cpp
```diff
@@ -194,6 +194,14 @@ void ConfigParser::addOptionsModel(cli::CLIWrapper& cli) {
   cli.add<int>("--dim-emb",
       "Size of embedding vector",
       512);
+  cli.add<int>("--factors-dim-emb",
+      "Embedding dimension of the factors. Only used if concat is selected as factors combining form");
+  cli.add<std::string>("--factors-combine", 
+    "How to combine the factors and lemma embeddings. Options available: sum, concat",
+    "sum");
```
pedrodiascoelho, line 201: Since concatenation was implemented, two new options were created. One (--factors-combine) that controls if we want to embed the lemmas and factors by summing them (the default option), or apply concatenation.
If sum is chosen it will follow the already implemented embedding from Frank.
In case you chose concatenation a dimension of that embedding must be specified with --factors-dim-emb.

src/common/config_parser.cpp
pedrodiascoelho, line 204: For the target-factors, since different decoding options were already implemented, a more user-friendly config variable was introduced to make more clear which are the available options

src/data/factored_vocab.cpp
```diff
@@ -655,6 +655,39 @@ std::string FactoredVocab::surfaceForm(const Words& sentence) const /*override f
   return res;
 }
 
+size_t FactoredVocab::getTotalFactorCount() const {
+  return factorVocabSize() - groupRanges_[0].second;
+}
```
pedrodiascoelho, line 660: Auxiliary function that return the total number of factors (no lemmas) in a factored vocabulary.

src/layers/generic.cpp
```diff
@@ -439,18 +442,26 @@ namespace marian {
...
     if (factoredVocab_) {
       dimVoc = (int)factoredVocab_->factorVocabSize();
       LOG_ONCE(info, "[embedding] Factored embeddings enabled");
+      if (opt<std::string>("factorsCombine") == "concat") {
+        ABORT_IF(dimFactorEmb == 0, "Embedding: If concatenation is chosen to combine the factor embeddings, a factor embedding size should be specified.");
+        int numberOfFactors = (int) factoredVocab_->getTotalFactorCount();
+        dimVoc -= numberOfFactors;
+        FactorEmbMatrix_ = graph_->param("factor_" + name, {numberOfFactors, dimFactorEmb}, initFunc, fixed);
+        LOG_ONCE(info, "[embedding] Combining factors concatenation enabled");
+      }
```
pedrodiascoelho, line 462: If concatenation is chosen the factors will be embedded with a different matrix from the lemma embeddings.

src/layers/generic.cpp
```diff
@@ -462,6 +473,21 @@ namespace marian {
...
+    factoredVocab_->lemmaAndFactorsIndexes(data, lemmaIndices, factorIndices);
+    auto lemmaEmbs = rows(E_, lemmaIndices);
+    int dimFactors =  FactorEmbMatrix_->shape()[0];
+    auto factEmbs = dot(graph->constant({(int) data.size(), dimFactors}, inits::fromVector(factorIndices), Type::float32), FactorEmbMatrix_);
+
+    auto out = concatenate({lemmaEmbs, factEmbs}, -1);
+
+    return out;
+  }
+
```
pedrodiascoelho, line 490: To embed using concatenation basically we call the lemmaAndFactorsIndexes() function that decodes each wordIndex into its lemma and factor information. Then we embed lemmas and factors separately. To embed the lemmas we just choose the rows of the lemma embedding matrix (E_) that correspond to the indexes of the lemmas in the vocab given by lemmaAndFactorsIndexes. Then to embed the factors, since that a token could have more than one factor, we cannot simple chose the rows of the factor embedding matrix as we did for the lemmas. So for the factors an array of 0's and 1's is returned, where for each token we have the information if a certain factor appears or not in a token. This vector is turned into a sparse matrix that is multiplied by the factor embedding matrix to get the factor embeddings. This two result embeddings (lemmaEmbs and factEmbs) are then concatenated.

src/layers/generic.cpp
```diff
@@ -517,12 +543,12 @@ namespace marian {
...
-    auto batchMask = graph->constant({dimWidth, dimBatch, 1},
-                                     inits::fromVector(subBatch->mask()));
-#else
     // experimental: hide inline-fix source tokens from cross attention
     auto batchMask = graph->constant({dimWidth, dimBatch, 1},
                                      inits::fromVector(subBatch->crossMaskWithInlineFixSourceSuppressed()));
+#else
+    auto batchMask = graph->constant({dimWidth, dimBatch, 1},
+                                     inits::fromVector(subBatch->mask()));
 #endif
```
pedrodiascoelho, line 552: Here only the order of what is inside the #if part of the macro and in the #else part was switched. Mainly because this was trowing a couple of warnings if the tags <IDELIM> <IOPEN> and <ICLOSE> where not used.

src/models/transformer.h
```diff
@@ -285,7 +285,8 @@ class Transformer : public EncoderOrDecoderBase {
       kh = cache_[prefix + "_keys"];                                                   // then return cached tensor
     }
     else {
-      auto Wk = graph_->param(prefix + "_Wk", {dimModel, dimModel}, inits::glorotUniform());
+      int dimKeys =  keys->shape()[-1];
+      auto Wk = graph_->param(prefix + "_Wk", {dimKeys, dimModel}, inits::glorotUniform());
```
pedrodiascoelho, line 289: The weight attention transformer matrices were being initialized with shape [dimModel x dimModel]. If concatenation is chosen and we are only using source factors, the embedding size of the encoder and decoder will be different, so in the enc-dec attention layer of the decoder, we need to initialize the weight matrices of keys and values with whatever dimension is outputted from the encoder.

src/models/transformer.h
```diff
@@ -299,7 +300,8 @@ class Transformer : public EncoderOrDecoderBase {
         && cache_[prefix + "_values"]->shape().elements() == values->shape().elements()) {
       vh = cache_[prefix + "_values"];
     } else {
-      auto Wv = graph_->param(prefix + "_Wv", {dimModel, dimModel}, inits::glorotUniform());
+      int dimValues = values->shape()[-1];
+      auto Wv = graph_->param(prefix + "_Wv", {dimValues, dimModel}, inits::glorotUniform());
```
pedrodiascoelho, line 304: see https://github.com/marian-nmt/marian-dev/pull/748/files#r520678729

src/data/factored_vocab.cpp
```diff
@@ -655,6 +655,39 @@ std::string FactoredVocab::surfaceForm(const Words& sentence) const /*override f
...
+        auto factorIndex = lemmaAndFactorIndices[g]; 
+        ABORT_IF(factorIndex == FACTOR_NOT_SPECIFIED, "Attempted to embed a word with a factor not specified");
+        for (int i = 0; i < factorShape_[g] - 1; i++) {
+          factorIndices.push_back((float) (factorIndex == i));
+        }
+      }      
+    }
+  }
+}
+
```
pedrodiascoelho, line 690: This function returns two data structures that contain separate information regarding lemmas and factors indexes and usage by receiving a list with the word Indexes of a batch.

emjotde:
Hm, I wanted to assign @kpu as a reviewer, but since you submitted in @pedrodiascoelho's stead it seems github doesn't let me do that. Do you have access to the reviewing interface at all?

emjotde:
@ykim362 Can you share the files you shared with NVIDIA with @snukky for regression tests? That's an FS model, right?

emjotde:
@pedrodiascoelho Thanks a lot for the comments; that will be helpful.

CHANGELOG.md
```diff
@@ -52,6 +52,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Changed compile flags -Ofast to -O3 and remove --ffinite-math
 - Moved old graph groups to depracated folder
 - Make cublas and cusparse handle inits lazy to save memory when unused
+- Factor groups and concatenation: doc/factors.md
```
kpu, line 55: Empirically I can start a review.

kpu (commented):
Apparently I can review my own pull request.

ykim362:
> @ykim362 Can you share the files you shared with NVIDIA with @snukky for regression tests? That's an FS model, right?

@emjotde sure. I will.

pedrodiascoelho:
> @pedrodiascoelho Thanks a lot for the comments; that will be helpful.

You're welcome @emjotde. I'll be glad and completely available to discuss any question or doubt regarding my implementation.

doc/factors.md
```diff
@@ -0,0 +1,192 @@
+# Using marian with factors
```
snukky, line 1: The documentation looks very good, it's clear and concise. Thanks.

I notice that it describes training, but for completeness, it would be nice to have basic one-line examples with training and decoding commands. Maybe in a separate short section at the end. People will surely ask about this.

snukky:
I got the pre-trained factored model, generated outputs with the current marian-dev master and tested it against this branch (marian-cef/master). Decoding with marian-decoder produces an error:
```
Error: Graph was reloaded and parameter 'F0::decoder_ff_logit_out_factor1_Wq' with type float32 (specified: false) is newly created
Error: Aborted from marian::Expr marian::ExpressionGraph::param(const string&, const marian::Shape&, marian::Ptr<marian::inits::NodeInitializer>&, marian::Type, bool, bool) in /home/romang/marian/marian-dev-public/src/graph/expression_graph.h:330
```
I don't use any fancy options during decoding, basically:
```
marian-decoder -m model.bin -v vocab.fsv vocab.fsv -b 1
```
@pedrodiascoelho Did you do some regression testing yourself? Have you seen this using your models?
@ykim32 or @emjotde can comment on the model architecture details if needed.

pedrodiascoelho:
> I got the pre-trained factored model, generated outputs with the current marian-dev master and tested it against this branch (marian-cef/master). Decoding with marian-decoder produces an error:
> 
> ```
> Error: Graph was reloaded and parameter 'F0::decoder_ff_logit_out_factor1_Wq' with type float32 (specified: false) is newly created
> Error: Aborted from marian::Expr marian::ExpressionGraph::param(const string&, const marian::Shape&, marian::Ptr<marian::inits::NodeInitializer>&, marian::Type, bool, bool) in /home/romang/marian/marian-dev-public/src/graph/expression_graph.h:330
> ```
> 
> I don't use any fancy options during decoding, basically:
> 
> ```
> marian-decoder -m model.bin -v vocab.fsv vocab.fsv -b 1
> ```
> 
> @pedrodiascoelho Did you do some regression testing yourself? Have you seen this using your models?
> @ykim32 or @emjotde can comment on the model architecture details if needed.

No I have not run any regression tests. I believe that the problem might be the fact that I found the need to change a small thing in the transformer architecture, more specifically in the size of some matrices in one of the attention layers, so the pre-trained model with the marian-dev code does not work. You can see the explanation of why I did that change here: https://github.com/marian-nmt/marian-dev/pull/748#discussion_r520678729. Although it is trowing an error in something on the logits so I'll take a look to it because it is probably not related to what I suggested

emjotde:
Don't think the dim change causes this. This seems like you broke back-compat with existing factored code by introducing a new required parameter `decoder_ff_logit_out_factor1_Wq`. That will be in the output layer, probably `layers/generic.{cpp,h}`.

Also please do run regression tests for everything else. There are no factored ones at the moment, but since there are other things going on here, we need to check.

emjotde:
Also, I need to take a better look, but it is not quite clear to me what this actually achieves. The title says "improvements", but what gets improved? Quality? Performance? Why is this being done?

Lacking these things this would be just added complexity, no?

pedrodiascoelho:
> Also, I need to take a better look, but it is not quite clear to me what this actually achieves. The title says "improvements", but what gets improved? Quality? Performance? Why is this being done?
> 
> Lacking these things this would be just added complexity, no?

I would say that the improvement is basically the implementation of concatenation embedding when using factors. Besides that, the factor code is exactly what was previously implemented. Besides concatenation, the main contribution is making the use of target factors more user-friendly by adding an option in the config, that allows to choose more clearly how we want to predict the factors.  

We decided to implement concatenation because we were attempting to apply Terminology Constraints by using factors. Something similar to this: https://www.aclweb.org/anthology/P19-1294.pdf. Since concatenation was not a part of marian, and that was what it was used in the paper we wondered if that could improve the system. We found that using concatenation actually resulted in a very slightly increase of the results.

pedrodiascoelho:
Here is list of the contributions:
- Implemented factor embeddings using concatenation;
- Added options to the CLI interface that allow the choice of:
        - The type of factor embedding (sum or concatenation);
	- The factor embedding dimension if concatenation is chosen;
	- The type of method to use to predict the factors if target factors are used.
- Documentation / User Guide regarding how to use factors;

emjotde:
OK, thanks. Do you have a paper or report on those experiments?

pedrodiascoelho:
> OK, thanks. Do you have a paper or report on those experiments?

Hello @emjotde sorry for the late response. I've shared with @snukky a table with the results of our experiences, and the comparison between summing and concatenating embeddings. A formal document with the results of the experiments will be ready by the end of next month as this work was done in the scope of my master thesis and I'm still writing it.

snukky:
FYI, I marked this PR as a draft because it breaks the compatibility with factored models trained without the proposed changes.

pedrodiascoelho:
I am writing this message to repoen the discussion in this PR after committing changes that I believe that solve the problem mentioned above.

> I got the pre-trained factored model, generated outputs with the current marian-dev master and tested it against this branch (marian-cef/master). Decoding with marian-decoder produces an error:
> 
> ```
> Error: Graph was reloaded and parameter 'F0::decoder_ff_logit_out_factor1_Wq' with type float32 (specified: false) is newly created
> Error: Aborted from marian::Expr marian::ExpressionGraph::param(const string&, const marian::Shape&, marian::Ptr<marian::inits::NodeInitializer>&, marian::Type, bool, bool) in /home/romang/marian/marian-dev-public/src/graph/expression_graph.h:330
> ```
> 
> I don't use any fancy options during decoding, basically:
> 
> ```
> marian-decoder -m model.bin -v vocab.fsv vocab.fsv -b 1
> ```
> 
> @pedrodiascoelho Did you do some regression testing yourself? Have you seen this using your models?
> @ykim32 or @emjotde can comment on the model architecture details if needed.

More info about how the problem was solved here (marian-cef/marian-dev#5).

snukky:
@pedrodiascoelho I run regression tests from `marian-regression-tests/tests/factors` and our test with an internal model using marian-cef/master, and both sets of tests fail. All tests work with the current marian-dev/master, so I guess something is still not working properly.

Have you been using `marian-regression-tests/tests/factors` to make sure that the proposed changes are backward compatible? Do those tests pass for you for the current marian-cef/master?

pedrodiascoelho:
Hello @snukky,

Could you give more information about the error? I Also tested both `marian-nmt/marian-dev` and `marian-cef/marian-dev`, and I got the same output with the code from both repos. I pass the training test (`tests/factors/test_factors.sh`) and fail the decoding test (`tests/factors/test_factors_decoder.sh`) only on this line: 
```
78c78
< in|s0 Haus|s1 besit|s1 zer|s0 fallenden|s1 enden|s0 Zentren|s0
---
> in|s0 den|s0 Heimat|s1 besit|s0 fallenden|s0 Zentren|s0
```

Code compiled with the following options: 
```
CC=/usr/bin/gcc-8 CXX=/usr/bin/g++-8 CUDAHOSTCXX=/usr/bin/g++8 cmake -DUSE_SENTENCEPIECE=ON -DUSE_FBGEMM=on     -DCOMPILE_CPU=on -DCOMPILE_TESTS=ON -DCOMPILE_EXAMPLES=ON -DCUDA_TOOLKIT_ROOT_DIR=/usr/local/cuda-10.1 ..
```
And ran on a Tesla T4 GPU

snukky:
Tests failed because several translated lines were different than expected ones. I think I used the same options, but will double check. Do you run it with marian-cef/master updated with marian-dev/master?

pedrodiascoelho:
Yes, tested with the current version of marian-cef/master and with marian-cef/master updated with the most recent marian-dev/master and got the same behaviour. It passes the training test and fails the decoding one in that same exact line reported above

kpu:
@snukky Currently deadlocked on reply from you?

snukky:
So I run the recent (a) marian-dev/master (v1.10.19; 1c8ee95) and (b) the marian-cef updated with that master (https://github.com/marian-nmt/marian-dev/tree/marian-cef-factors) again on (i) regression tests you added in marian-regression-test/tests/factors and (ii) 3 other tests that I have using an internal factored model.

Compilation commands:

    CC=/usr/bin/gcc-8 CXX=/usr/bin/g++-8 CUDAHOSTCXX=/usr/bin/g++8 cmake -DUSE_SENTENCEPIECE=ON -DUSE_FBGEMM=on     
		-DCOMPILE_CPU=on -DCOMPILE_TESTS=ON -DCOMPILE_EXAMPLES=ON -DCUDA_TOOLKIT_ROOT_DIR=/usr/local/cuda-10.1 
		-DCOMPILE_SERVER=on -DCOMPILE_MAXWELL=on ..

Regression tests from marian-regression-tests (i) work with both versions of Marian, but my tests (ii) work only with the recent master (a), not the marian-cef-factors branch. My tests (ii) include 3 translations and n-best list translations, and there are 17/100, 19/100, 332/1997 different lines, so it's above randomness that we might observe due to different GPUs, etc.

So it looks like the proposed changes still break backward compatibility at least for some model architectures. The model I use in (ii) was created with Marian v1.7.7 2019-05-01. It's a transformer with SSRU in the decoder and tied layers (transformer-decoder-autoreg: rnn, dec-cell: ssru, transformer-tied-layers: [1,1,1,1,1,1]). I think I can also share the config logs for more details about the architecture if that's not enough for you to debug.

pedrodiascoelho:
Hello @snukky, thank you for the feedback.
Could you provide the model tou tested with for me to try to reproduce your error?
If it is not possible, could you provide the config list of parameters used to train the model so that I can try to train a model myself with marian 1.7.7 and with the exact same parameters as yours?

snukky:
I don't have the training configs, but I should be able to share config logs with the architecture definition, let me check.

snukky:
After checking with @pedrodiascoelho, the older model was producing different output using Marian from this PR because now marian-decoder requires also `--lemma-dependency re-embedding`. Adding this option makes the tests pass successfully, so it wasn't related to model architecture or to a Marian version that that model has been trained with.

@pedrodiascoelho I think the change in how `--lemma-dim-emb` used to work and how it works now need to be described in the changelog, so that we will not face this backward incompatibility later again. Could you also update this branch with the newest changes from master?

pedrodiascoelho:
Hello @snukky 
Following our discussion I have added some logic to ensure backward compatibility with the older cli when choosing the lemma-dependency method, could you check my work?
Also updated the code with the most recent marian-dev

snukky:
@pedrodiascoelho I confirm that after this fix: https://github.com/marian-cef/marian-dev/pull/7, the factors code introduced in this PR is backward compatible, AFAICT.

doc/factors.md
```diff
@@ -0,0 +1,208 @@
...
+                                Options available: sum, concat
+```
+
+The dimension of the factor embeddings must be specified if using combine option `concat`. If using `sum`, the factor embedding dimension matches that of the lemmas.
+
+```
+--factors-dim-emb INT           Embedding dimension of the factors. Only used if concat is selected as factors combining form
+```
+
+Note: At the moment `concat` is only implemented for usage in the source side.
```
snukky, line 145: Just checking, is there a clear error if someone tries to use `concat` with target factors?

doc/factors.md
```diff
@@ -0,0 +1,208 @@
+# Using marian with factors
```
snukky, line 1: @graemenail Any suggestions how we should integrate this document into the new format of `doc/` and Sphinx-based docs? I'm thinking of creating a new section on the nav bar for user documentation and put it there.

doc/factors.md
```diff
@@ -0,0 +1,208 @@
...
+* `soft-transformer-layer`: Uses an additional transformer layer to predict the factors using the previously predicted lemma
+* `hard-transformer-layer`: Like `soft-transformer-layer` but with hard-max
+* `lemma-dependent-bias`: Adds a learned bias term based on the predicted lemma to the logits of the factors. There is no additional transformer layer introduced with this option
+* `re-embedding`: After predicting a lemma, re-embed the lemma and add this new vector before predicting the factors
+* `lemma-dim-emb`: Controls the dimension of the re-embedded lemma when using the option `re-embedding`
+
+
+### Weight tying
+
+If using factors only on the source or target side but using a joint vocabulary, there are two different options to tie source and target embedding weights:
```
snukky, line 167: I'm not sure if I got this sentence correctly. Do you mean "If factors are used only on one side (either source or target) with a joint vocabulary, there are two options for tying"?

doc/factors.md
```diff
@@ -0,0 +1,208 @@
...
+
+1. Use combine option `concat` (If using factors only on the source side).
+2. Use combine option `sum`, and create "dummy" factors for the side initially without factors. This entails creating a factored vocabulary where the same number of factors are present as are on the side with meaningful factors. In the previous example, if we have the capitalization and subword factors on the source side, the target side would have five different dummy factors (they can all be in the same group). In the *lemma* section of the `.fsv` file we would just not put `_has_X` for any lemma.
```
snukky, line 170: Would the term "the non-factorized side" be clearer?

doc/factors.md
```diff
@@ -0,0 +1,208 @@
...
+
+## Examples
+Some examples of possible commands to train factored models in marian:
+* Using factors on both source and target. Using `sum` to combine lemma and factor embeddings. No tied embeddings and no lemma dependency when predicting the factors:
+```
+path_to/build/marian -t corpus.fact.{src,trg} -v vocab.{src,trg}.fsv
+```
+* Using factors only on the source side. Using `concat` to combine lemma and factor embeddings. Source, target and output embeddings matrices tied:
+```
+path_to/build/marian -t corpus.fact.src corpus.trg -v vocab.src.fsv vocab.trg.yml --factors-combine concat --factors-dim-emb 8 --tied-embeddings-all
```
snukky, line 203: Consider spliting commands into multiple lines to avoid the slider on narrow screens. This will become more problematic in Sphinx docs, I think.

src/common/config.cpp
```diff
@@ -116,6 +116,21 @@ void Config::initialize(ConfigParser const& cp) {
     config_["tsv-fields"] = tsvFields;
   }
 
+  // ensures factors backward compability whilst keeping the more user friendly CLI
```
snukky, line 119: Heh, I had overlooked this misspelling when I was editing the code: `compability` -> `compatibility`.

src/data/factored_vocab.cpp
snukky: Please add this to the code as function documentation. Preferably use the doxygen syntax that we've established in the PR with documentation.

src/data/factored_vocab.cpp
```diff
@@ -655,6 +655,39 @@ std::string FactoredVocab::surfaceForm(const Words& sentence) const /*override f
...
+  lemmaIndices.reserve(words.size());
+  factorIndices.reserve(words.size() * getTotalFactorCount());
+
+  auto numGroups = getNumGroups();
+  std::vector<size_t> lemmaAndFactorIndices;
+
+  for (auto &word : words) {
+    if (vocab_.contains(word.toWordIndex())) {
+      word2factors(word, lemmaAndFactorIndices);
+      lemmaIndices.push_back((IndexType) lemmaAndFactorIndices[0]);
```
snukky, line 679: Consider adding a comment describing the logic/what is happening here in these for loops.

src/layers/embedding.cpp
```diff
@@ -8,19 +8,31 @@ Embedding::Embedding(Ptr<ExpressionGraph> graph, Ptr<Options> options)
...
+      /*fanIn=*/false, /*fanOut=*/true);  // -> embedding vectors have roughly unit length
+
   factoredVocab_ = FactoredVocab::tryCreateAndLoad(options_->get<std::string>("vocab", ""));
   if(factoredVocab_) {
     dimVoc = (int)factoredVocab_->factorVocabSize();
     LOG_ONCE(info, "[embedding] Factored embeddings enabled");
+    if(opt<std::string>("factorsCombine") == "concat") {
+      ABORT_IF(dimFactorEmb == 0,
+               "Embedding: If concatenation is chosen to combine the factor embeddings, a factor "
+               "embedding size should be specified.");
```
snukky, line 26: "must be specified"?

src/layers/embedding.cpp
```diff
@@ -32,6 +44,24 @@ Embedding::Embedding(Ptr<ExpressionGraph> graph, Ptr<Options> options)
...
+  auto lemmaEmbs = rows(E_, lemmaIndices);
+  int dimFactors = FactorEmbMatrix_->shape()[0];
+  auto factEmbs
+      = dot(graph->constant(
+                {(int)data.size(), dimFactors}, inits::fromVector(factorIndices), Type::float32),
+            FactorEmbMatrix_);
+
+  auto out = concatenate({lemmaEmbs, factEmbs}, -1);
+
+  return out;
```
snukky, line 62: Can this be a single statement?

src/layers/embedding.cpp
```diff
@@ -32,6 +44,24 @@ Embedding::Embedding(Ptr<ExpressionGraph> graph, Ptr<Options> options)
   E_ = graph_->param(name, {dimVoc, dimEmb}, initFunc, fixed);
 }
 
+//Embeds a sequence of words (given as indices), where they have factor information. The matrices are concatenated
```
snukky, line 47: Please add a whitespace after `//`.

src/layers/embedding.cpp
```diff
@@ -96,13 +128,13 @@ std::tuple<Expr /*embeddings*/, Expr /*mask*/> Embedding::apply(Ptr<data::SubBat
   //        more slowly
 
   auto batchEmbeddings = apply(subBatch->data(), {dimWidth, dimBatch, dimEmb});
-#if 1
+#if 0
+    // @TODO: this is dead code now, get rid of it
+    // experimental: hide inline-fix source tokens from cross attention
+    auto batchMask = graph->constant({dimWidth, dimBatch, 1},
+                                     inits::fromVector(subBatch->crossMaskWithInlineFixSourceSuppressed()));
+#else
```
snukky, line 136: Can we just delete this?

src/layers/embedding.h
```diff
@@ -12,8 +12,10 @@ class FactoredVocab;
 // EncoderDecoderLayerBase, which knows to pass on all required parameters from options.
 class Embedding : public LayerBase, public IEmbeddingLayer {
   Expr E_;
+  Expr FactorEmbMatrix_;
```
snukky, line 15: Please add an inline comment what is it/what is it for.

src/models/transformer.h
```diff
@@ -295,7 +295,8 @@ class Transformer : public EncoderOrDecoderBase {
       kh = cache_[prefix + "_keys"];                                                   // then return cached tensor
     }
     else {
-      auto Wk = graph_->param(prefix + "_Wk", {dimModel, dimModel}, inits::glorotUniform(true, true, depthScaling_ ? 1.f / sqrtf((float)depth_) : 1.f));
+      int dimKeys =  keys->shape()[-1];
```
snukky, line 298: Add a comment when this will be different from `dimModel`. Concat factors only?

src/models/transformer.h
```diff
@@ -309,7 +310,8 @@ class Transformer : public EncoderOrDecoderBase {
         && cache_[prefix + "_values"]->shape().elements() == values->shape().elements()) {
       vh = cache_[prefix + "_values"];
     } else {
-      auto Wv = graph_->param(prefix + "_Wv", {dimModel, dimModel}, inits::glorotUniform(true, true, depthScaling_ ? 1.f / sqrtf((float)depth_) : 1.f));
+      int dimValues = values->shape()[-1];
```
snukky, line 313: As above, please add a comment when this will be different from `dimModel`.

snukky (changes requested):
Thanks @pedrodiascoelho! Please take a look at my comments and feel free to resolve all conversations that you agree with and you think they don't need more our attention after you fix them.

doc/factors.md
```diff
@@ -0,0 +1,208 @@
+# Using marian with factors
+
+Following this README should allow the user to train a model with source- and/or target-side factors. To train with factors, the data must be formatted in a certain way. A special vocabulary file format is also required. See details below.
+
+### Requirements:
+
+Marian from this repository.
```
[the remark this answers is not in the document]
pedrodiascoelho: I see in the CHANGELOG (https://github.com/marian-nmt/marian/blob/master/CHANGELOG.md) that the first reference to the factor code is Marian 1.8.0. Although the entry says "Code for Factored Vocabulary, currently not usable yet without outside tools”,  and I am not sure which are this outside tools. Can I make as minimal dependency Marian 1.9 as it is the lowest version I’ve tested the code on and I know it works? What do you think about:
```
In order to use factors in Marian, you should use at least Marian 1.9.0 unless you want to use a factors functionality that requires setting one of the following command line options to their non default values: `--factors-combine`, `-—factors-dim-emb` and `--lemma-dependency` as they were only introduced after Marian 1.10.x.
```
Should I replace the x with a specific number?

doc/factors.md
```diff
@@ -0,0 +1,208 @@
+# Using marian with factors
+
+Following this README should allow the user to train a model with source- and/or target-side factors. To train with factors, the data must be formatted in a certain way. A special vocabulary file format is also required. See details below.
```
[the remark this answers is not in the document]
pedrodiascoelho: done

doc/factors.md
```diff
@@ -0,0 +1,208 @@
...
+
+Factors within a single group should start with the same string.
+
+For example, for a capitalization factor group, the individual factors could be:
+
+`c0`: all lowercase
+
+`c1`: first character capitalized, rest lowercase
+
+`c2`: all uppercase
```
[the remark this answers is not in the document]
pedrodiascoelho: done

doc/factors.md
```diff
@@ -0,0 +1,208 @@
...
+```
+trump|c1|s0 test|c0|s1 ed|c0|s0 positive|c0|s0 for|c0|s0 c|c2|s1 o|c2|s1 v|c2|s1 i|c2|s1 d|c2|s0 -|c0|s0 19|c0|s0 .|c0|s0
+```
+
+
+## Create the factored vocabulary
+
+Factored vocabularies should have the extension `.fsv`. How to structure the vocabulary file is described below. If using factors only on the source or target side, the vocabulary of the other side can be a normal `json`, `yaml`, etc.
+
+The `.fsv` vocabulary should have three sections:
```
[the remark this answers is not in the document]
pedrodiascoelho: The third section that is missing was the 'factor distribution section'. This was removed from the documentation because its not needed for now based on the current state of the code https://github.com/marian-nmt/marian-dev/blob/c228c77a527a30488e2279a90111c3f8906bc7eb/src/data/factored_vocab.cpp#L68-L72 
As you can see only checks are made and nothing is stored anywhere. Changed `three sections` to `two sections`.

doc/factors.md
```diff
@@ -0,0 +1,208 @@
...
+
+#### Other suggestions
+
+Certain characters are used by the `.fsv` vocabulary that will have to be escaped/replaced in the data: `#:_\|`
+
+The tokens in the factor vocabularies (`c0`, `c1`, `s0`, etc.) cannot be present in any of the *lemmas*.
+
+### Full `.fsv` file
+
+Putting everything together, the final `.fsv` file should look like this. It can have comments.
```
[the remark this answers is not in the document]
pedrodiascoelho: done

doc/factors.md
pedrodiascoelho: No there is not. Should I add one to the code?

doc/factors.md
pedrodiascoelho: Exactly. We added this section because it might not be obvious that if using factors only on one side (source or target) but using a joint vocabulary for the lemmas, you could still tie the embeddings.
If you use factors on both sides and they are the same you can also tied the embeddings of course. Added this to the text to remove any doubt. Also rephrased for your suggestion.

doc/factors.md
pedrodiascoelho: changed

doc/factors.md
pedrodiascoelho: done

src/common/config.cpp
pedrodiascoelho: done

src/data/factored_vocab.cpp
pedrodiascoelho: done

src/layers/embedding.cpp
pedrodiascoelho: done

src/layers/embedding.cpp
pedrodiascoelho: done

src/layers/embedding.cpp
pedrodiascoelho: done

src/layers/embedding.cpp
pedrodiascoelho: Yeah of course, this code is outdated as noticed. I've deleted. Do you want me to delete other places where there are unused code related to this? Namely the definition of function crossMaskWithInlineFixSourceSuppressed https://github.com/marian-nmt/marian-dev/blob/fe74576dc3e957f97b4842951fb6ee7ec52fa324/src/data/corpus_base.cpp#L564-L610
and
https://github.com/marian-nmt/marian-dev/blob/fe74576dc3e957f97b4842951fb6ee7ec52fa324/src/data/factored_vocab.h#L82-L89

src/layers/embedding.h
pedrodiascoelho: done

src/models/transformer.h
pedrodiascoelho: done. Yes because when using concat factors the output of the encoder will be dimModel + dimFactEmb

src/models/transformer.h
pedrodiascoelho: done

graemenail (changes requested):
Hi @pedrodiascoelho; I've left some comments - ping me if something isn't clear.

Also, I added small comment on marian-nmt/marian-regression-tests#78

doc/factors.md
pedrodiascoelho: Even being a older message, just replying here that this was already added to the [documentation](https://github.com/marian-cef/marian-dev/blob/master/doc/factors.md#examples), to mark it as done once you are closing the discussion threads.

doc/factors.md
```diff
@@ -0,0 +1,218 @@
...
+
+* `c2`: all uppercase
+
+If there were a second factor group for subword divisions, the individual factors could be:
+
+* `s0`: end of word, whitespace should follow
+
+* `s1`: join token with next subword
+
+There is no limit on the number of factor groups barring some practical limitations having to do with how the vocabulary is stored by `marian`. If the limit is exceeded `marian` will throw this error.
```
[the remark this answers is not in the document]
pedrodiascoelho: It's an error related to the fact that we have to many indexes for the `WordIndex` variable, that stores the indexes of the vocabulary. Do you think its better to explicitly say what is the error? Or change `this` for `an` is enough?

doc/factors.md
```diff
@@ -0,0 +1,218 @@
...
+```
+
+Preprocessed sentence:
+```
+trump test@@ ed positive for c@@ o@@ v@@ i@@ d - 19 .
+```
+
+Apply factors:
+```
+trump|c1|s0 test|c0|s1 ed|c0|s0 positive|c0|s0 for|c0|s0 c|c2|s1 o|c2|s1 v|c2|s1 i|c2|s1 d|c2|s0 -|c0|s0 19|c0|s0 .|c0|s0
```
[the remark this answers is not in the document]
pedrodiascoelho: No, not that I'm aware of

doc/factors.md
```diff
@@ -0,0 +1,218 @@
...
+    , : _lemma _has_c _has_s
+    . : _lemma _has_c _has_s
+    the : _lemma _has_c _has_s
+    for: _lemma _has_c _has_s
+    ```
+
+
+#### Other suggestions
+
+Certain characters are used by the `.fsv` vocabulary that will have to be escaped/replaced in the data: `#:_\|`
```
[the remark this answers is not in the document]
pedrodiascoelho: Yeah actually I also never saw this character having a critical part on the vocabulary like the other 4. We suggest to escape it because that is what we were told in one of the early meetings we had with who implemented the code in the first place. I escape the characters by replacing those special characters in the corpus before creating the vocabulary. You could do it with this `sed` line for example: 
```bash
cat corpus | sed -e 's/#/\&htg;/g' -e 's/:/\&cln;/g' -e 's/_/\&usc;/g' -e 's/|/\&ppe;/g' -e 's/\\/\&esc;/g' > corpus.esc
```

src/layers/output.cpp
```diff
@@ -44,8 +44,12 @@ namespace mlp {
...
 
   /*const*/ int lemmaDimEmb = options_->get<int>("lemma-dim-emb", 0);
+  std::string lemmaDependency = options_->get<std::string>("lemma-dependency", "");
   ABORT_IF(lemmaDimEmb && !factoredVocab_, "--lemma-dim-emb requires a factored vocabulary");
-  if(lemmaDimEmb > 0) {  // > 0 means to embed the (expected) word with a different embedding matrix
+  if(lemmaDependency == "re-embedding") {  // embed the (expected) word with a different embedding matrix
+    ABORT_IF(
+        lemmaDimEmb <= 0,
+        "In order to predict factors by re-embedding them, a lemma-dim-emb must be specified.");
 #define HARDMAX_HACK
```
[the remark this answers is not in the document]
pedrodiascoelho: This was done for what I believe was quick experimentation by who wrote initially this code. Basically if you used a odd number in `--lemma-dim-emb` you could easily change between what is now the options `re-embedding` and `hard-transformer-layer` set through `--lemma-dependency`. Actually I think I can remove this part of the code as it is so experimental. What do you think? I'm suggesting removing: 
https://github.com/marian-nmt/marian-dev/blob/9fa166be885b025711f27b35453e0f2c00c9933e/src/layers/output.cpp#L49-L52
and
https://github.com/marian-nmt/marian-dev/blob/9fa166be885b025711f27b35453e0f2c00c9933e/src/layers/output.cpp#L246-L255

doc/factors.md
```diff
@@ -0,0 +1,218 @@
...
+* `soft-transformer-layer`: Uses an additional transformer layer to predict the factors using the previously predicted lemma
+* `hard-transformer-layer`: Like `soft-transformer-layer` but with hard-max
+* `lemma-dependent-bias`: Adds a learned bias term based on the predicted lemma to the logits of the factors. There is no additional transformer layer introduced with this option
+* `re-embedding`: After predicting a lemma, re-embed the lemma and add this new vector before predicting the factors
+* `lemma-dim-emb`: Controls the dimension of the re-embedded lemma when using the option `re-embedding`
+
+
+### Weight tying
+
+If you use factors both on the source and target side, and the factors are the same for both sides you can tie the embeddings exactly as you do for non factored models.
```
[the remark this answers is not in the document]
pedrodiascoelho: no they are not checked as being the same. Only the sizes of the matrices are checked to be the same. But the same thing happens for the lemmas. You can tie the embeddings if you have two  different vocabularies with the same size, even though that would not make much sense to do so, there is nothing in the code that forbids you to do it

doc/factors.md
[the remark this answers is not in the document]
graemenail: If the error message makes it obvious how to fix the problem, I'd include it. If not, replacing it with `an` is sufficient.

doc/factors.md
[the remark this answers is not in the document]
graemenail: I guess `\` could interpreted as some base64-encoded string.

It seems that `<->` is a special string too; source code gives an example of: `_x <-> _has_x`.

src/layers/output.cpp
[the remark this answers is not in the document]
graemenail: If the `hard-transformer-layer` implementation is the same, it can be removed. Though, in order to preserve the old API, when fed an odd integer via`--lemma-dim-emb` (without specifying `--lemma-dependency`), it should select `hard-transformer-layer` with `lemma-dim-emb` adjusted to the desired even number. This should be accompanied by a warning stating this behaviour and that the user _really_ wants an odd number they should also specify `--lemma-dependency`.

doc/factors.md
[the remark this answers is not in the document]
snukky: Yes, better to state a higher version than too low, so 1.9 is fine. As for '1.10.x', I would put '1.10.20+' for now and update it later.

doc/factors.md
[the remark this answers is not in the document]
snukky: Thanks. Please change also "should have the extension" to "must have the extension" as it is strictly required.

doc/factors.md
```diff
@@ -0,0 +1,218 @@
...
+    </s> : _lemma
+    <unk> : _lemma
+    , : _lemma _has_c _has_s
+    . : _lemma _has_c _has_s
+    the : _lemma _has_c _has_s
+    for: _lemma _has_c _has_s
+    ```
+
+
+#### Other suggestions
```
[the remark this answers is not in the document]
snukky: +1

doc/factors.md
snukky: Yes, please

src/layers/embedding.cpp
snukky: I would say yes, go ahead and remove. The cleaner this code with factors is, the better. Thanks

doc/factors.md
```diff
@@ -0,0 +1,218 @@
...
+    These are the vocabulary entries themselves. They have the format of `LEMMA : _lemma [_has_c] [_has_s]`. The `_has_X` should only apply to lemmas that can have an `X` factor anywhere in the data (which will likely be all of the tokens except `</s>` and `<unk>`).
+
+    Examples:
+    ```
+    </s> : _lemma
+    <unk> : _lemma
+    , : _lemma _has_c _has_s
+    . : _lemma _has_c _has_s
+    the : _lemma _has_c _has_s
+    for: _lemma _has_c _has_s
```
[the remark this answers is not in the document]
pedrodiascoelho: White space added

doc/factors.md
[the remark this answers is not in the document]
pedrodiascoelho: You are right. Changed

doc/factors.md
pedrodiascoelho: Explicitly error added to the code. If `--factors-combine concat` is set and target factors are used the following error will be thrown: "Combining lemma and factors embeddings with concatenation on the target side is currently not supported"

src/data/factored_vocab.cpp
pedrodiascoelho: comments added

src/data/factored_vocab.h
```diff
@@ -48,12 +48,13 @@ class FactoredVocab : public IVocab {
   size_t virtualVocabSize() const { return factorShape_.elements<size_t>(); } // valid WordIndex range (representing all factor combinations including gaps); virtual and huge
 
   CSRData csr_rows(const Words& words) const; // sparse matrix for summing up factors from the concatenated embedding matrix for each word
-
+  void lemmaAndFactorsIndexes(const Words& words, std::vector<IndexType>& lemmaIndices, std::vector<float>& factorIndices) const;
 #ifdef FACTOR_FULL_EXPANSION
   const CSRData& getGlobalFactorMatrix() const { return globalFactorMatrix_; }   // [v,u] (sparse) -> =1 if u is factor of v  --only used in getLogits()
 #endif
   size_t getNumGroups() const { return groupRanges_.size(); }
   std::pair<size_t, size_t> getGroupRange(size_t g)    const { return groupRanges_[g]; }   // [g] -> (u_begin, u_end)
```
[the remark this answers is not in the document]
pedrodiascoelho: changed

doc/factors.md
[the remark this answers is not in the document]
pedrodiascoelho: The message is "Too many factors, virtual index space 24,945,901,568 exceeds the bit limit of WordIndex type". So that I change the `the` for `an`

doc/factors.md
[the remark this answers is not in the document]
pedrodiascoelho: As you can see here https://github.com/marian-nmt/marian-dev/pull/748#discussion_r638153012 That characters are no longer used in the code thus not needed to be escaped. Also since it is a line that starts with _ it already enters the special line treatment in the factored vocab code. So it should not be a problem if present in the lemma vocabulary

doc/factors.md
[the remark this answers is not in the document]
pedrodiascoelho: ok thanks. Updated

doc/factors.md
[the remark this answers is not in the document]
graemenail: If it's required not to break older vocabs, could you update the comment in the code to reflect the that it's no longer used.

src/layers/output.cpp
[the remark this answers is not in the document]
pedrodiascoelho: I see your point and agree with you that the old API should be preserved at all costs to guarantee backwards compatibility. Although for this particular case I think it could be deleted. Here
https://github.com/marian-nmt/marian-dev/blob/9fa166be885b025711f27b35453e0f2c00c9933e/src/layers/output.cpp#L246-L248
You can see that that is a message saying that this was done for quick experimentation. And also, choosing hard max be setting a odd value in `lemma-dim-emb` does not seem a clean solution to me. As @snukky said here https://github.com/marian-nmt/marian-dev/pull/748#discussion_r643878326 we should try to delete this experimental code, and I do think that in this case there's no need to guarantee backwards compatibility for such an experimental scenario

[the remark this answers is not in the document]
graemenail: If the plan is to not support this experimental mode, then it can be deleted with no associated changes

src/layers/embedding.cpp
pedrodiascoelho: done

src/layers/output.cpp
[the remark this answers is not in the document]
pedrodiascoelho: Can I remove this section @snukky? Can we delete this experimental mode?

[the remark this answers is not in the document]
snukky: I think so. Let's remove this and I will re-run internal regression tests again before merging this PR to make sure it was only an experimental code.

doc/factors.md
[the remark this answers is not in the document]
pedrodiascoelho: No, you can safely omit the `_x <-> _has_x` form the vocabulary file.

src/layers/output.cpp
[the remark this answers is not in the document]
pedrodiascoelho: Thanks to both for the inputs and I also agree that deleting and rerunning regressions tests is the best approach. I've deleted the code.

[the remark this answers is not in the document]
snukky: Internal regression tests seem to work, I will re-run all regression tests now. Thanks!

snukky:
All existing regression tests pass with https://github.com/marian-cef/marian-dev/commit/cb44bbc435afdd65619bb1cba317c3748983ba5f, which is new factors + the most recent Marian master. There is one new test that fails, but this is tracked in https://github.com/marian-nmt/marian-regression-tests/pull/78#issuecomment-863839190

snukky (approved):

graemenail (approved):

pedrodiascoelho:
Hello @snukky! Just to ping you regarding this PR as you asked me to do during the EU marian project meeting. I've updated the test that was failing in https://github.com/marian-nmt/marian-regression-tests/pull/78#issuecomment-864157181. Let me know if something else is missing and thank you again for the help and review in this.

snukky:
Yes, thanks for the reminder. I updated your branch and tested it again with an internal model last week. All looks good to me. @emjotde wanted to take a look at the changes too as the factors code is critical for our models, so now we're waiting for his approval. I will keep reminding him.

kpu:
> @emjotde wanted to take a look at the changes too as the factors code is critical for our models, so now we're waiting for his approval. I will keep reminding him.

https://www.youtube.com/watch?v=Hr_9b-Lt-pk

pedrodiascoelho:
Hello @snukky. As requested, I'll leave here a summarized version of my Master Thesis where you can check the results from the experiments that compare the concatenation vs sum (section 3.3).
[Extended_Abstract_Pedro_Dias_Coelho_83719.pdf](https://github.com/marian-nmt/marian-dev/files/6741309/Extended_Abstract_Pedro_Dias_Coelho_83719.pdf)
Let me know if you need any further explanation or clarification regarding this.

PR #749, opened 2020-11-02
Add Triton Marian backend running in AzureML Inference Environment

delong-coder:
### Description

Adding Triton Mairan backend 2.0 according to [Triton Backend Shared Library](https://docs.nvidia.com/deeplearning/triton-inference-server/user-guide/docs/backend.html#backend-shared-library). We can get more flexibility and a few performance gains from the new version backend.

This backend running in the triton server as a custom backend, with the AzureML inference image environment. AzureML image doesn't have all the necessary share library dependencies, so the backend needs static libraries of Marian and its dependencies. In other words, the backend is compiled with the static library of Marian on a specific version, and It should be compiled individually.

List of changes:
- marian_backend: New Triton Marian backend source code.
- src: Some changed Marian code and CMakeLists.txt.
- Dockerfile: This Dockerfile is used for compiling the new backend with the static library of Marian on a specific version.
- build.sh: A simple shell script to run the Dockerfile and get the generated libtriton_marian.so. 

Added dependencies: none

### How to test

Run ./build.sh to build the Triton Marian backend. 
Run the backend with the mcr.microsoft.com/azureml/aml-triton base image in the Azure Machine Learning Inference stack.

### Checklist

- [X] I have run regression tests


emjotde:
Hi @delong-coder 
Thank you for that. Since this doesn't touch any of the original code, I do not really have many remarks here. A couple of things to decide:

* @snukky Should this rather go to `contrib/triton-aml`
* should we try to build this by default to make sure that it stays functional? (If yes, then we need to spend some more time on reviewing this).

ykim362:
+1 for `contrib/triton-aml` and readme.

snukky:
> Should this rather go to contrib/triton-aml than triton-aml?

I think `contrib/triton-aml` will be a better place too, so yes.

> should we try to build this by default to make sure that it stays functional?

We can keep it disabled by default and keep testing it by adding to existing or new checks in our GitHub and Azure pipelines.

delong-coder:
Thanks for your feedback.
Update:
- move the backend to contrib/triton-aml.
- add a README.md to describe what this is.

emjotde (approved):
Since this is isolated in `contrib` I am just pulling this in. You will be responsible for keeping this alive. If there is desire to have that build with our continuous integration then we will have to take a closer look at some point.

contrib/triton-aml/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+For the AzureML Inference team members, you can put it into the following place of *aml-triton* base image:
+
+- <backend_directory>/marian/libtriton_marian.so
+
+Where <backend_directory> is by default /opt/tritonserver/backends.
+
+## Make changes
+
+If you want to compile with another version of Marian, you need to replace `RUN git checkout youki/quantize-embedding` in the Dockerfile, then copy the new CMakeLists.txt replace the old one, add src/cmarian.cpp into CMakeLists.txt and make some changes to make sure it will build a static library of Marian.
```
snukky, line 32: @delong-coder Does this work only with `youki/quantize-embedding`? Is there a reason why we cannot try to merge that branch into master? Maybe @ykim362 is in a better position to comment.

contrib/triton-aml/Dockerfile
```diff
@@ -0,0 +1,96 @@
...
+RUN apt-get update --fix-missing
+RUN apt-get install -y curl git autoconf automake libtool curl make g++ unzip cmake build-essential cpio
+RUN apt-get -y clean && \
+    rm -rf /var/lib/apt/lists/*
+
+# install zlib
+WORKDIR /
+RUN git clone --no-checkout https://github.com/madler/zlib
+WORKDIR /zlib
+RUN git checkout tags/v1.2.10 && \
```
snukky, line 12: We have zlib in `src/3rd_party/zlib`. It needs to be installed for getting a specific version or because protobuf requires it?

contrib/triton-aml/README.md
delong-coder: > @delong-coder Does this work only with `youki/quantize-embedding`? Is there a reason why we cannot try to merge that branch into master? Maybe @ykim362 is in a better position to comment.

No, it can work with the master by doing a little change. We built this backend for the NLX-Sequence-Modeling of the Office team, they used `youki/quantize-embedding` branch, so we did the same thing. We will consider compiling this with the master of Marian.

snukky: Great, thanks for the clarification.

contrib/triton-aml/Dockerfile
delong-coder: > We have zlib in `src/3rd_party/zlib`. It needs to be installed for getting a specific version or because protobuf requires it?

I think it's for getting a specific version that works well for the Office team. We got this from their original Dockerfile.

PR #690, opened 2020-07-31
Quantized model finetuning 

afaji:
### Description
This PR implements training the model in the simulated quantized model. This fine-tuning scheme is useful for restore BLEU drop on a quantized model.

List of changes:
- enables fixed-point and log-quantized model training

see the comments on each file for more detail

Added dependencies: none

### How to test

- Train a normal Marian model: `model.npz`
- Finetune the trained model by adding these extra commands: `--compress-bit 8 --compress-skip-bias`
Since fine-tuning is fast, it is desirable to reduce the validation step for more frequent checking.

### Checklist

- [x] I have tested the code manually
- [x] I have run regression tests
- [x] I have read and followed CONTRIBUTING.md
- [x] I have updated CHANGELOG.md

PR #751, opened 2020-11-04
Merge quantization code with fp16 branch

emjotde:
### Description
This is meant to isolate the changes from PR https://github.com/marian-nmt/marian-dev/pull/690 (Quantized finetuning) into the fp16 training branch. Seems it might not have been too difficult.

@snukky can you work with @afaji and @kpu to verify this works? All the changes are basically your own changes, I only adapted `graph_group_sync.cpp`. Merge went through relatively smoothly.


snukky:
This branch is not already with `--clip-norm 0`?

The quantized finetuning also needs hotfixes from #759 and #761, so I will pull them into this PR.

emjotde:
It's not a PR into master, so does it matter?

snukky:
The regression tests we already have for the quantized finetuning have been generated with the default `clip-norm 1` and the two hotfixes. The tests fail, so I wanted to make sure first what settings this branch uses.

emjotde:
Ah, we should get rid of `--clip-norm x` then.

emjotde:
@afaji @ykim362  Are you ever using `--clip-gemm x` in your quantized training? If not we can probably remove that? That was introduced by me and @kpu back in 2018 for WNGT, but probably never used again. Just clutter?

ykim362:
@emjotde I don't use it. I am fine with removing it.

kpu:
`--clip-gemm` has effectively been replaced by fitting to each tensor then @afaji's continued training

emjotde:
Cool, happy to get rid of that soonish.

snukky:
After applying the two Alham's hotfixes that are already in the master branch, the tests for quantized finetuning work. One test failed because this branch does not support continued training with `--after-batches` (i.e. the second run finished immediately due to `after: 20u,100u` in the config), but this is an irrelevant issue to this PR.

snukky (approved):

snukky:
fp16.4 is in the internal repo, so this can be closed. Please re-open if something is left here.

PR #761, opened 2020-11-11
fix the allocation size for opt-step Tensor

afaji:
### Description

A fix to pass the regression test https://github.com/marian-nmt/marian-regression-tests/pull/69

Issue: weird behaviour when opt-step is used and the model is trained with no shared vocabulary.
The reason is the temporary Tensor used for quantization was too small.


snukky:
> A fix to pass the regression test

I'd say it's a fix for an issue revealed by the regression tests :) I confirm the tests work now on all tested machines.

snukky (approved):

PR #762, opened 2020-11-12
Intgemm refactor

emjotde:
Hi, 
this is a first draft of my small refactoring changes. All in all it was easier than thought. 

* Similar to FBGEMM the compute type is now determined by the Element type in the model.bin file.
* If that's only intgemm{8,16} it will do the hardware-specific stuff on the fly.
* If intgemm8avx2 etc. it is already pre-packed and does no conversion, but is hardware specific. This should be possible to mmap.
* Got rid of the code in backends (for now). Open to add it back later.

TODO:
* ~~Verify I did not break FBGEMM back-compat with the new types.~~ (regression tests for FBGEMM pass, @ykim362 you may want to do some more testing)
* ~~Test out mmapping~~ (seems to work out-of-the-box, see below)
* Take a look at the TODOs I mentioned, especially concerning aborting with wrong hardware during conversion.
* It seems sse2 or ssse3 models do not work for me on avx2, please help verify/debug. AVX2 on AVX2 hardware works fine.
* Verify AVX512 works on AVX512 hardware. 

### Checklist

### DIFF
--- a/src/graph/expression_operators.cpp
+++ b/src/graph/expression_operators.cpp
@@ -474,14 +474,16 @@ Expr dot(Expr a, Expr b, bool transA, bool transB, float scale) {
   // Currently only true when command line options
   // --optimize --cpu-thread=N with N > 0 are set.
   if(device == DeviceType::cpu) {
-    if(isFloat(aElementType) && (isFloat(bElementType) || isIntgemm(bElementType))) {
-      if(a->graph()->getBackend()->isInt8() || matchType<intgemm8>(bElementType)) {
+    if(isFloat(aElementType) && isFloat(bElementType)) {
+      return Expression<DotNodeOp>(clip(a, clipValue), clip(b, clipValue), transA, transB, scale);
+    } else if(isFloat(aElementType) && isIntgemm(bElementType)) {
+      // @TODO: this branch should move into cpu::integer::*
+      if(sizeOf(bElementType) == 1) {
         return cpu::integer::dot<Type::int8>(a, b, transA, transB, scale);
-      } else if(a->graph()->getBackend()->isInt16() || matchType<intgemm16>(bElementType)) {
+      } else if(sizeOf(bElementType) == 2) {
         return cpu::integer::dot<Type::int16>(a, b, transA, transB, scale);
       } else {
-        return Expression<DotNodeOp>(
-          clip(a, clipValue), clip(b, clipValue), transA, transB, scale);
+        ABORT("Wrong size for Intgemm type {}??", sizeOf(bElementType));
       }
     } else if(isFloat(aElementType) && isPacked(bElementType)) {
 #if USE_FBGEMM
@@ -546,13 +548,16 @@ Expr affine(Expr a, Expr b, Expr bias, bool transA, bool transB, float scale) {
   Type bElementType = b->value_type();
 
   if(device == DeviceType::cpu) {
-    if(isFloat(aElementType) && (isFloat(bElementType) || isIntgemm(bElementType))) {
-      if(a->graph()->getBackend()->isInt8()  || matchType<intgemm8>(bElementType) ) {
+    if(isFloat(aElementType) && isFloat(bElementType)) {
+      return affineDefault(a, b, bias, transA, transB, scale);
+    } else if(isFloat(aElementType) && isIntgemm(bElementType)) {
+      // @TODO: this branch should move into cpu::integer::*
+      if(sizeOf(bElementType) == 1) {
         return cpu::integer::affine<Type::int8>(a, b, bias, transA, transB, scale, clipValue);
-      } else if(a->graph()->getBackend()->isInt16()  || matchType<intgemm16>(bElementType) ) {
+      } else if(sizeOf(bElementType) == 2) {
         return cpu::integer::affine<Type::int16>(a, b, bias, transA, transB, scale, clipValue);
       } else {
-        return affineDefault(a, b, bias, transA, transB, scale);
+        ABORT("Wrong size for Intgemm type {}??", sizeOf(bElementType));
       }
     } else if(isFloat(aElementType) && isPacked(bElementType)) {
 #if USE_FBGEMM
--- a/src/graph/expression_operators.cpp
+++ b/src/graph/expression_operators.cpp
@@ -477,14 +477,7 @@ Expr dot(Expr a, Expr b, bool transA, bool transB, float scale) {
     if(isFloat(aElementType) && isFloat(bElementType)) {
       return Expression<DotNodeOp>(clip(a, clipValue), clip(b, clipValue), transA, transB, scale);
     } else if(isFloat(aElementType) && isIntgemm(bElementType)) {
-      // @TODO: this branch should move into cpu::integer::*
-      if(sizeOf(bElementType) == 1) {
-        return cpu::integer::dot<Type::int8>(a, b, transA, transB, scale);
-      } else if(sizeOf(bElementType) == 2) {
-        return cpu::integer::dot<Type::int16>(a, b, transA, transB, scale);
-      } else {
-        ABORT("Wrong size for Intgemm type {}??", sizeOf(bElementType));
-      }
+      return cpu::integer::affineOrDot(a, b, nullptr, transA, transB, scale);
     } else if(isFloat(aElementType) && isPacked(bElementType)) {
 #if USE_FBGEMM
       // 07/10/2019 - Use packed GEMM only if the cpu architecture supports AVX2
@@ -543,7 +536,7 @@ static Expr affineDefault(Expr a, Expr b, Expr bias, bool transA, bool transB, f
 Expr affine(Expr a, Expr b, Expr bias, bool transA, bool transB, float scale) {
   auto device = a->graph()->getDeviceId().type;
 
-  float clipValue = a->graph()->getBackend()->getClip();
+  float clipValue = a->graph()->getBackend()->getClip(); // @TODO: do we use that at all?
   Type aElementType = a->value_type();
   Type bElementType = b->value_type();
 
@@ -551,14 +544,7 @@ Expr affine(Expr a, Expr b, Expr bias, bool transA, bool transB, float scale) {
     if(isFloat(aElementType) && isFloat(bElementType)) {
       return affineDefault(a, b, bias, transA, transB, scale);
     } else if(isFloat(aElementType) && isIntgemm(bElementType)) {
-      // @TODO: this branch should move into cpu::integer::*
-      if(sizeOf(bElementType) == 1) {
-        return cpu::integer::affine<Type::int8>(a, b, bias, transA, transB, scale, clipValue);
-      } else if(sizeOf(bElementType) == 2) {
-        return cpu::integer::affine<Type::int16>(a, b, bias, transA, transB, scale, clipValue);
-      } else {
-        ABORT("Wrong size for Intgemm type {}??", sizeOf(bElementType));
-      }
+      return cpu::integer::affineOrDot(a, b, bias, transA, transB, scale);
     } else if(isFloat(aElementType) && isPacked(bElementType)) {
 #if USE_FBGEMM
       // 07/10/2019 - Use packed GEMM only if the cpu architecture supports AVX2
--- a/src/layers/generic.cpp
+++ b/src/layers/generic.cpp
@@ -7,7 +7,6 @@
 #include "rnn/types.h"     // for State::select()
 #include "models/states.h" // for EncoderState
 #include "layers/lsh.h"
-#include "tensors/cpu/intgemm_interface.h"
 
 namespace marian {
   Logits::Logits(Expr logits) : Logits(New<RationalLoss>(logits, nullptr)) {} // single-output constructor from Expr only (RationalLoss has no count)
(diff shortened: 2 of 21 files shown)


emjotde:
* Mmapping just works, how nice :) At least for intgemm{8,16}avx2. 
* Can be enabled for testing here: https://github.com/marian-nmt/marian-dev/blob/5990142047faaa50850e0e3fc5409de4fd69cd14/src/translator/translator.h#L21 (just set the define to 1, and you should see a message like: `Memory mapping model at 0x7f2fcae1f000`).

emjotde:
@snukky with this change the following regression test fails:
```
tests/models/wnmt18/test_student_small_aan_optimize.sh
```
This is expected as I removed the `--optimize` option. The correct way to adapt this test would be to run `marian-conv -f model.npz -t model.bin --gemm-type intgemm16` and then point `marian-decoder` to the `model.bin` file.

This could also be extended to more regression tests, at least for `--gemm-type intgemm8`.

XapaJIaMnu:
Hey,

It looks good. Glad MMAP'ing works. I would like to have the shifted version exposed (as a command line switch or something), as it is noticeably faster on VNNI.

As for SSSE/SSE not working on avx2, this is because the multiply here: https://github.com/marian-nmt/marian-dev/blob/intgemm_refactor/src/tensors/cpu/intgemm_interface.h#L407 is width dependent. Potentially we might want to make it read in the type.

Cheers,

Nick

src/tensors/cpu/backend.h
```diff
@@ -10,25 +10,11 @@ namespace marian {
 namespace cpu {
 
 class Backend : public marian::Backend {
-protected:
-  bool int16_{false};
-  bool int8_{false};
-  bool int8shift_{false};
```
XapaJIaMnu, line 16: Any plans for alternative passing of the "shifted" parameter?

src/layers/generic.cpp
```diff
@@ -283,31 +282,7 @@ namespace marian {
       };
 
       if (shortlist_ && !cachedShortWt_) { // shortlisted versions of parameters are cached within one batch, then clear()ed
```
XapaJIaMnu, line 284: How difficult is to have the shortlist_ be passed to the Affine routine, then we can do it on the fly without extra cached nodes?

src/graph/expression_operators.cpp
```diff
@@ -546,13 +548,16 @@ Expr affine(Expr a, Expr b, Expr bias, bool transA, bool transB, float scale) {
...
-    if(isFloat(aElementType) && (isFloat(bElementType) || isIntgemm(bElementType))) {
-      if(a->graph()->getBackend()->isInt8()  || matchType<intgemm8>(bElementType) ) {
+    if(isFloat(aElementType) && isFloat(bElementType)) {
+      return affineDefault(a, b, bias, transA, transB, scale);
+    } else if(isFloat(aElementType) && isIntgemm(bElementType)) {
+      // @TODO: this branch should move into cpu::integer::*
+      if(sizeOf(bElementType) == 1) {
         return cpu::integer::affine<Type::int8>(a, b, bias, transA, transB, scale, clipValue);
-      } else if(a->graph()->getBackend()->isInt16()  || matchType<intgemm16>(bElementType) ) {
+      } else if(sizeOf(bElementType) == 2) {
```
XapaJIaMnu, line 557: I don't understand this, why not use matchtype? It's not obvious what `sizeOf(bElementType) == 2` means

src/graph/expression_operators.cpp
```diff
@@ -474,14 +474,16 @@ Expr dot(Expr a, Expr b, bool transA, bool transB, float scale) {
...
-    if(isFloat(aElementType) && (isFloat(bElementType) || isIntgemm(bElementType))) {
-      if(a->graph()->getBackend()->isInt8() || matchType<intgemm8>(bElementType)) {
+    if(isFloat(aElementType) && isFloat(bElementType)) {
+      return Expression<DotNodeOp>(clip(a, clipValue), clip(b, clipValue), transA, transB, scale);
+    } else if(isFloat(aElementType) && isIntgemm(bElementType)) {
+      // @TODO: this branch should move into cpu::integer::*
+      if(sizeOf(bElementType) == 1) {
         return cpu::integer::dot<Type::int8>(a, b, transA, transB, scale);
-      } else if(a->graph()->getBackend()->isInt16() || matchType<intgemm16>(bElementType)) {
+      } else if(sizeOf(bElementType) == 2) {
```
XapaJIaMnu, line 483: Again, what does `sizeOf(bElementType) == 2)` and can we use `matchType` or something else more obvious instead?

XapaJIaMnu (approved):
Looks good, biggest issue is that the meaning of `sizeOf(bElementType) == 2)` is unclear.

snukky:
I will refresh regression tests and check with this branch later today.

emjotde:
@XapaJIaMnu `sizeOf(...)` is just our equivalent for C++ `sizeof` that works with the `Type` enum. As you know the types are defined by their type classes and their size. 

* intgemm16 has `isIntgemm: true`, `size: 2`, every hardware-specific flag `false`;
* intgemm8avx2 has `isIntgemm: true`, `size: 1`, `isAvx2: true`
* etc.

So size of the type is a correct distinctive feature, no? That being said, at least a comment would be helpful there, yes.

src/graph/expression_operators.cpp
emjotde: You would need to list all 8 types here. See explanation below.

src/layers/generic.cpp
emjotde: Something to think about, but not in this PR maybe? This whole shortlist business is quite horrible already, having to pass even further down makes it worse IMO. This would benefit from a cleaner, better thought out approach.

src/tensors/cpu/backend.h
emjotde: Not against it, but can we do that in the next PR? Maybe there are better ways than having it in backend (maybe not).

emjotde:
> As for SSSE/SSE not working on avx2, this is because the multiply here: https://github.com/marian-nmt/marian-dev/blob/intgemm_refactor/src/tensors/cpu/intgemm_interface.h#L407 is width dependent. Potentially we might want to make it read in the type.

I don't get that. The code clearly is width-dependent, so is the type, no? How is there a mismatch?

emjotde:
IMO the big thing missing here is now making sure that everything that's hardware-specific either runs correctly (like intgemm16sse2 on avx2) or croaks FBGEMM-style. So we can pull this into your branch and then look at the ABORTs etc. there. Or we can finish it here first. Maybe with matching regression tests.

src/tensors/cpu/backend.h
kpu: Should we just get rid of the unshifted version?

XapaJIaMnu:
> > As for SSSE/SSE not working on avx2, this is because the multiply here: https://github.com/marian-nmt/marian-dev/blob/intgemm_refactor/src/tensors/cpu/intgemm_interface.h#L407 is width dependent. Potentially we might want to make it read in the type.
> 
> I don't get that. The code clearly is width-dependent, so is the type, no? How is there a mismatch? Or are you saying it is using the AVX2 code?

Sorry, this "width" here is just int8 or int16. The instruction "width" is determined by runtime initialised function pointers. If you want the avx2 for example, you need to do `intgemm::avx2_int8::Multiply`.
If you do `intgemm::Int8::Multiply` you get dispatched to the highest architecture your machine supports

XapaJIaMnu:
> @XapaJIaMnu `sizeOf(...)` is just our equivalent for C++ `sizeof` that works with the `Type` enum. As you know the types are defined by their type classes and their size.
> 
> * intgemm16 has `isIntgemm: true`, `size: 2`, every hardware-specific flag `false`;
> * intgemm8avx2 has `isIntgemm: true`, `size: 1`, `isAvx2: true`
> * etc.
> 
> So size of the type is a correct distinctive feature, no? That being said, at least a comment would be helpful there, yes.

Is intgemm the only 16bit one? There's GPU fp16, etc..?

src/tensors/cpu/backend.h
XapaJIaMnu: It's slower on pre-VNNI without precomputed Alphas, which we don't have here.

src/tensors/cpu/expression_graph_packable.h
```diff
@@ -200,25 +161,81 @@ class ExpressionGraphPackable : public ExpressionGraph {
...
+                                                   cols(val));
+          } else if(isAvx512(gemmElementType)) {
+            // @TODO: we should croak here if the hardware type is not supported?
+            intgemm::AVX512_8bit::PrepareBTransposed(tmp->data(), /*input*/
+                                                     paramMat->data<int8_t>(), /*output*/
+                                                     quantMult, /*Quant Mult*/
+                                                     rows(val),
+                                                     cols(val));
+          } else {
+            ABORT_IF(gemmElementType != Type::intgemm8, "Type {} is not supported {}", gemmElementType); // shouldn't really happen, but let's make sure
```
snukky, line 192: The surplus `{}` causes a formatting error before the abort.

src/tensors/cpu/expression_graph_packable.h
```diff
@@ -200,25 +161,81 @@ class ExpressionGraphPackable : public ExpressionGraph {
...
+                                                    cols(val));
+          } else if(isAvx512(gemmElementType)) {
+            // @TODO: we should croak here if the hardware type is not supported?
+            intgemm::AVX512_16bit::PrepareBTransposed(tmp->data(), /*input*/
+                                                      paramMat->data<int16_t>(), /*output*/
+                                                      quantMult, /*Quant Mult*/
+                                                      rows(val),
+                                                      cols(val));
+          } else {
+            ABORT_IF(gemmElementType != Type::intgemm8, "Type {} is not supported {}", gemmElementType); // shouldn't really happen, but let's make sure
```
snukky, line 228: The surplus `{}` causes a formatting error before the abort.

snukky: Also, it should be `gemmElementType != Type::intgemm16` here, otherwise I'm getting:

    Error: Type intgemm16 is not supported

when converting a model using `intgemm16`:

    marian-conv -f model.npz -t model.bin --gemm-type intgemm16

src/tensors/cpu/expression_graph_packable.h
emjotde: Oops. Good catch.

emjotde:
> Is intgemm the only 16bit one? There's GPU fp16, etc..?

Of course, that's why there is a check for `isIntgemm` above :) I suppose this should actually have all 8 Intgemm types which would also avoid the sse2 vs avx2 issue?

```
 } else if(isFloat(aElementType) && isIntgemm(bElementType)) {
      // @TODO: this branch should move into cpu::integer::*
      if(sizeOf(bElementType) == 1) {
        return cpu::integer::dot<Type::int8>(a, b, transA, transB, scale);
      } else if(sizeOf(bElementType) == 2) {
        return cpu::integer::dot<Type::int16>(a, b, transA, transB, scale);
      } else {
        ABORT("Wrong size for Intgemm type {}??", sizeOf(bElementType));
      }
    } 
```

emjotde:
So it seems to be we should have a `cpu::integer::dot` function (and affine) rather than the two distinguished by built-in type and that one would dispatch the correct multiplies.

XapaJIaMnu:
> > Is intgemm the only 16bit one? There's GPU fp16, etc..?
> 
> Of course, that's why there is a check for `isIntgemm` above :) I suppose this should actually have all 8 Intgemm types which would also avoid the sse2 vs avx2 issue?
> 
> ```
>    } else if(isFloat(aElementType) && isIntgemm(bElementType)) {
>       // @TODO: this branch should move into cpu::integer::*
>       if(sizeOf(bElementType) == 1) {
>         return cpu::integer::dot<Type::int8>(a, b, transA, transB, scale);
>       } else if(sizeOf(bElementType) == 2) {
>         return cpu::integer::dot<Type::int16>(a, b, transA, transB, scale);
>       } else {
>         ABORT("Wrong size for Intgemm type {}??", sizeOf(bElementType));
>       }
>     } 
> ```

No, it will not avoid (automatically) the issue of sse2 vs avx2. We need to update all of `intgemm_interface.h` to not use the dispatcher: `ingemm::width::X` needs to become `intgemm::avx2_8bit::X` (For example)

emjotde:
> We need to update all of `intgemm_interface.h` to not use the dispatcher: `ingemm::width::X` needs to become `intgemm::avx2_8bit::X` (For example)

OK, then we should do that.

src/tensors/cpu/backend.h
ykim362: From my curiosity, could you let me how much the perf diff is between shifted and unshifted?

ykim362 (approved):
Looks good!

emjotde:
BTW. On-the-fly conversion will be easy to add back once we merge with my fp16 changes. There are `--precision` flags that operate on the model-loading level. We can just do an on-the-fly conversion there as I do for fp16 (but to intgemm8 or intgemm16) and then the corrected dispatching will just take care of the rest.

emjotde:
@ykim362  That will also enable on-the-fly conversion with fbgemm when we do that correctly. :)

ykim362:
> @ykim362 That will also enable on-the-fly conversion with fbgemm when we do that correctly. :)

@emjotde That sounds good! AML is using a branch for that.

src/tensors/cpu/backend.h
emjotde: If the difference isn't dramatic I would be in favor of being pragmatic about this :)

emjotde:
OK, as far as I am concerned this is more or less done. Removes some functionality, but it will be OK to add it back after we merge with the fp16 branch which should come with a few things to make that easier.

emjotde:
Mmapping also works now for everything that runs on my AVX2 machine. Couldn't test AVX512.

emjotde:
@kpu any remarks before this goes into the original PR? (And then into master if compilation etc. succeeds)

emjotde:
OK, pulling into the original branch. @snukky and @XapaJIaMnu are working on regression tests, once those are updated and tested, we can pull it in.

PR #768, opened 2020-12-02
Small optimizations

rhenry-nv:
### Description

This PR splits out some small optimizations from PR #743.

Performance improvements from this PR as measured on a Titan V using a proxy transformer model:

Times with 1 Stream
-- | -- | -- | -- | --
1 | 204.41 | 166.74 | 0.184286483 | 1.225920595
2 | 136.77 | 114.46 | 0.163120567 | 1.194915254
4 | 84.76 | 69.39 | 0.181335536 | 1.221501657
8 | 50.32 | 41.52 | 0.174880763 | 1.21194605
16 | 29.94 | 25.07 | 0.162658651 | 1.194256083
32 | 18.51 | 15.98 | 0.136682874 | 1.158322904
64 | 12.09 | 10.57 | 0.125723739 | 1.143803217
128 | 8.19 | 7.15 | 0.126984127 | 1.145454545
256 | 6.02 | 5.39 | 0.104651163 | 1.116883117

Times with two streams
-- | -- | -- | -- | --
1 | 156.74 | 119.03 | 0.240589511 | 1.316810888
2 | 106.13 | 79.63 | 0.249693772 | 1.33278915
4 | 65.1 | 48.64 | 0.252841782 | 1.338404605
8 | 38.46 | 29.07 | 0.244149766 | 1.323013416
16 | 22.68 | 17.29 | 0.237654321 | 1.311740891
32 | 14.31 | 10.98 | 0.232704403 | 1.303278689
64 | 9.51 | 7.19 | 0.243953733 | 1.322670376
128 | 6.66 | 5.18 | 0.222222222 | 1.285714286
256 | 5.05 | 4.06 | 0.196039604 | 1.243842365

List of changes:
- Uses the per thread default stream for cublas
- Uses the strided batched gemm cublas call when possible for the batchedGemm. 
- In the general batchedGemm case, reduces the number of memcpy calls from 3 to 1.
- Rounds the width of input batches to a multiple of 8 when the GPU backend is being used. This is to enable better use of tensorcores on Volta architectures and newer
- Adds notices to the files changed

### How to test
I ran the regression tests. On Volta, some regression tests fail due to the additional use of tensor cores. This is because prior to CUDA 11, cublas does not use tensorcores if matrices fail to meet alignment requirements. This restriction is lifted in CUDA >= 11.

The differences in the float output from the failing regression tests are small. A list of them are provided here:
  - tests/interface/input-tsv/test_tsv_train_with_align.sh
  - tests/interface/input-tsv/test_tsv_train_with_align_and_weights.sh
  - tests/interface/input-tsv/test_tsv_train_with_align_and_weights_inputtypes.sh
  - tests/interface/input-tsv/test_tsv_train_with_align_pos0.sh
  - tests/interface/input-tsv/test_tsv_train_with_align_shuffle_in_ram.sh
  - tests/interface/input-tsv/test_tsv_train_with_align_stdin.sh
  - tests/scorer/nbest/test_compare_parallel_and_nbest.sh
  - tests/training/features/mixed-ensembles/test_ensemble_of_different_s2s.sh
  - tests/training/features/mixed-ensembles/test_ensemble_of_s2s_and_transformer.sh
  - tests/training/features/guided-alignment/test_guided_alignment_rnn.sh
  - tests/training/features/guided-alignment/test_guided_alignment_transformer.sh

OS: Ubuntu 18.04.3 LTS
Compiler gcc 7.5.0
nvcc version: 10.1.243

cmake command:

cmake .. -DCOMPILE_CPU=on -DCOMPILE_CUDA=on -DUSE_SENTENCEPIECE=on -DUSE_STATIC_LIBS=off -DCOMPILE_SERVER=off -DUSE_FBGEMM=on   -DCOMPILE_CUDA_SM35=off -DCOMPILE_CUDA_SM50=off -DCOMPILE_CUDA_SM60=off -DCOMPILE_CUDA_SM70=on -DCOMPILE_CUDA_SM75=off -DCOMPILE_TESTS=on

### Checklist

### DIFF
--- a/src/data/corpus.cpp
+++ b/src/data/corpus.cpp
@@ -1,3 +1,8 @@
+/* Part of this file was contributed by NVIDIA under license:
+ *   Copyright (C) 2020 NVIDIA Corporation
+ *   SPDX-License-Identifier: MIT
+ */
+
 #include "data/corpus.h"
 
 #include <numeric>
@@ -265,6 +270,14 @@ CorpusBase::batch_ptr Corpus::toBatch(const std::vector<Sample>& batchVector) {
         maxDims[i] = (int)ex[i].size();
     }
     sentenceIds.push_back(ex.getId());
+  }
+  
+  // When running on GPU, we want the batchWidth to be a multiple of 8 for better tensorcore usage
+  if(options_->get<int>("cpu-threads") == 0) {
+    constexpr int roundingFactor = 8;
+    for(size_t j = 0; j < maxDims.size(); ++j) {
+      maxDims[j] = roundingFactor * ((maxDims[j] + roundingFactor - 1) / roundingFactor);
+    }
   }
 
   std::vector<Ptr<SubBatch>> subBatches;
--- a/src/rnn/types.h
+++ b/src/rnn/types.h
@@ -1,4 +1,4 @@
-/* All or part of this file was contributed by NVIDIA under license:
+/* Part of this file was contributed by NVIDIA under license:
  *   Copyright (C) 2020 NVIDIA Corporation
  *   SPDX-License-Identifier: MIT
  */
--- a/src/rnn/types.h
+++ b/src/rnn/types.h
@@ -1,3 +1,8 @@
+/* All or part of this file was contributed by NVIDIA under license:
+ *   Copyright (C) 2020 NVIDIA Corporation
+ *   SPDX-License-Identifier: MIT
+ */
+
 #pragma once
 
 #include "common/definitions.h"
--- a/src/rnn/types.h
+++ b/src/rnn/types.h
@@ -1,5 +1,7 @@
 #pragma once
 
+#include "common/definitions.h"
+#include "common/shape.h"
 #include "marian.h"
 
 #include <iostream>
@@ -12,7 +14,7 @@ struct State {
   Expr output;
   Expr cell;
 
-  State select(const std::vector<IndexType>& selIdx, // [beamIndex * activeBatchSize + batchIndex]
+  State select(Expr selIdx, // [beamIndex * activeBatchSize + batchIndex]
                int beamSize, bool isBatchMajor) const {
     return{ select(output, selIdx, beamSize, isBatchMajor),
             select(cell,   selIdx, beamSize, isBatchMajor) };
@@ -20,15 +22,14 @@ struct State {
 
   // this function is also called by Logits
   static Expr select(Expr sel, // [beamSize, dimTime, dimBatch, dimDepth] or [beamSize, dimBatch, dimTime, dimDepth] (dimTime = 1 for RNN)
-                     const std::vector<IndexType>& selIdx, // [beamIndex * activeBatchSize + batchIndex]
+                     Expr selIdx, // [beamIndex * activeBatchSize + batchIndex]
                      int beamSize, bool isBatchMajor)
   {
     if (!sel)
       return sel; // keep nullptr untouched
 
     sel = atleast_4d(sel);
-
-    int dimBatch = (int)selIdx.size() / beamSize;
+    int dimBatch =(int) selIdx->shape().elements()/beamSize;
     int dimDepth = sel->shape()[-1];
     int dimTime  = isBatchMajor ? sel->shape()[-2] : sel->shape()[-3];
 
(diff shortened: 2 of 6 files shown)


src/data/corpus.cpp
```diff
@@ -266,6 +271,14 @@ CorpusBase::batch_ptr Corpus::toBatch(const std::vector<Sample>& batchVector) {
     }
```
frankseide, line 271: Would you mind updating the comment with the `maxDims` variable? It is not obvious from the code what it means, and why it is OK to round it to multiples of 8. (Based on the variable name, it seems you are changing the hidden dimension, which would not work.) And if the variable is misleadingly named, then would you mind naming it better (according to your change below, it seems related to batch width = max # words)? Thanks!

src/data/corpus.cpp
```diff
@@ -1,3 +1,8 @@
+/* Part of this file was contributed by NVIDIA under license:
+ *   Copyright (C) 2020 NVIDIA Corporation
+ *   SPDX-License-Identifier: MIT
```
frankseide, line 3: What is this for? It seems the only change is the rounding of `maxDims`. What NVidia contribution was made here?

In general, I would be opposed to changing the comment style for licence. Marian source code does not have license information in the source files directly, but rather in a separate license file. Please let's continue to follow that pattern.

src/rnn/types.h
```diff
@@ -83,8 +89,24 @@ class States {
   States select(const std::vector<IndexType>& selIdx, // [beamIndex * activeBatchSize + batchIndex]
                 int beamSize, bool isBatchMajor) const {
     States selected;
+    Expr indices;
+    // I think this doesn't work if model split among gpus but not sure if it matters
+    
+    for (auto& state : states_) {
```
frankseide, line 95: Please add a comment what this logic does, as it is not obvious why the old code is not working.

src/rnn/types.h
```diff
@@ -83,8 +89,24 @@ class States {
...
+      if (state.cell) {
+        indices = state.cell->graph()->indices(selIdx);
+        break;
+      }
+
+      if (state.output) {
+        indices = state.output->graph()->indices(selIdx);
+        break;
+      }
```
frankseide, line 104: What happens if neither? Is that a valid condition? If not, let's change this to `else`

src/tensors/gpu/prod.cpp
```diff
@@ -289,50 +350,68 @@ void ProdBatchedTyped(marian::Tensor C,
...
+                                               ldb, strideB, 
+                                               A->data<const T>(), 
+                                               lda, strideA, 
+                                               &beta, 
+                                               C->data<T>(), 
+                                               ldc, strideC, 
+                                               batchA));
+    unsetTensorMode(cublasHandle);
+  } else {
+    auto batchC = std::max(batchA, batchB);
```
frankseide, line 374: I was confused about what `batchA` etc. mean. Since you are at it, would you mind helping to rename these variables to be more clear? I think it should be `batchDimA`, right? Thanks!

src/data/corpus.cpp
rhenry-nv: It seemed to be the max # of words in a batch. I just kept the same name but I can rename to `maxWordsInBatch`

src/data/corpus.cpp
rhenry-nv: Yes that was the only contribution. However, I was told to include a notice even in files where I make one line changes. This is just me following instructions.

src/rnn/types.h
rhenry-nv: There is nothing wrong with the old code. The comment you highlighted is also wrong I think. I wrote this when I was much less familiar with Marian so I will just remove that comment.

src/tensors/gpu/prod.cpp
rhenry-nv: Ok I can rename those

src/rnn/types.h
rhenry-nv: I think neither is a valid condition

src/data/corpus.cpp
rhenry-nv: Also, I just saw the second part of your comment. Is there a process for adding NVIDIA to the license file? I'm not sure what the solution is here.

kpu: The one that hasn't been updated since 2016 and doesn't even name Microsoft?  I guess add it to your PR and shame Marcin.

rhenry-nv: @kpu I think this is the one Frank was referring to. I'll ask Marcin if this is ok. I will also need to check to see if we are internally ok with removing the notices assuming we are added to the license file.

@emjotde What do you suggest as the way forward?

src/rnn/types.h
frankseide: Ah, because it's a loop, sorry. But what happens if it never matches any of the conditions? Then `indices` ends up being NULL. Is it worth to `ABORT_IF` for that?

rhenry-nv: Good point. I think `indices` can only end up being NULL if all of the states' `output` and `cell` fields are null. In that case, we will return a vector of nulls which was the behavior of the original code. (Also, I think these values are NULL on the first run of a network so we want this behavior).

src/tensors/gpu/prod.cpp
rhenry-nv: (Done is recent commits)

src/rnn/types.h
rhenry-nv: I added a comment in a recent commit but I'm not sure if it's clear enough. If so, feel free to resolve this.

src/data/corpus.cpp
rhenry-nv: (renamed in recent commits)

src/data/corpus.cpp
rhenry-nv: I checked internally and I can add NVIDIA to the main license file and remove the notices in all the other files! 

I will take care of this in all the PRs I have submitted.

PR #770, opened 2020-12-04
Topk refactor

rhenry-nv:
### Description
This PR rewrites the topk implementation. It is an optimization to benefit large batch and does not hinder small batch performance. This PR also changes the topk operator in topk.cu to use the new implementation.

There is a temporary template parameter in the topk launcher to enable/disable indexing relative to a given row. This can be removed if nth_element.cu is removed.

Performance improvements relative to PR #768 

Times with 1 stream
--
-- | -- | -- | -- | --
1 | 166.74 | 166.392 | 0.002087082 | 1.002091447
2 | 114.46 | 113.005 | 0.012711864 | 1.012875536
4 | 69.39 | 69.5436 | -0.002213575 | 0.997791314
8 | 41.52 | 40.7636 | 0.018217726 | 1.01855577
16 | 25.07 | 23.9823 | 0.043386518 | 1.045354282
32 | 15.98 | 14.7954 | 0.074130163 | 1.080065426
64 | 10.57 | 9.6 | 0.091769158 | 1.101041667
128 | 7.15 | 6.48 | 0.093706294 | 1.103395062
256 | 5.39 | 4.65 | 0.13729128 | 1.159139785

Times with 2 streams
--
-- | -- | -- | -- | --
1 | 119.03 | 116.7 | 0.019574897 | 1.019965724
2 | 79.63 | 78.46 | 0.014692955 | 1.014912057
4 | 48.64 | 47.83 | 0.016652961 | 1.016934978
8 | 29.07 | 28.107 | 0.033126935 | 1.034261928
16 | 17.29 | 16.69 | 0.03470214 | 1.03594967
32 | 10.98 | 10.24 | 0.067395264 | 1.072265625
64 | 7.19 | 6.57 | 0.086230876 | 1.094368341
128 | 5.18 | 4.65 | 0.102316602 | 1.113978495
256 | 4.06 | 3.47 | 0.145320197 | 1.170028818

List of changes:
- Introduces cub as a dependency
- Reimplements the topK kernel and replaces the call in nth_element.cu and topk.cu

Added dependencies: cub

### How to test
Run the unit tests. There are unit tests that cover the topk operator and they all still passed.

CMake command: cmake .. -DCOMPILE_CPU=on -DCOMPILE_CUDA=on -DUSE_SENTENCEPIECE=on -DUSE_STATIC_LIBS=off -DCOMPILE_SERVER=off -DUSE_FBGEMM=on   -DCOMPILE_CUDA_SM35=off -DCOMPILE_CUDA_SM50=off -DCOMPILE_CUDA_SM60=off -DCOMPILE_CUDA_SM70=on -DCOMPILE_CUDA_SM75=off -DCOMPILE_TESTS=on

Ubuntu - 18.04.3 LTS
nvcc - 10.1.243
gcc - 7.5.0

### Checklist

- [x ] I have tested the code manually
- [x ] I have run regression tests
- [x ] I have read and followed CONTRIBUTING.md
- [x ] I have updated CHANGELOG.md

### DIFF
--- a/src/3rd_party/topk.cuh
+++ b/src/3rd_party/topk.cuh
@@ -23,9 +23,13 @@
 #pragma once
 #include <cuda_runtime.h>
 #include <cuda_fp16.h>
+#if CUDA_VERSION >= 11000
+#include <cub/cub.cuh>
+#include <cub/util_type.cuh>
+#else
 #include "cub/cub/cub.cuh"
 #include "cub/cub/util_type.cuh"
-
+#endif
 #define MAX_BLOCKS_PER_BEAM 8
 
 struct TopK {
--- a/src/3rd_party/topk.cuh
+++ b/src/3rd_party/topk.cuh
@@ -23,7 +23,6 @@
 #pragma once
 #include <cuda_runtime.h>
 #include <cuda_fp16.h>
-#include "cub/cub/util_type.cuh"
 #if CUDA_VERSION >= 11000
 #include <cub/cub.cuh>
 #else
--- a/src/3rd_party/topk.cuh
+++ b/src/3rd_party/topk.cuh
@@ -23,7 +23,6 @@
 #pragma once
 #include <cuda_runtime.h>
 #include <cuda_fp16.h>
-#include <condition_variable>
 #if CUDA_VERSION >= 11000
 #include <cub/cub.cuh>
 #else
--- a/src/3rd_party/topk.cuh
+++ b/src/3rd_party/topk.cuh
@@ -23,6 +23,7 @@
 #pragma once
 #include <cuda_runtime.h>
 #include <cuda_fp16.h>
+#include "cub/cub/util_type.cuh"
 #if CUDA_VERSION >= 11000
 #include <cub/cub.cuh>
 #else
@@ -36,22 +37,34 @@ struct TopK {
   IndexType p = 0;
   T u = cub::FpLimits<T>::Lowest();
 
-  __device__ __forceinline__ void insert(T elem, IndexType elem_id) {
+  __device__ __forceinline__ void insertKeepMax(T elem, IndexType elem_id) {
     if(elem > u) {
       u = elem;
       p = elem_id;
     }
   }
 
-  __device__ __forceinline__ void init() {
-    u = cub::FpLimits<T>::Lowest();
+  __device__ __forceinline__ void insertKeepMin(T elem, IndexType elem_id) {
+    if(elem < u) {
+      u = elem;
+      p = elem_id;
+    }
+  }
+
+  __device__ __forceinline__ void init(bool descendingOrder) {
+    u = descendingOrder? cub::FpLimits<T>::Lowest() : cub::FpLimits<T>::Max();
     p = 0;
   }
 };
 
 template<typename IndexType, typename T>
-__device__ __forceinline__ TopK<IndexType, T> reduce_topk_op(const TopK<IndexType, T>& a, const TopK<IndexType, T>& b) {
+__device__ __forceinline__ TopK<IndexType, T> reduce_topk_max(const TopK<IndexType, T>& a, const TopK<IndexType, T>& b) {
   return a.u > b.u ? a : b;
+}
+
+template<typename IndexType, typename T>
+__device__ __forceinline__ TopK<IndexType, T> reduce_topk_min(const TopK<IndexType, T>& a, const TopK<IndexType, T>& b) {
+  return a.u < b.u ? a : b;
 }
 
 template<typename IndexType, typename T, int BLOCK_SIZE_, int BLOCKS_PER_BEAM_>
@@ -59,7 +72,8 @@ __global__ void topk_stage_1(T* log_probs,
                              IndexType* topk_tmp_id_buf,
                              T* topk_tmp_val_buf, 
                              const int k, 
-                             const int vocab_size) {
+                             const int vocab_size,
+                             const int descendingOrder) {
 
   typedef cub::BlockReduce<TopK<IndexType, T>, BLOCK_SIZE_> BlockReduce;
   __shared__ typename BlockReduce::TempStorage temp_storage;
@@ -71,9 +85,10 @@ __global__ void topk_stage_1(T* log_probs,
   const IndexType tmp_log_buf_index = row_id * vocab_size; 
   const IndexType tmp_topk_buf_index = row_id * BLOCKS_PER_BEAM_ * k + block_lane * k;
   TopK<IndexType, T> partial;
+  const T minimal = descendingOrder? cub::FpLimits<T>::Lowest() : cub::FpLimits<T>::Max();
 
   for(int ite = 0; ite < k; ite++) {
-    partial.init();
+    partial.init(descendingOrder);
     const IndexType threadStart = tid + block_lane * BLOCK_SIZE_;
 
     // This is needed to ensure the indices for the threads in each valid block starts in a valid range for that block.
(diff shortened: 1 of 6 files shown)


src/3rd_party/topk.cuh
```diff
@@ -0,0 +1,302 @@
...
+
+  typedef cub::BlockReduce<TopK<IndexType, T>, BLOCK_SIZE_> BlockReduce;
+  __shared__ typename BlockReduce::TempStorage temp_storage;
+  extern __shared__ char array[];
+  T *s_val = topk_tmp_val_buf + batch_id * size;
+  TopK<IndexType, T> *topks = (TopK<IndexType, T>*)(array);
+  
+  TopK<IndexType, T> partial;
+
+  for(int ite = 0; ite < k; ite++) {
```
frankseide, line 168: Would you mind adding a headline comment to each of the sections in this function to summarize in a single line what they achieve? It is faster to read a single line than having to parse the code and understand the logic. Thanks!

src/3rd_party/topk.cuh
```diff
@@ -0,0 +1,302 @@
...
+        beams_per_batch, \
+        k, descendingOrder); \
+  break; \
+
+// The getRowOffsets template parameter is added so the topk implementation works with both nth_element.cu and the topk operator.
+// It is a template parameter since we know at compile time which version of topk we want to call. This flag can be removed whenever nth
+// element.cu is removed. When this flag is true, the indices returns are the offsets within the row. When the flag is false, the indices
+// returned are offset from the base pointer.
+template <typename IndexType, typename T, bool getRowOffsets=false>
+void topK_kernelLauncher(T* log_probs,
```
frankseide, line 217: In addition to the comment that mentions some detail information, please also add a function-level comment saying what this function is meant to do.

src/3rd_party/topk.cuh
```diff
@@ -0,0 +1,302 @@
...
+                                                                                                      outIndices,
+                                                                                                      outVals,
+                                                                                                      beams_per_batch,
+                                                                                                      k,
+                                                                                                      descendingOrder);
+    break;
+  }
+}
+
+template <typename IndexType, typename T, bool getRowOffsets=false>
```
frankseide, line 260: Please add a function-level comment what htis does, and how it differs from the previous overload above.

src/translator/nth_element.cu
```diff
@@ -284,82 +27,38 @@ public:
...
-    CUDA_CHECK(cudaMalloc((void**)&d_res,     maxBatchSize * maxBeamSize * sizeof(float)));
-
-    CUDA_CHECK(cudaHostAlloc((void**)&h_res,     maxBeamSize * maxBatchSize * sizeof(float), cudaHostAllocDefault));
-    CUDA_CHECK(cudaHostAlloc((void**)&h_res_idx, maxBeamSize * maxBatchSize * sizeof(int), cudaHostAllocDefault));
-
-    CUDA_CHECK(cudaMalloc((void**)&d_breakdown, maxBeamSize * sizeof(float)));
-    CUDA_CHECK(cudaMalloc((void**)&d_batchPosition, (maxBatchSize + 1) * sizeof(int)));
-    CUDA_CHECK(cudaMalloc((void**)&d_cumBeamSizes,  (maxBatchSize + 1) * sizeof(int)));
+    const int tempElts = maxBatchSize * maxBeamSize * maxBeamSize * MAX_BLOCKS_PER_BEAM;
+    CUDA_CHECK(cudaMalloc((void**)&topk_tmp_id_buf, tempElts * sizeof(IndexType)));
```
frankseide, line 36: Can we change this to use a Marian allocator?

src/translator/nth_element.cu
```diff
@@ -373,68 +72,47 @@ public:
...
     if(scores->type() == Type::float32) {
-      float disabledPathScore = NumericLimits<float>(scores->type()).lowest;
-      selectNBest(scores->data<float>(), batchFirstElementIdxs, cumulativeBeamSizes, disabledPathScore);
+      selectNBest(scores->data<float>(), dimBatch, inputN, N, vocabSize);
+      getPairs<float>(dimBatch * N, outKeys, outCosts);
 #if COMPILE_FP16
     } else if(scores->type() == Type::float16) {
-      float disabledPathScore = NumericLimits<float>(scores->type()).lowest;
-      selectNBest(scores->data<half>(), batchFirstElementIdxs, cumulativeBeamSizes, disabledPathScore);
+      selectNBest(scores->data<half>(), dimBatch, inputN, N, vocabSize);
```
frankseide, line 86: What's the difference between `dimBatch` and `beamsPerBatch`?

frankseide (changes requested):
Thanks for doing this! I like this work very much.

My main general comment is that it would be very helpful to have more high-level comments, for all functions and data structures, as well as high-level blocks inside functions, to state what it does and what it is for.

I understand that some people do not like comments, as it makes the code look less clean. However, high-level comments like this are a service to the reader of the code, since reading a 1 or 2 line comment will be much faster than having to parse and understand the logic of the code itself. Consider that most of the time, readers of the code do not actually care or need to understand the detail implementation of everything. Often, readers need to quickly orient themselves to find the one place that they need to look at.

Thanks!

src/translator/nth_element.cu
rhenry-nv: `dimBatch` is the batch dimension and `beamsPerBatch' is the number of hypotheses for each sentence in the batch. I can add a comment in the code.

src/translator/nth_element.cu
rhenry-nv: I don't plan on doing this right now. The original code used cudaMallocs so I kept the same pattern. I believe the Marian team plans to remove this in the future entirely and use the topk operator so topk will use the Marian allocator when that work occurs.

src/3rd_party/topk.cuh
```diff
@@ -0,0 +1,302 @@
...
+
+  __device__ __forceinline__ void insertKeepMin(T elem, IndexType elem_id) {
+    if(elem < u) {
+      u = elem;
+      p = elem_id;
+    }
+  }
+
+  __device__ __forceinline__ void init(bool descendingOrder) {
+    u = descendingOrder? -FpInfinity<T>::infinity() : FpInfinity<T>::infinity();
```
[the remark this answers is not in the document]
rhenry-nv: I don't think I understand what you mean entirely. We need to initialize negative inf for descending order and positive inf for ascending. It seems that this check needs to happen at runtime.

One option that would remove the check at runtime would be to make the descending order flag a template parameter. In that case, the topk operator in topk.cu would call the correct kernel which wouldn't have this check. However, I would expect that the performance benefits to doing this are small since this is a small % of the overall execution and all threads in a warp will end up taking the same path.

If I misunderstood what you meant here please clarify!

src/translator/nth_element.cu
frankseide: Thanks. This is an incorrect use of the term "beam." [@emjotde -- this is an example of what I mentioned to you last week]. In correct usage, "beam" refers to the set of hypotheses, not a single one. How about rename it to `beamWidths`?

src/3rd_party/topk.cuh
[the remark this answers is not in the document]
frankseide: My question was whether we ever use it both ways, or only one way, throughout Marian. If we only ever use `descending = true`, for example, we could remove this option altogether. Probably OK to keep as is. Will close this.

src/translator/nth_element.cu
rhenry-nv: I see. I will address this when I address the rest of your feedback. (Thanks for all the helpful feedback btw!)

(Also, leave this open so that I remember to address it. I will close as resolved after.)

rhenry-nv:
I think I have addressed all of your feedback @frankseide.

I added detailed comments about the functions and made the stylistic changes you suggested. Let me if there's anything else that you think needs to be changed.

PR #772, opened 2020-12-08
Refactor of beam search to process factor groups in parallel

rhenry-nv:
### Description
This PR refactors the beam search to process the secondary factors in parallel.

As is, this work significantly reduces the H2D communication required when processing the secondary factors in a model with a factored vocabulary.

Note - The changes in this PR are not integrated into PR #743. The table below shows the improvements on top of PR #770.

Times   with 1 stream
--
Batch | Initial Time (s) | Current Time(s) | % Runtime reduction | Speedup factor
-- | -- | -- | -- | --
1 | 166.392 | 162.169 | 0.025379826 | 1.026040735
2 | 113.005 | 109.386 | 0.032025132 | 1.033084673
4 | 69.5436 | 66.8378 | 0.038907966 | 1.04048308
8 | 40.7636 | 39.2434 | 0.037293075 | 1.038737724
16 | 23.9823 | 23.2354 | 0.031143802 | 1.032144917
32 | 14.7954 | 14.3573 | 0.029610555 | 1.030514094
64 | 9.6 | 9.01939 | 0.060480208 | 1.064373533
128 | 6.48 | 6.04882 | 0.066540123 | 1.071283325
256 | 4.65 | 4.28963 | 0.077498925 | 1.084009577

Times   with 2 streams
--
Batch | Initial Time (s) | Current Time(s) | % Runtime reduction | Speedup factor
-- | -- | -- | -- | --
1 | 116.7 | 110.365 | 0.05428449 | 1.057400444
2 | 78.46 | 74.248 | 0.053683406 | 1.056728801
4 | 47.83 | 45.4441 | 0.049882919 | 1.052501865
8 | 28.107 | 26.7974 | 0.046593375 | 1.048870413
16 | 16.69 | 15.6832 | 0.060323547 | 1.064196082
32 | 10.24 | 9.79104 | 0.04384375 | 1.045854169
64 | 6.57 | 6.22622 | 0.052325723 | 1.055214882
128 | 4.65 | 4.22329 | 0.091765591 | 1.101037343
256 | 3.47 | 3.2422 | 0.065648415 | 1.070260934

List of changes:
- Adds a kernel to perform a max reduction on the last axis of a tensor for GPU. This cuts down on the kernel launches needed and removes a stream synchronize for every call.
- Beam search refactor to batch secondary factors
- Some changes from PR #768 to reduce index copying.

Added dependencies: cub

### How to test
I ran the regression tests and tested manually with a proxy model.

[...]

PR #776, opened 2020-12-15
Factor maxes op

rhenry-nv:
### Description
This PR adds a new inference operator to GPU for getting the lemma logits for a factored vocabulary. It demonstrates significant speedup on GPU inference over PR #772 

Here are some perf numbers relative to PR #772 

Times from a proxy model with 1 stream as measured on a Titan V.

Batch | Time from #772 (s) | Current Time(s) | Speedup factor
-- | -- | -- | --
1 | 162.169 | 105.638 | 1.53513887
2 | 109.386 | 73.2865 | 1.492580489
4 | 66.8378 | 45.1652 | 1.479851744
8 | 39.2434 | 27.1344 | 1.446260098
16 | 23.2354 | 16.1808 | 1.43598586
32 | 14.3573 | 10.1867 | 1.4094162
64 | 9.01939 | 6.45151 | 1.398027749
128 | 6.04882 | 4.39468 | 1.376396006
256 | 4.28963 | 3.2189 | 1.332638479

Times from a proxy model with two streams as measured on a Titan V

Batch | Time from #772 (s) | Current Time(s) | Speedup factor
-- | -- | -- | --
1 | 110.365 | 94.137 | 1.172387053
2 | 74.248 | 61.736 | 1.202669431
4 | 45.4441 | 37.2677 | 1.219396421
8 | 26.7974 | 21.81 | 1.22867492
16 | 15.6832 | 12.6072 | 1.243987563
32 | 9.79104 | 7.70231 | 1.271182282
64 | 6.22622 | 4.86047 | 1.280991344
128 | 4.22329 | 3.3429 | 1.263361153
256 | 3.2422 | 2.52556 | 1.28375489

List of changes:
- Adds a new operator used only for GPU inference 
- Adds a cache by name function to the expression graph so store a tensor containing whether each lemma has a given factor group. (There may be a better way to do this but it wasn't obvious to me). This tensor needs to reside permanently on the GPU since the D2H copy time would be too great for this optimization to be worthwhile otherwise.

Added dependencies: cub

### How to test
I ran the regression tests and they all passed. I also manually tested on a proxy model and the outputs from master exactly match the outputs after this change was made.

Ubuntu - 18.04.3 LTS
nvcc - 10.1.243
gcc - 7.5.0

### Checklist

### DIFF
--- a/src/layers/generic.cpp
+++ b/src/layers/generic.cpp
@@ -1,3 +1,8 @@
+/* Part of this file was contributed by NVIDIA under license:
+ *   Copyright (C) 2020 NVIDIA Corporation
+ *   SPDX-License-Identifier: MIT
+ */
+
 #include "marian.h"
 
 #include "layers/generic.h"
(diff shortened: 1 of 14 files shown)


rhenry-nv:
FYI: I am currently requesting internally to remove the notices in each file and for NVIDIA to be added to the license file. I will take care of the licensing once I get confirmation.

src/layers/generic.cpp
```diff
@@ -77,15 +83,25 @@ namespace marian {
...
       sel = sel - max(sel, -1);
-    }
-    else {
+    } else {
       auto numGroups = getNumFactorGroups();
-      for (size_t g = 1; g < numGroups; g++) {
-        auto factorMaxima = max(logits_[g]->loss(), -1);
-        auto factorMasks = constant(getFactorMasks(g, shortlist ? shortlist->indices() : std::vector<WordIndex>()));
-        sel = sel + factorMaxima * factorMasks; // those lemmas that don't have a factor get multiplied with 0
+      if(numGroups > 1 && graph()->isInference() && graph()->getBackend()->getDeviceId().type == DeviceType::gpu) {
```
rhenry-nv, line 88: This fork is something I wasn't sure how to remove. It would be better if it was under the expression operator but moving it down causes the operator interface to be a bit ugly and introduces some code duplication. Feedback on this in particular would be greatly appreciated.

PR #777, opened 2020-12-15
Adds some cuda 11 FP16 SPMM support and fixes beam search to work with FP16

rhenry-nv:
### Description

This PR aims to allow factored vocabulary models to use FP16 with CUDA 11. It also fixes the beam search so that factored vocabulary models work with FP16.

List of changes:
- Adds a new code path in CSRProd for fp16 to support the simple SPMM case (no swapping or transposing)
- Changes getFactorLogits to use the graph constant method so that the factorMasks are casted to FP16 when the graph default type is fp16.
- Changes the INVALID_PATH_SCORE  in the beam search to negative infinity instead of float min. When running FP16, the float min was being cast to FP16 causing it to be -inf. This caused the beam search to fail when creating new hypotheses. I chose to set the INVALID_PATH_SCORE to -infinity since it maintains the same value before and after the cast so we can reliably check for this regardless of if we are using FP32 or FP16.
- Changes the minimal value in TopK to be negative infinity to account for the change above.

Added dependencies: none

### How to test

I ran the regression tests and most passed with CUDA 11. With CUDA 10, the regression tests pass as expected. I also manually tested on a proxy model with fp16 and it no longer gives errors. The results from the proxy model look sensible.

Ubuntu - 18.04.3 LTS
nvcc - 10.1.243
gcc - 7.5.0

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 
 ### Added
+- Changes the invalid path score in the beam search to be negative infinity instead of the lowest float. The beam search did not work with fp16 for factored vocab models due to float min being cast to fp16.
+- Adds spmm fp16 support for CUDA 11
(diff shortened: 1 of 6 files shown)


rhenry-nv:
Closing work achieving a similar result was done and is integrated into master.

PR #803, opened 2021-02-22
Add documentation platform based on Sphinx+Doxygen+Breathe+Exhale

snukky:
### Description
This PR sets up a framework based on Sphinx, Doxygen, Breathe, and Exhale for creating developer documentation.

At this moment, changes are independent from the source code.

Added dependencies: none

### How to test
See `doc/README.md`. There will be a lot of warnings when the C++ API is generated, but it works.

### Checklist
- [ ] I have run regression tests (**not relevant**)


doc/index.rst
```diff
@@ -0,0 +1,48 @@
...
+   api/library_index
+
+   contributing
+
+
+Indices and tables
+==================
+
+* :ref:`genindex`
+* :ref:`modindex`
```
qianqianzhu, line 26: I guess the links of `modindex` and `search` have not been set yet.  The `Module Index` page cannot be accessed. The search function does not work.

qianqianzhu (approved):
Many thanks for taking care of the new doc platform! I can build the website successfully and it looks nice. Only some links need to be set up, but we can improve this later.

doc/index.rst
snukky: Good catch! I disabled `modindex` as it isn't very relevant to C++ docs.
Search had two issues:
1. Searching didn't work in Sphinx 2.0.0, so I updated it to highest 2.x.x, which currently is 2.4.4.
2. The search page is empty by default in the `sphinx_rtd_theme`, but there seem to be some PRs open for this.
So I simply removed `search`, because it's accessible from the top of the side bar anyway.

graemenail (approved):

PR #809, opened 2021-03-01
Update filter pattern in GitHub release action

snukky:
This PR updates the filter pattern in the GitHub release action to match git tags we actually use during releases.


snukky:
GitHub Actions do not work if the branch has a `/` in its name... Created a new PR.

PR #810, opened 2021-03-01
Update GitHub Ubuntu actions

snukky:
This PR should fix the recent issue with the 'Ubuntu 18.04 CUDA 10.2 gcc-8' action failing due to out of disk space. Unit tests and examples will not be compiled for this action.

We may also consider compiling only towards specific GPU architectures.


snukky:
Closing. GitHub Actions do not work if the branch has a `/` in its name.

PR #819, opened 2021-03-02
Generate documentation artifact with GitHub action

### LINKED ISSUE #813: Create GitHub workflow for building code documentation
Create a GitHub workflow building the developer documentation from `doc/`.
* Running on `ubuntu-latest`.
* Starting for each pull request and push to master.
* Generating a downloadable artifact from `doc/build/html/`.

graemenail:
### Description
Automates the building of the documentation in `doc/` and provides a downloadable artifact. Runs on pushes and PRs to master.

Closes #813

List of changes:
- Adds new documentation workflow using GitHub actions.

Added dependencies: none

### How to test
Tested in a fork. Workflow uses `ubuntu-latest` which is currently 18.04. I have also tested 20.04 explicitly due to actions/virtual-environments#1816

### Checklist

- [ ] I have run regression tests *not needed*

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -1,0 +1,47 @@
+name: Documentation
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  api-documentation:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          submodules: recursive
+
+      - name: Set up Doxygen
+        run: sudo apt-get install -y doxygen
+
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.7
+
+      - name: Set up dependency cache
+        uses: actions/cache@v2
+        with:
+          path: ~/.cache/pip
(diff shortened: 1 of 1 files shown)


.github/workflows/documentation.yml
```diff
@@ -0,0 +1,47 @@
...
+
+      - name: Build documentation
+        working-directory: ./doc
+        run: make html
+
+      - name: Upload documentation
+        uses: actions/upload-artifact@v2
+        with:
+          name: api-docs
+          path: ./doc/build/html
```
snukky, line 46: I think this produces an archive without the parent directory, correct? If so, could you add a comment that `api-docs.zip` will only include the content of `./doc/build/html` without the parent directory?

snukky (approved):
Thanks, looks good! I have only one suggestion to add a comment on what the structure of the generated artifact is.

.github/workflows/documentation.yml
graemenail: Correct - it's the HTML output of the sphinx only.

A common the the YAML would suffice?

snukky: Yes, please add a comment to the .yml file, for example above the step name. Thanks!

Issue #744, opened 2020-10-10
[Question] What prevents DefaultCommunicator from working with multi-node MPI


XapaJIaMnu:
Hey,

I have been experimenting with some CPU training (without CUDA/NCCL) and I run into this `ABORT` statement.
https://github.com/marian-nmt/marian-dev/blob/master/src/training/communicator.h#L128

However, I don't see immediately what implementation part is missing in order to get full MPI support. I tried commenting out the `ABORT` and it seems that it works. Are gradients not exchanged? Git blame shows @frankseide put it there?

Could you guys let me know what part I should add to properly fix this issue?

Cheers,

Nick

emjotde:
We talked per e-mail, but basically you are welcome to own that part.

PR #833, opened 2021-03-17
[RFC] Initial CPU MPI implementation

XapaJIaMnu:
### Description
This add initial CPU MPI support, fixing https://github.com/marian-nmt/marian-dev/issues/744 the limitation is that you need to have only one cpu-thread per process. It also only supports "global" sharding mode (for now).

Are you guys upstream interested in a CPU MPI implementation? Comments on my implementation are much welcome.
 
### How to test
`mpirun -n X /path/to/marian -c config.yml` as long as `--cpu-threads: 1 `

### Checklist

### DIFF
--- a/src/training/communicator.cpp
+++ b/src/training/communicator.cpp
@@ -8,6 +8,7 @@
 #if MPI_FOUND
 #include "mpi.h"
 #endif
+#include "training/communicator_mpi.h"
 
 namespace marian {
 
@@ -25,6 +26,9 @@ Ptr<ICommunicator> createCommunicator(
   const std::vector<Ptr<ExpressionGraph>>& graphs,
   bool noNccl, ShardingMode shardingMode, Ptr<IMPIWrapper> mpi) {
   mpi;
+  if (mpi && mpi->numMPIProcesses() > 1 && graphs[0]->getBackend()->getDeviceId().type == DeviceType::cpu) {
+    return New<MpiCommunicator>(graphs, shardingMode, mpi);
+  }
 #if defined(CUDA_FOUND) && defined(USE_NCCL)
   if(noNccl) {
     LOG(warn, "[comm] NCCL communicator overridden");
@@ -136,6 +140,24 @@ public:
     if (sendbuf == recvbuf)
       sendbuf = MPI_IN_PLACE; // MSMPI requires this
     HANDLE_MPI_ERROR(MPI_Allreduce(sendbuf, recvbuf, (int)count, datatype, op, comm));
+  }
+  virtual void reduceScatter(const void * sendbuf, void * recvbuf, int * recvcounts, MPI_Datatype datatype, MPI_Op op, MPI_Comm comm) const override {
+    if (sendbuf == recvbuf) {
+      sendbuf = MPI_IN_PLACE; // OpenMPI hopefully works with this: https://www.open-mpi.org/doc/v3.1/man3/MPI_Reduce_scatter.3.php
+    }
+    HANDLE_MPI_ERROR(MPI_Reduce_scatter(sendbuf, recvbuf, recvcounts, datatype, op, comm));
+  }
+  virtual void reduceScatterBlock(const void * sendbuf, void * recvbuf, int count, MPI_Datatype datatype, MPI_Op op, MPI_Comm comm) const override {
+    if (sendbuf == recvbuf) {
+      sendbuf = MPI_IN_PLACE; // OpenMPI hopefully works with this: https://www.open-mpi.org/doc/v3.0/man3/MPI_Reduce_scatter_block.3.php
+    }
+    HANDLE_MPI_ERROR(MPI_Reduce_scatter_block(sendbuf, recvbuf, count, datatype, op, comm));
+  }
+  virtual void Allgather(const void *sendbuf, int sendcount, MPI_Datatype sendtype, void *recvbuf, int recvcount, MPI_Datatype recvtype, MPI_Comm comm) const override {
+    if (sendbuf == recvbuf) {
+      sendbuf = MPI_IN_PLACE; // OpenMPI hopefully works with this: https://www.open-mpi.org/doc/v3.0/man3/MPI_Allgather.3.php
+    }
+    HANDLE_MPI_ERROR(MPI_Allgather(sendbuf, sendcount, sendtype, recvbuf, recvcount, recvtype, comm));
   }
   virtual void finalize() override {
     HANDLE_MPI_ERROR(MPI_Finalize());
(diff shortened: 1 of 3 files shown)


snukky:
Windows builds fail properly and need to be fixed: some warnings from the new code show up and we treat them as errors.

XapaJIaMnu:
@snukky thank you. I was more so asking whether we are at all interested to have this and is it a problem that for now it has the limitation that it would only work with one thread per process and only in "global" gradient shard mode.

Also, we don't have any MPI tests, or we do?

emjotde:
I guess having it doesn't hurt. Did you have a use case?

snukky:
> Also, we don't have any MPI tests, or we do?

No, I think we don't, but it would be great to have at least a unit test.

XapaJIaMnu:
> I guess having it doesn't hurt. Did you have a use case?

We're expecting some hardware that isn't made by nvidia.

emjotde:
@XapaJIaMnu Oh interesting. BTW, local sharding for a setup like 8 GPUs and 4 nodes is about 25% faster.

XapaJIaMnu:
@emjotde yeah I looked at it, that's what I expect to be more efficient, but due to differences in MPI and NCCL, implementing it is a bit more complicated and I've put it on the queue.

Cheers,

Nick

src/training/communicator.cpp
```diff
@@ -25,6 +26,9 @@ Ptr<ICommunicator> createCommunicator(
   const std::vector<Ptr<ExpressionGraph>>& graphs,
   bool noNccl, ShardingMode shardingMode, Ptr<IMPIWrapper> mpi) {
   mpi;
```
emjotde, line 28: Looks like `mpi;'` line is redundant now.

src/training/communicator.h
```diff
@@ -84,6 +84,18 @@ struct/*interface*/ IMPIWrapper {
...
   virtual void allReduce(const void* sendbuf, void* recvbuf, size_t count, MPI_Datatype datatype, MPI_Op op, MPI_Comm comm = MPI_COMM_WORLD) const = 0;
+  virtual void reduceScatter(const void * sendbuf, void * recvbuf, int * recvcounts, MPI_Datatype datatype, MPI_Op op, MPI_Comm comm = MPI_COMM_WORLD) const {
+    sendbuf; recvbuf; recvcounts; datatype; op; comm; // unused in the fakeMPI wrapper
+    ABORT("ReduceScatter is only implemented when compiled with -DUSE_MPI=ON");
+  }
+  virtual void reduceScatterBlock(const void * sendbuf, void * recvbuf, int count, MPI_Datatype datatype, MPI_Op op, MPI_Comm comm = MPI_COMM_WORLD) const {
+    sendbuf; recvbuf; count; datatype; op; comm; // unused in the fakeMPI wrapper
+    ABORT("ReduceScatter is only implemented when compiled with -DUSE_MPI=ON");
+  }
+  virtual void Allgather(const void *sendbuf, int sendcount, MPI_Datatype sendtype, void *recvbuf, int recvcount, MPI_Datatype recvtype, MPI_Comm comm = MPI_COMM_WORLD) const {
```
emjotde, line 95: Casing: `allGather`?

src/training/communicator.h
```diff
@@ -84,6 +84,18 @@ struct/*interface*/ IMPIWrapper {
   virtual void sSend(void* buf, size_t count, MPI_Datatype datatype, size_t destRank, int tag, MPI_Comm comm = MPI_COMM_WORLD) const = 0;
   virtual void recv(void* buf, size_t count, MPI_Datatype datatype, size_t sourceRank, int tag, MPI_Comm comm = MPI_COMM_WORLD, MPI_Status* status = MPI_STATUS_IGNORE) const = 0;
   virtual void allReduce(const void* sendbuf, void* recvbuf, size_t count, MPI_Datatype datatype, MPI_Op op, MPI_Comm comm = MPI_COMM_WORLD) const = 0;
+  virtual void reduceScatter(const void * sendbuf, void * recvbuf, int * recvcounts, MPI_Datatype datatype, MPI_Op op, MPI_Comm comm = MPI_COMM_WORLD) const {
+    sendbuf; recvbuf; recvcounts; datatype; op; comm; // unused in the fakeMPI wrapper
+    ABORT("ReduceScatter is only implemented when compiled with -DUSE_MPI=ON");
+  }
```
emjotde, line 90: Since those are non-pure virtual functions with implementations, let's add empty lines around them for readability.

emjotde: Similar below.

src/training/communicator_mpi.h
```diff
@@ -0,0 +1,324 @@
...
+    return graphs_[0]->params()->vals()->size();
+  }
+
+  // determine the (max) shard size
+  // All shards except the last one have this size.
+  // Presently, even all shards must have identical size, due to a limitation in NCCL we have not yet worked around.
+  size_t shardSize() const {
+    size_t numShards = shardingMode_ == ShardingMode::global ? numRanks() : numLocalRanks();
+    size_t size = (dataSize() + numShards - 1) / numShards;
+#if 1 // for now, all shards must have the same size, since NCCL does not allow a sub-slice for the last shard
```
emjotde, line 46: Adapt comment to reflect that this was copied from the NCCL communicator.

emjotde: Similar above and any other mention of NCCL in this file.

src/training/communicator_mpi.h
```diff
@@ -0,0 +1,324 @@
...
+
+      auto grads = graphs_[i]->params()->grads();
+      const auto* sendbuf = grads->data();
+      //auto*       recvbuf = grads->subtensor(begin, end-begin)->data();
+      size_t      bufsize = shardSize();
+      ABORT_IF(grads->subtensor(begin, end-begin)->size() != bufsize, "unexpected subtensor size??");
+
+      MPI_Datatype mpiFLoatType = MPI_FLOAT;
+      if(grads->type() == Type::float16)
+        ABORT("Half precision is datatype is not supported by MPI.");
```
emjotde, line 149: Grammar? One "is" too many?

src/training/communicator_mpi.h
```diff
@@ -0,0 +1,324 @@
+// Note: This must only be included if defined(CUDA_FOUND) && defined(USE_NCCL)
```
emjotde, line 1: Outdated?

src/training/communicator_mpi.h
```diff
@@ -0,0 +1,324 @@
...
+      }
+    }
+    else { // no MPI: localData is the complete result already
+      data = std::move(localData);
+    }
+    return data;
+  }
+};
+
+}  // namespace marian
```
emjotde, line 324: Generally a ton of code duplication with NCCL communicator. Once you think about adding local sharding, might be worth to combine? Will make it easier to maintain and make sure that it does not go down the path of the other MPI implementations that ended up not being used and removed in the end.

emjotde (approved):
Generally OK. Issues are mostly about comments. I am assuming this to be early code, so not insisting on combining with NCCLCommunicator (yet).

emjotde:
Unless you would like to start combining it with NCCL communicator already?

XapaJIaMnu:
MPI communicator offers a subset of the functionality that NCCL does. When (eventually) we want to allow multiple threads per process, the implementation of the two big collective operation functions will start differing. 

What do you propose? 

I *thought* it made more sense to keep separate communicators, but I admit there is more code duplication than what I would like. I can try to inherit the NCCL communicator, but it has a lot of cuda specific stuff that would need to be hidden away. Furthermore, intel provices oneCCL, an abstraction over MPI, which supports more datatypes than MPI (But only works with intel MPI). I guess we're not interested in having more than one communication backends as there would be very few interested users.

I'm ok with delaying the merge until I have a more complete implementation, as the code is fairly self contained, and i find it unlikely any changes to master would break it. (Unless you are planning on some extra MPI work?)

Issue #835, opened 2021-03-17
Issue with Marian FBGEMM build for cascadelake CPU (AVX512)


GokulNC:
### Bug description
I am try to build Marian-NMT for my Intel CascadeLake CPU which supports AVX512.

```sh
$ grep avx512 /proc/cpuinfo
flags           : fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush mmx fxsr sse sse2 ss ht syscall nx pdpe1gb rdtscp lm constant_tsc rep_good nopl xtopology nonstop_tsc cpuid tsc_known_freq pni pclmulqdq ssse3 fma cx16 pcid sse4_1 sse4_2 x2apic movbe popcnt aes xsave avx f16c rdrand hypervisor lahf_lm abm 3dnowprefetch invpcid_single ssbd ibrs ibpb stibp ibrs_enhanced fsgsbase tsc_adjust bmi1 hle avx2 smep bmi2 erms invpcid rtm mpx avx512f avx512dq rdseed adx smap clflushopt clwb avx512cd avx512bw avx512vl xsaveopt xsavec xgetbv1 xsaves arat avx512_vnni md_clear arch_capabilities
```

But I face the following error during `make`:
```
[ 27%] Building CXX object src/3rd_party/fbgemm/CMakeFiles/fbgemm_avx2.dir/src/QuantUtilsAvx2.cc.o
/tmp/ccAtxjvY.s: Assembler messages:                                                              
/tmp/ccAtxjvY.s:49401: Error: invalid use of register 
...
make[2]: *** [src/3rd_party/fbgemm/CMakeFiles/fbgemm_avx2.dir/build.make:128: src/3rd_party/fbgemm/CMakeFiles/fbgemm_avx2.dir/src/QuantUtilsAvx2.cc.o] Error 1                                      
make[1]: *** [CMakeFiles/Makefile2:1053: src/3rd_party/fbgemm/CMakeFiles/fbgemm_avx2.dir/all] Error 2                                                                                               
make: *** [Makefile:152: all] Error 2
```

 But I do not face this error on my Broadwell CPU which has support only till AVX2.

### How to reproduce

Try on a Intel Cascade Lake CPU maybe? (not sure if the issue is because of this arch)
To check if this bug is because of original FBGEMM (and not because of Marian), I tried building the latest FBGEMM from the original repo. It got built perfectly without any errors.

[...]

PR #836, opened 2021-03-18
Fix FBGEMM build with gcc 9.3+

ykim362:
### Description
Build failed with gcc 9.3+ when USE_FBGEMM is on.

This PR fixes bugs reported #835 and #709 .

List of changes:
- In top level CMakeList.txt, '-march=native' is turned off when built with FBGEMM.

Added dependencies: none

### How to test
Compile with gcc 9.3.

### Checklist


emjotde:
Hm, that's a very broad change to fix such a specific issue. Can we narrow this down a bit? Do we know what exactly is going on here?

ykim362:
@emjotde When compiling avx2 target on avx512 machines, the newer compilers are doing some target specific optimizations which generates broken assembly. This only affects when USE_FBGEMM is on though. The best thing we can do is removing '-march=native' for avx2 target on avx512 machines. But, it's not likely we are getting too much from the native architecture flag.

ykim362:
@emjotde I made a change that the original CMAKE_CXX_FLAGS is not touched.

snukky:
I updated the branch with master to fix failures in GitHub checks that were unrelated to this PR.

CMakeLists.txt
```diff
@@ -216,6 +216,11 @@ else(MSVC)
   set(CMAKE_C_FLAGS_PROFILE         "${CMAKE_C_FLAGS_RELEASE} -pg")
   set(CMAKE_C_FLAGS_PROFGEN         "${CMAKE_C_FLAGS_RELEASE} -fprofile-generate -fprofile-correction")
   set(CMAKE_C_FLAGS_PROFUSE         "${CMAKE_C_FLAGS_RELEASE} -fprofile-use -fprofile-correction")
+
+  # '-march=native' makes avx2 assembly generation fail on AVX512 CPUs with gcc 9.3+
+  if (USE_FBGEMM)
+    string(REPLACE "-march=native" "" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
```
snukky, line 222: Maybe we should apply this only if GCC is >= 9.2? For example (I did not test it):

    if(USE_FBGEMM AND CMAKE_COMPILER_IS_GNUCC AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 9.2)

emjotde:
That is OK, I think. Please add a message that gets displayed when GCC 9.3+ is used.

ykim362:
@emjotde I've added a message.

kpu:
All the more reason to dis FBGEMM but really scope this stuff instead of breaking it for everything.

emjotde:
@kpu Sorry, can be a bit clearer regarding the second part?

kpu:
This isn't really about `-march=native`, it's about the flags that it happens to map to on some hardware.  And should be able to compile Marian with whatever architecture flags, such as `-mavx512bw`, I want. If a specific file has a problem with that, unset the flags only for that file, not the entire project.

emjotde:
Happy to comply. Feel free to use the review interface to leave a hint where and how to do that.

kpu:
Go fix your cmake around her to filter options: https://github.com/marian-nmt/FBGEMM/blob/64888200cbc9e9c287601f23c58b4088cdf51cba/CMakeLists.txt#L126

emjotde:
@ykim362 can you take a look? That seems indeed like a better idea.

ykim362:
@emjotde @kpu Sounds good. I will try to fix there, and will close this PR if that works. Thanks!

ykim362:
@emjotde @kpu Now, all the changes are in the FBGEMM submodule. It works fine. Please take a look. Thanks!

snukky:
Could you also mention this fix in the changelog before merging it into master? Thanks!

ykim362:
@snukky I added the fix in the changelog. Thanks!

emjotde:
Thanks @ykim362

emjotde (approved):

Issue #828, opened 2021-03-08
Uncontrollable memory growth when growing the workspace during training/decoding


XapaJIaMnu:
### Bug description
In cases where the workspace is too small for a mini-batch during training  (when not using mini-batch-fit) or decoding (when there is unusually long sentence inside the mini-batch), the marian workspace [reallocs](https://github.com/marian-nmt/marian-dev/blob/db771e09bd90b839efa580bca62a6b08500a1f33/src/tensors/cpu/device.cpp#L17) more memory. This is done through the [grow](https://github.com/marian-nmt/marian-dev/blob/db771e09bd90b839efa580bca62a6b08500a1f33/src/tensors/allocator.h#L127) function. 

However, it seems that as this function is called multiple times, there is an issue of uncontrollable memory growth, which often results in an OOM crash. If the workspace is big enough so that grow is seldom called there is no uncontrollable memory growth.

Reproducible example can be found here: https://nbogoychev.com/files/memory_leak_example.tar.xz  . When training (on the CPU) with `mini-batch-fit: false` and `workspace: 2000` (config file provided in the archive: `config.debuglower.yml`), we get uncontrollable memory growth at the 11th mini-batch. Marian starts allocating more and more workspace memory, however the memory consumption grows much faster than the workspace. At some point marian reaches 770 GBs of memory usage, while the workspace sits at just 18 GBs. Eventually the system runs out of memory and kills marian.

In contrast if we increase the workspace to 40 GBs (config file provided in the archive: `config.debuglower40g.yml`), from the beginning, we have no extra allocations and no uncontrollable memory growth.

This is not a memory leak, as the memory is still reachable and when marian exits cleanly it deallocates it. Valgrind doesn't detect anything abnormal

I believe the issue is due to an STL container continuously growing and/or some unfortunate smart pointer circular references, which gets triggered as a by-process of the workspace expansion. I suspect https://github.com/marian-nmt/marian-dev/blob/db771e09bd90b839efa580bca62a6b08500a1f33/src/tensors/allocator.h#L91 as it increases its size by 1 after each realloc, but I don't have a proof. I believe this issue has existed in the codebase for a number of years and various people have run into this issue during decoding and the solution has always been "Increase the workspace" or "decrease the mini-batch size".

### How to reproduce
Run CPU training with the configurations provide here:  https://nbogoychev.com/files/memory_leak_example.tar.xz

### Context
[...]

PR #840, opened 2021-03-21
remove TC_MALLOC from optional dependencies

emjotde:
### Description
Removes optional dependency on TCMALLOC following discussion in #828
Benchmarking showed it doesn't actually seem to increase performance, but early version cause weird memory leak.

### Checklist


emjotde:
Decided to remove the dependency since our cmake script doesn't seem to have the option to discriminate based on version and I cannot be bothered.

snukky (approved):

PR #847, opened 2021-03-30
Online quantization

ykim362:
### Description
This is a PR to enable online quantization. This makes us to quantize output projection embedding matrix.

List of changes:
- expression_operators - add logic to quantize (pack) online

Added dependencies: none

### How to test
Run a model with online quantization. Add parameters'--optimize --gemm-type {type to quantize}'

### Checklist

### DIFF
--- a/src/graph/expression_operators.cpp
+++ b/src/graph/expression_operators.cpp
@@ -522,6 +522,8 @@ Expr dot(Expr a, Expr b, bool transA, bool transB, float scale) {
         return Expression<DotNodeOp>(
           a, b, transA, transB, scale);
       }
+    } else if(isFloat(aElementType) && isIntgemm(bElementType)) {
+      return cpu::integer::affineOrDot(a, b, nullptr, transA, transB, scale);
     } else if(isFloat(aElementType) && isPacked(bElementType)) {
 #if USE_FBGEMM
       // 07/10/2019 - Use packed GEMM only if the cpu architecture supports AVX2
@@ -620,6 +622,8 @@ Expr affine(Expr a, Expr b, Expr bias, bool transA, bool transB, float scale) {
       } else {
         return affineDefault(a, b, bias, transA, transB, scale);
       }
+    } else if(isFloat(aElementType) && isIntgemm(bElementType)) {
+      return cpu::integer::affineOrDot(a, b, bias, transA, transB, scale);
     } else if(isFloat(aElementType) && isPacked(bElementType)) {
 #if USE_FBGEMM
       // 07/10/2019 - Use packed GEMM only if the cpu architecture supports AVX2


XapaJIaMnu:
intgemm used to implement this in the past but @emjotde said that the whole tensor caching mechanism would be removed in the future so the feature got removed from the pull request. Is there a change in the development plans?

kpu:
This feature was already in intgemm but removed in the pull request. . . will you reconsider?

kpu:
Not that Nick and I coordinate messages at all

ykim362:
There's a use-case for this online qunatization. In that scenario, a separate branch is used, so it couldn't get all the latest changes.

XapaJIaMnu:
Does it mean it's ready to go into master? Can we get our intgemm online quantization here?

ykim362:
I hope we can get both online quantization into into the master. And, maybe we can support only 8-bit quantizations only, for the online quantization? That may make the code cleaner?

XapaJIaMnu:
We had both 8bit and 16bit, as well online quantisation of the output layer (or the shortlist). We can update those to the upstream version and perhaps merge with you?

emjotde:
@XapaJIaMnu 
For some context. This is needed for an MS-internal team, unfortunately. I am actually not really happy with how we are solving this at the moment and would prefer to not pull it, but have to for the moment. Adding more stuff on-top of this which would cement this solution in more, which would not be great.

emjotde:
@ykim362 this seems to be breaking intgemm, though. Checkout these regression tests:

  - tests/decoder/intgemm/test_intgemm_16bit.sh
  - tests/decoder/intgemm/test_intgemm_16bit_avx2.sh
  - tests/decoder/intgemm/test_intgemm_16bit_sse2.sh
  - tests/decoder/intgemm/test_intgemm_8bit.sh
  - tests/decoder/intgemm/test_intgemm_8bit_avx2.sh
  - tests/decoder/intgemm/test_intgemm_8bit_ssse3.sh
  - tests/models/wnmt18/test_student_small_aan_intgemm16.sh
  - tests/models/wnmt18/test_student_small_aan_intgemm8.sh

All of these now fail with an abort.

emjotde:
Note, some of these may fail on different machines anyway, but due to numeric issues on the outputs, not during actual inference.

ykim362:
Sure, @emjotde . I will check the regression tests.

ykim362:
@emjotde I fixed the regression tests for intgemm. But, the model tests seem broken for the master as well. They give this error. 
This is common for 
  - tests/models/wnmt18/test_student_small_aan.sh
  - tests/models/wnmt18/test_student_small_aan_intgemm16.sh
  - tests/models/wnmt18/test_student_small_aan_intgemm8.sh

```
[2021-05-24 21:32:31] Error: Invalid activation name 'sigmoid'
[2021-05-24 21:32:31] Error: Aborted from std::function<IntrusivePtr<marian::Chainable<IntrusivePtr<marian::TensorBase> > >(IntrusivePtr<marian::Chainable<IntrusivePtr<marian::TensorBase> > >)> marian::activationByName(const string&) in /data2/youki/marian-dev-external/src/layers/generic.h:182
```

emjotde:
Yes, I have a fix for that. Forgot to push.

emjotde:
Pushed. Thanks for the fix.

ykim362:
The model tests also work fine with the patch. Thanks!

emjotde (approved):

PR #860, opened 2021-04-20
Scorer model loading

### LINKED ISSUE #831: [Enhancement] Loading a model currently loads the model 2Xnum_threads times
### Bug description
Currently when loading a model for decoding in marian we load the full model twice per thread. 
Once in order to load the yaml which is stored inside the model: https://github.com/marian-nmt/marian-dev/blob/f74d055d204d7ed417f3dd26ad469192d70e8112/src/translator/scorers.cpp#L69

Which calls getItem: https://github.com/marian-nmt/marian-dev/blob/f74d055d204d7ed417f3dd26ad469192d70e8112/src/common/binary.cpp#L110
Which is basically loading the model into a vector and discarding it.

And another time in order to load the actual model: https://github.com/marian-nmt/marian-dev/blob/f74d055d204d7ed417f3dd26ad469192d70e8112/src/translator/scorers.cpp#L88

Ideally we want to load the model into a Vector once by calling https://github.com/marian-nmt/marian-dev/blob/f74d055d204d7ed417f3dd26ad469192d70e8112/src/common/binary.cpp#L30

And passing this vector down to the individual threads when they create their scorers, and also reuse it to get the yaml configuration.

graemenail:
### Description

When creating scorers, there are two calls to loadItems:
  1. `getYamlFromModel` loads the 'special' YAML from the model
  2. When the scorer is created, a `EncoderDecoder` method results in a call to `ExpressionGraph::load`

This PR addresses these by calling `io::loadItems` in advance and passing the Item vector when creating scorers.
  1. `getYamlFromModel` can now be called directly from an Item vector
  2. `ScorerWrapper` can be constructed from an Item vector.

`Amun` and `Nematus` do some specific preprocessing before loading the graph, this PR moves that logic into a load from Items.

Closes #831 

List of changes:
- Adds `--model-mmap` as a command line option (uses mio)
- Preloads model to Item vector before creating scorers
- try-catch when loading model YAML is replaced with if-else
- Extended `IEncoderDecoder` to load from Item vector
- `Amun` and `Nematus` models load overload for Item vector

Added dependencies: None

### How to test
Used marian-conv to test mmap loading
Loading from Items passes regression tests

### Checklist


src/translator/translator.h
```diff
@@ -275,6 +276,15 @@ class TranslateService : public ModelServiceTask {
...
 
+    // preload models
+    std::vector<std::vector<io::Item>> model_items_;
+    auto models = options->get<std::vector<std::string>>("models");
+    for(auto model : models) {
+      auto items = io::loadItems(model);
+      model_items_.push_back(std::move(items));
+    }
+
+
```
snukky, line 287: Double empty line.

src/common/config_parser.cpp
```diff
@@ -184,6 +184,8 @@ void ConfigParser::addOptionsModel(cli::CLIWrapper& cli) {
     }
   }
 
+  cli.add<bool>("--model-mmap",
+      "Use memory-mapping when loading model (CPU only)");
```
snukky, line 188: Should this option be added only in CPU compilations? I.e. `#if COMPILE_CPU`? I'm not sure what the convention actually is.

src/translator/translator.h
```diff
@@ -72,15 +66,19 @@ class Translate : public ModelTask {
...
     graphs_.resize(numDevices_);
 
-#if MMAP
     auto models = options->get<std::vector<std::string>>("models");
-    for(auto model : models) {
-      marian::filesystem::Path modelPath(model);
-      ABORT_IF(modelPath.extension() != marian::filesystem::Path(".bin"),
-              "Non-binarized models cannot be mmapped");
-      mmaps_.push_back(std::move(mio::mmap_source(model)));
+    if(options_->get<bool>("model-mmap")) {
```
snukky, line 70: Consider using the getter with a default value as this might be safer/cleaner, i.e. `get<bool>("model-mmap", false)`.

src/common/config_parser.cpp
snukky: Also, is this option relevant to the training as well, or can it be added to the translation options only?

src/translator/translator.h
```diff
@@ -92,11 +90,14 @@ class Translate : public ModelTask {
         graph->reserveWorkspaceMB(options_->get<size_t>("workspace"));
         graphs_[id] = graph;
 
-#if MMAP
-        auto scorers = createScorers(options_, mmaps_);
-#else
-        auto scorers = createScorers(options_);
-#endif
+        std::vector<Ptr<Scorer>> scorers;
+        if(options_->get<bool>("model-mmap")) {
```
snukky, line 94: As above, maybe the getter with a default value provided.

snukky (approved):
Looks good, only minor comments from me. Thanks! I will ask @emjotde to take a look too.

snukky:
Please add an entry about the model-mmap option to CHANGELOG.

src/common/config_parser.cpp
graemenail: The mmap code is currently only in translator; the option will only appear when compiled with CPU support and when performing translation.

I added a validation to ensure it is only enabled when cpu-threads is not 0 (i.e. is a non-GPU backend)

snukky:
This regression test fails for me with this PR:

    tests/training/restarting/test_restarting_finished.sh

It fails because logs no longer have the message saying `'Loading model'` when trying to restart a training that has already finished. @graemenail, Do you think this is expected now? If so, I will just update the test.

This was tested after rebasing with the current master. That's the only regression test that fails.

graemenail:
> This regression test fails for me with this PR:
> 
> ```
> tests/training/restarting/test_restarting_finished.sh
> ```
> 
> It fails because logs no longer have the message saying `'Loading model'` when trying to restart a training that has already finished. @graemenail, Do you think this is expected now? If so, I will just update the test.
> 
> This was tested after rebasing with the current master. That's the only regression test that fails.

Thanks for looking at this @snukky. I believe this is expected, the `Loading model from {}` log messages contained the filename of the model loaded. In this PR, the model is loaded beforehand, and so the filename is replaced with the `vector<io:Items>` from that file.

That being said, I think it's a useful log message to include. I'll see where is best to reinstate it, and update this PR.

graemenail:
I have reinstated the log messages for Amun and Netamus, which should fix the regression.
I have also added log messages when loading model ahead of `createScorer` as well.

PR #863, opened 2021-04-29
Enable compute86 where supported

XapaJIaMnu:
### Description
Add compute86 support where available. Compute86 devices are all RTX 30XX cards. Note that compute80 seems to be enough for them to run, but compute86 might enable extra features, @rhenry-nv ?

### Checklist


rhenry-nv:
@XapaJIaMnu 

In general SM Major.x code will run on SM Major.y enabled hardware when x <= y, but without extra features as you said.

The SM86 arch has 2X FP32 throughput compared to the SM80 arch. You will miss out on this without the compute86 flag. There may be other features as well that I am not aware of but you right.

CMakeLists.txt
```diff
@@ -325,6 +325,16 @@ if(CUDA_FOUND)
...
     LIST(APPEND COMPUTE -Wno-deprecated-gpu-targets)
   endif()
+  if(CUDA_VERSION VERSION_EQUAL "11.1" OR CUDA_VERSION VERSION_GREATER "11.1")
+    option(COMPILE_KEPLER      "Compile GPU version with SM35 support" OFF) # deprecated for CUDA 11
+    option(COMPILE_MAXWELL     "Compile GPU version with SM50 support" OFF) # deprecated for CUDA 11
+    option(COMPILE_PASCAL      "Compile GPU version with SM60 support" ON)
+    option(COMPILE_VOLTA       "Compile GPU version with SM70 support" ON)
+    option(COMPILE_TURING      "Compile GPU version with SM75 support" ON)
+    option(COMPILE_AMPERE      "Compile GPU version with SM80 support" ON)
+    option(COMPILE_AMPERE_RTX  "Compile GPU version with SM80 support" ON)
```
rhenry-nv, line 335: Should the message here say SM86?

XapaJIaMnu: Yes, sorry

CMakeLists.txt
```diff
@@ -325,6 +325,16 @@ if(CUDA_FOUND)
...
     LIST(APPEND COMPUTE -Wno-deprecated-gpu-targets)
   endif()
+  if(CUDA_VERSION VERSION_EQUAL "11.1" OR CUDA_VERSION VERSION_GREATER "11.1")
+    option(COMPILE_KEPLER      "Compile GPU version with SM35 support" OFF) # deprecated for CUDA 11
+    option(COMPILE_MAXWELL     "Compile GPU version with SM50 support" OFF) # deprecated for CUDA 11
+    option(COMPILE_PASCAL      "Compile GPU version with SM60 support" ON)
+    option(COMPILE_VOLTA       "Compile GPU version with SM70 support" ON)
+    option(COMPILE_TURING      "Compile GPU version with SM75 support" ON)
+    option(COMPILE_AMPERE      "Compile GPU version with SM80 support" ON)
+    option(COMPILE_AMPERE_RTX  "Compile GPU version with SM86 support" ON)
```
snukky, line 335: It seems that all options are defined twice for CUDA 11.1? To avoid code duplication, can you just add an if statement defining COMPILE_AMPERE_RTX only for CUDA 11.1+ into the if-block for CUDA 11+ above? Alternatively, a condition checking that CUDA is lower than 11.1 could be added into the first if statement.

XapaJIaMnu: So you want something like
```
if(CUDA_VERSION VERSION_EQUAL "11.0" OR CUDA_VERSION VERSION_GREATER "11.0")
    option(COMPILE_KEPLER  "Compile GPU version with SM35 support" OFF) # deprecated for CUDA 11
    option(COMPILE_MAXWELL "Compile GPU version with SM50 support" OFF) # deprecated for CUDA 11
    option(COMPILE_PASCAL  "Compile GPU version with SM60 support" ON)
    option(COMPILE_VOLTA   "Compile GPU version with SM70 support" ON)
    option(COMPILE_TURING  "Compile GPU version with SM75 support" ON)
    option(COMPILE_AMPERE  "Compile GPU version with SM80 support" ON)
    if (CUDA_VERSION VERSION_EQUAL "11.1" OR CUDA_VERSION VERSION_GREATER "11.1")
      option(COMPILE_AMPERE_RTX  "Compile GPU version with SM86 support" ON)
    endif()
    LIST(APPEND COMPUTE -Wno-deprecated-gpu-targets)
  endif()
  ```
  ?

XapaJIaMnu: tbh, we can do nested ifs for older versions as well

snukky: Or adding `AND CUDA_VERSION VERSION_LESS "11.1"` to the first if-block for CUDA 11, or removing all except COMPILE_AMPERE_RTX from the last if-block?

snukky: Oh, I noticed that previous blocks repeat option definitions anyway, so feel free to ignore, we may clean this up later.

XapaJIaMnu: It's going to be a lot of nested ifs... Shall we merge this for now "as is" and in the future, if/when we do refactoring, we can fix it?

snukky (approved):

PR #886, opened 2021-10-25
Task alias validation during training mode

### LINKED ISSUE #740: --task transformer doesn't do anything and gives somewhat suboptimal defaults
### Bug description
It used to be the case that we had two separate `--task` possibilities: `transformer` and `transformer-big` and this is still reflected in the options ( https://github.com/marian-nmt/marian-dev/blob/master/src/common/config_parser.cpp#L536 ) or on the website.

This is supposed to be matched by an alias entry ( https://github.com/marian-nmt/marian-dev/blob/master/src/common/aliases.cpp#L74 ) but it got renamed to `transformer-base`.  As a result, new users end up putting `--task transformer` and are getting  some largely unexpected defaults.

We should update the help description and also warn/crash when unknown task is provided, so that the user knows when they are using outdated/wrong parameters.

Cheers,

Nick

XapaJIaMnu:
### Description
This PR fixes #740 by adding an explicit validation of the task alias. Beforehand it used to be the case that if a task alias was provided that is not in `aliases.cpp` you would get random default options and things would break in unexpected ways with no feedback.

List of changes:
- Added validation of the task option if the mode is training.
- Updated the help at `config_parser.cpp`
- Removed the behaviour where aliases are removed from the config after being parsed so that we can check them later. Not sure if this breaks anything else. @graemenail  is there a better way to achieve this?

Added dependencies: none

### How to test
If you provide something like `task: transformer` you will get something like:
```
[2021-10-25 15:25:47] Error: Unrecognised task option: "transformer". Permited "transformer-base-prenorm" "transformer-big-prenorm" "transformer-big" "transformer-base" "" 
[2021-10-25 15:25:47] Error: Aborted from int mainTrainer(int, char**) in /home/USER/marian-dev/src/command/marian_train.cpp:26
```

### Checklist

### DIFF
--- a/src/common/cli_wrapper.cpp
+++ b/src/common/cli_wrapper.cpp
@@ -162,9 +162,10 @@ void CLIWrapper::parseAliases() {
   }
 
   // Remove aliases from the global config to avoid redundancy when writing/reading config files
-  for(const auto &alias : aliases_) {
-    config_.remove(alias.key);
-  }
+  // This code is commented out so we can validate the actual options of tasks
+  // for(const auto &alias : aliases_) {
+  //   config_.remove(alias.key);
+  // }
 }
 
 void CLIWrapper::updateConfig(const YAML::Node &config, cli::OptionPriority priority, const std::string &errorMsg) {
--- a/src/common/cli_wrapper.cpp
+++ b/src/common/cli_wrapper.cpp
@@ -135,11 +135,11 @@ void CLIWrapper::parseAliases() {
   // Find the set of values allowed for each alias option.
   // Later we will check and abort if an alias option has an unknown value.
   std::unordered_map<std::string, std::unordered_set<std::string>> allowedAliasValues;
-  for(const auto &alias : aliases_)
+  for(auto &&alias : aliases_)
     allowedAliasValues[alias.key].insert(alias.value);
 
   // Iterate all known aliases, each alias has a key, value, and config
-  for(const auto &alias : aliases_) {
+  for(auto &&alias : aliases_) {
     // Check if the alias option exists in the config (it may come from command line or a config
     // file)
     if(config_[alias.key]) {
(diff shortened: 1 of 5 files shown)


snukky:
@XapaJIaMnu I made some changes to your code. Can you take a look if my proposed solution does exactly what you initially wanted to change with this PR? Thanks!

I tested it manually, added a simple regression test (https://github.com/marian-nmt/marian-regression-tests/pull/82), and run regression tests again.

src/common/cli_wrapper.cpp
```diff
@@ -132,6 +132,12 @@ void CLIWrapper::parseAliases() {
   if(aliases_.empty())
     return;
 
+  // Find the set of values allowed for each alias option.
+  // Later we will check and abort if an alias option has an unknown value.
+  std::unordered_map<std::string, std::unordered_set<std::string>> allowedAliasValues;
+  for(const auto &alias : aliases_)
```
XapaJIaMnu, line 138: I prefer `auto&&` here to make sure it doesn't copy (but that's mostly an issue with 3rd party container that have poorly defined iterators... Doesn't make a difference here.

src/common/cli_wrapper.cpp
```diff
@@ -145,6 +151,12 @@ void CLIWrapper::parseAliases() {
       bool expand = false;
       if(config_[alias.key].IsSequence()) {
         auto aliasOpts = config_[alias.key].as<std::vector<std::string>>();
+        // Abort if an alias option has an unknown value, i.e. value that has not been defined
+        // in common/aliases.cpp
+        for(const auto &aliasOpt : aliasOpts)
```
XapaJIaMnu, line 156: Same as above.

src/common/cli_wrapper.cpp
```diff
@@ -145,6 +151,12 @@ void CLIWrapper::parseAliases() {
...
+          if(allowedAliasValues[alias.key].count(aliasOpt) == 0)
+            ABORT("Unknown value '" + aliasOpt + "' for alias option --" + alias.key + ". "
+                  "Run with --help to see allowed values.");
```
XapaJIaMnu, line 159: Maybe just print the allowed values here, since we have access to them?

src/common/cli_wrapper.cpp
```diff
@@ -145,6 +151,12 @@ void CLIWrapper::parseAliases() {
       bool expand = false;
       if(config_[alias.key].IsSequence()) {
         auto aliasOpts = config_[alias.key].as<std::vector<std::string>>();
+        // Abort if an alias option has an unknown value, i.e. value that has not been defined
+        // in common/aliases.cpp
+        for(const auto &aliasOpt : aliasOpts)
+          if(allowedAliasValues[alias.key].count(aliasOpt) == 0)
```
XapaJIaMnu, line 157: I personally prefer to use 
`if(allowedAliasValues.find(alias.key) == allowedAliasValues::end())`

XapaJIaMnu (commented):
So, I can't approve the pull request, because I can't approve my own pull request xD.

I think what you have done works, some stylistic suggestions based on what code i like to write, but i don't think it matters. The only suggestion is that we should print all available aliases when there is a mistake, as opposed to asking the user to rerun with `--help`.

src/common/cli_wrapper.cpp
snukky: Changed, thanks!

src/common/cli_wrapper.cpp
snukky: Added.

src/common/cli_wrapper.cpp
snukky: It would be `if(allowedAliasValues[alias.key].find(aliasOpt) == allowedAliasValues[alias.key].end())` and I kept the previous version because it was shorter.

snukky (approved):

PR #895, opened 2021-12-06
Fix AVX2+ detection on Mac

XapaJIaMnu:
MacOS is weird and its CPUflags are separated in two separate fields returned by the sysctl interfrace. To get around this, we need to test both of them, so here goes

### Checklist


snukky (commented):
I think the change in the regression-tests submodule reverts it back to an old version. Is it by mistake?

XapaJIaMnu:
@snukky now?

snukky (approved):

PR #916, opened 2022-02-09
Add dependabot for git submodules

snukky:
### Description
The dependabot will be checking for updates in submodules daily and opening pull requests with their newer versions.
Thanks, @graemenail 

Added dependencies: none

### Checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,8 @@
+version: 2
+
+updates:
+  # Maintain dependencies for Git Submodules
+  - package-ecosystem: "gitsubmodule"
+    directory: "/"
+    schedule:
+      interval: "daily"


graemenail (approved):
Looks good! 

Do you have a recommendation with regards to running regression tests on dependabot updates?

snukky:
Right, most updates in submodules will likely need to be tested via regression testing. `examples/` may be an exception. We may be adding people who done the changes in the submodule as reviewers and then it should be their responsibility to approve and/or test it properly. In ideal scenario they would already had it tested.

graemenail:
> Right, most updates in submodules will likely need to be tested via regression testing. `examples/` may be an exception. We may be adding people who done the changes in the submodule as reviewers and then it should be their responsibility to approve and/or test it properly. In ideal scenario they would already had it tested.

Seems sensible; lets merge as is.

PR #954, opened 2022-07-19
CI for Python bindings - try again

alvations:
Trying out CI for Python bindings


kpu (commented):
Disabling tcmalloc hurts performance pretty badly.

PR #968, opened 2022-10-05
best-deep alias broken

XapaJIaMnu:
### Description
The best-deep alias in marian is currently broken, because it doesn't set the model type and the default is `amum` which is incompatible with multiple layers. This commit just adds the type to the best-deep alias entry.
List of changes:
- Add `type: s2s` to the `--best-deep` alias entry.

Added dependencies: none

### How to test
Try to train a `--best-deep` model with a bare minimum config.

### Checklist


snukky (approved):
Thank you. It seems that `--type s2s` was required for `--best-deep` since forever, so that's a nice usability improvement. I think none of the regression tests should be affected, I found only one that tests best-deep but it already sets `--type s2s`: https://github.com/marian-nmt/marian-regression-tests/blob/7d612ca5e4b27a76f92584dad76d240e34f216d0/tests/interface/config/test_alias_best_deep.sh

PR #983, opened 2023-03-29
quote CPUINFO in cmake

josharian:
Without these quotes, cmake fails in a confusing manner on systems whose cpuinfo output includes spaces.

This arose in the context of attempting to compile natively on an m1 mac.

```
$ /usr/sbin/sysctl -n machdep.cpu.features machdep.cpu.leaf7_features
sysctl: unknown oid 'machdep.cpu.leaf7_features'
```

Obviously, this didn't work out well; there is still much more to do. Still, the quotes are cheap and eliminate a confusing failure mode in favor of the real problem. For this reason, I added them to the linux as well as the darwin path.

Added dependencies: none

### How to test

Run cmake. Observe that nothing has changed.

### Checklist

- [ X ] I have tested the code manually
- [ N/A ] I have run regression tests
- [ X ] I have read and followed CONTRIBUTING.md
- [ N/A ] I have updated CHANGELOG.md


snukky (approved):
Thanks for this!

PR #1014, opened 2023-11-07
add iOS and armv8 build

vrnmthr:
### Description
Adds iOS compilation support to marian, and adds memory-mapping support to TranslateService in `translator.h`. Everything should be backwards-compatible. Builds off of the work done by @XapaJIaMnu in  https://github.com/marian-nmt/marian-dev/pull/1009. Compilation for iOS is intended to be done using the added cmake toolchain file. It is detected via `CMAKE_SYSTEM_NAME` in the cmake build. 

Also fixes one bug in the memory mapping in `binary.cpp` that allows us to read mmap files with fewer memory allocations.

Added dependencies: simd_utils, ruy

### How to test
Build `cmake . -Bbuild -DCOMPILE_CUDA=off -DCMAKE_TOOLCHAIN_FILE=./cmake/ios.toolchain.cmake -DPLATFORM=OS64 -DDEPLOYMENT_TARGET=13.0` for iOS

Build `cmake . -Bbuild -DCOMPILE_CUDA=off -DARM=on` for ARM

### Checklist


src/3rd_party/spdlog/async_logger.h
hieuhoang, line 1: Any way

hieuhoang (approved):

.gitmodules
```diff
@@ -20,3 +20,9 @@
 [submodule "src/3rd_party/simple-websocket-server"]
 	path = src/3rd_party/simple-websocket-server
 	url = https://github.com/marian-nmt/Simple-WebSocket-Server
+[submodule "src/3rd_party/ruy"]
```
snukky, line 23: For all other submodules, we have their frozen forks in marian-nmt. I think we can have forks for ruy and simd_utils as well. This prevents from breakable changes, but delays pulling recent fixes, which seems what we prefer.

CMakeLists.txt
```diff
@@ -80,6 +81,38 @@ if(NOT CMAKE_BUILD_TYPE)
...
+    set(EXT_LIBS ${EXT_LIBS} ruy)
+    option(USE_RUY_SGEMM "Compile with Ruy SGEMM" ON)
+  endif(APPLE)
+
+  # Define that we are using ARM
+  add_compile_definitions(ARM FMA SSE)
+  # Some warnings as errors. I don't feel comfortable about the strict aliasing.
+  set(ARM_WARNINGS "-fno-strict-aliasing -Wno-comment")
+
+  # @TODO this assumes ArmV8. We should also look at armv7
```
snukky, line 108: Can this be problematic if ArmV8 is not available? Will there be a clear warning or error message if that's the case?

CMakeLists.txt
```diff
@@ -249,24 +284,30 @@ else(MSVC)
     set(CMAKE_RDYNAMIC_FLAG "-rdynamic")
   endif(CMAKE_COMPILER_IS_GNUCC)
 
-  set(CMAKE_CXX_FLAGS                 "-std=c++11 -pthread ${CMAKE_GCC_FLAGS} -fPIC ${DISABLE_GLOBALLY} -march=${BUILD_ARCH} ${INTRINSICS}")
-  set(CMAKE_CXX_FLAGS_RELEASE         "-O3 -m64 -funroll-loops -g ${CMAKE_RDYNAMIC_FLAG}")
+  set(CMAKE_CXX_FLAGS                 "-std=c++11 -pthread ${CMAKE_GCC_FLAGS} -fPIC ${DISABLE_GLOBALLY} ${INTRINSICS}")
```
snukky, line 287: I wonder why we were using `-march` and if it's okay to drop it in all builds (probably yes).

src/3rd_party/CMakeLists.txt
```diff
@@ -9,10 +9,20 @@ add_subdirectory(./faiss)
 include_directories(./faiss)
 
 if(COMPILE_CPU)
-  if(NOT GENERATE_MARIAN_INSTALL_TARGETS)
+  if(NOT GENERATE_MARIAN_INSTALL_TARGETS AND NOT ARM)
     set(INTGEMM_DONT_BUILD_TESTS ON CACHE BOOL "Disable intgemm tests")
     add_subdirectory(./intgemm)
   endif()
+
+  if(USE_RUY_SGEMM)
```
snukky, line 17: Is this worth a short comment in the code why it's disabled?

src/3rd_party/spdlog/async_logger.h
snukky: I agree with Hieu, if this is only cosmetics, I would suggest not doing these changes. If it's required, that's fine too since we probably already modified those 3rd party files; but let's add a comment as well so that it will be clearer what changes should be kept when one updates the 3rd party package eventually.

src/functional/operators.h
```diff
@@ -217,8 +217,11 @@ struct Ops<double> {
 // __CUDACC__ is defined when compiling with NVCC regardless of device type
 // __CUDA_ARCH__ is defined when compiling device (GPU) code
 #ifndef __CUDACC__
-
+#ifndef ARM
 #include "3rd_party/sse_mathfun.h"
+#else
+#include "3rd_party/simd_utils/simd_utils.h" // @TODO this might be dependent on NEON
```
snukky, line 223: Do we need to address the todo comment in some way before merging or not?

src/translator/translator.h
```diff
@@ -282,36 +285,62 @@ class TranslateService : public ModelServiceTask {
...
+    if(options_->get<bool>("model-mmap", false)) {
+      for(auto model : models) {
+        ABORT_IF(!io::isBin(model), "Non-binarized models cannot be mmapped");
+        LOG(info, "Loading model from {}", model);
+        model_mmaps_.push_back(mio::mmap_source(model));
+      }
+    }
+    else {
+      for(auto model : models) {
+        LOG(info, "Loading model from {}", model);
```
snukky, line 303: We should check if additional logging doesn't break some regression tests, but I think it does not.

snukky (approved):
LGTM, I only had minor suggestions and clarifying questions.

src/translator/translator.h
vrnmthr: whenever we run the regression tests we can verify this

.gitmodules
vrnmthr: all changes we needed to these are already in upstream. until we find we need to make local changes, doesn't it make sense to keep upstream? they can't do breakable changes to us until we manually bump our submodule commit.

CMakeLists.txt
vrnmthr: probably in there for additional optimizations for people building & running marian locally. if you confirm I can take it out then I can simplify the logic a bit

src/3rd_party/CMakeLists.txt
vrnmthr: it is because the default codepath does not use ruy so we don't want to add the subdirectories to the build unless it is explicitly enabled. I can add a comment

src/3rd_party/spdlog/async_logger.h
vrnmthr: they are not just cosmetic, in apple xcode <> syntax means it will look in system libs. "" syntax means it will look in user libs. so without these changes I cannot import marian .h files in xcode to integrate into iOS app

vrnmthr: I am happy to add a comment but since there are so many changed files, where do you want me to add the comment? in every file?

hieuhoang: <> means default folder for any c/c++ compiler. Usually solveable by adding -I [folder with .h files] arg to compiler. That has to be done for both CMake and Xcode.

.gitmodules
snukky: Okay, then that's fine. Thanks.

src/3rd_party/CMakeLists.txt
snukky: Adding a comment will be useful.

src/translator/translator.h
snukky: All regression tests pass with the following compilation options: `CC=/usr/bin/gcc-8 CXX=/usr/bin/g++-8 CUDAHOSTCXX=/usr/bin/g++8 cmake -DDETERMINISTIC=on -DUSE_SENTENCEPIECE=ON -DUSE_FBGEMM=on -DCOMPILE_CPU=on -DCOMPILE_TESTS=ON -DCOMPILE_EXAMPLES=ON -DCOMPILE_SERVER=ON -DCUDA_TOOLKIT_ROOT_DIR=/usr/local/cuda-11.1 -DCOMPILE_MAXWELL=on -DCOMPILE_PASCAL=on ..`

Is anything missing there?

vrnmthr: thanks for running those roman. that is not building for iOS, but I think that should be fine. that at least shows us that the canonical codepath has no regressions

src/3rd_party/CMakeLists.txt
vrnmthr: added comment

src/3rd_party/spdlog/async_logger.h
vrnmthr: I will revert these changes for now and see if we can fix at build-time

vrnmthr: if not, I may need to re-add them later

CMakeLists.txt
vrnmthr: @snukky @hieuhoang would you prefer that I leave the -march in or take it out?

src/functional/operators.h
vrnmthr: we require ARMv8 which has NEON in cmake comment

CMakeLists.txt
vrnmthr: automatically detecting is a bit tough but I agreee is a good goal. I've made it clear that for now we only support v8 and will work on adding auto archdetect in future so as to not block this PR

CMakeLists.txt
vrnmthr: I am going to leave it as is right now to not break any backwardscompat. if we want, we can take it out later

.github/workflows/ios.yml
```diff
@@ -0,0 +1,43 @@
+name: iOS
```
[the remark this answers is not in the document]
vrnmthr: This cross compiles for iOS because it builds with the iOS toolchain. That means that it builds an ARM binary. I don't think we should combine them since macos pipeline builds an Intel binary

src/3rd_party/sentencepiece
emjotde, line 1: We are updating sentencepiece? Was there a corresponding change in the sentencepiece repo?

src/3rd_party/faiss/VectorTransform.cpp
```diff
@@ -19,6 +19,10 @@
 
 using namespace faiss;
 
+#ifdef ARM
+#include "3rd_party/simd_utils/simd_utils.h"
```
emjotde, line 23: Add a comment why this is here.

src/common/binary.cpp
```diff
@@ -109,7 +109,7 @@ void loadItems(const std::string& fileName, std::vector<io::Item>& items) {
 
 io::Item getItem(const void* current, const std::string& varName) {
   std::vector<io::Item> items;
-  loadItems(current, items);
+  loadItems(current, items, /*mapped=*/true);
```
emjotde, line 112: How sure are we this is always OK?

emjotde (commented):
Request for small clarifications, will be good to merge afterwards.

src/3rd_party/sentencepiece
vrnmthr: There was -- I made some minor changes to the cmake in sentencepiece to build for iOS correctly

vrnmthr: I think github will allow you to see the diff in submodule if you click on the diff here

src/common/binary.cpp
vrnmthr: I have tested this with the iOS build and with through our quicksand paths and have had no issues. If there are other scenarios that you know of that you want to test to ensure they don't break we can try and test them as well

vrnmthr: in binary.cpp in the loadItems function there is a comment that indicates that perhaps there may be some hardware-specific compatibility we should test?
// @TOOD: verify this actually works for the hardware-specific ones like intgemm8avx2

src/3rd_party/faiss/VectorTransform.cpp
vrnmthr: added

src/common/binary.cpp
vrnmthr: if you are uncomfortable with this, we can also pass the mapped variable through the callchain to let callers set it as they see fit. we need this change in iOS otherwise loading/parsing the YAML is too slow

emjotde (approved):

src/common/binary.cpp
emjotde: There have been a lot of changes in the mmapping interface in internal master, so we will take a look in the merge PR what is going on there.

PR #1026, opened 2024-08-05
Merge with internal master - 2024-08-05

emjotde:
Merges withe MS-internal master branch. Full sync


snukky (approved):

snukky:
The workflow fails for the current master as well, so this is unlikely related to the sync. I think we can merge. Let's make sure it's not a squash merge to preserve the commit history.

PR #1028, opened 2024-08-08
Scripts for producing PyPI-compatible manylinux wheel files

thammegowda:
### Description
While the cmake build produces *.whl files, they are not distributable via PyPI. 
PyPI enforces certain rules to improve compatibility for different Linux distributions. 
This PR adds scripts for producing pymarian wheel files that can be distributable on PyPI.  

List of changes:
- add `src/python/build.sh` and `src/python/build-manylinux.sh` scripts.  \
  The former invokes `docker run` whereas the latter runs within docker env to create wheels for python version 3.8 to 3.12
- fixed an issue with Python compatibility (previously wrong headers were included for some Python versions). Solution: `set(PYBIND11_NOPYTHON On)` before adding pybind11
- loosened the constraint for `huggingface-hub` version as the strict version causes conflicts with other libs (such as transformers) and also unavailable for some version of python

Added dependencies: require docker 

### How to test

Run `src/python/build.sh` to produce wheel files at `build-python/manylinux/*.whl`

### Checklist

### DIFF
--- a/cmake/PythonModules.cmake
+++ b/cmake/PythonModules.cmake
@@ -34,10 +34,8 @@ macro(py_exec)
     endif()
 endmacro()
 
-set(PYBIND11_NOPYTHON On)
-# this wont work if pybind11 is git submodule
-#find_package(pybind11 REQUIRED)
-
+# NOTE: this property must be set before including pybind11
+# set(PYBIND11_NOPYTHON On)
 ## =====================
 set(PYTHON_SEARCH_VERSIONS 3.7 3.8 3.9 3.10 3.11 3.12 3.13)
 set(PYTHON_DISABLE_VERSIONS "" CACHE STRING "")
--- a/src/python/pyproject.toml
+++ b/src/python/pyproject.toml
@@ -33,7 +33,7 @@ dependencies = [
   "pyyaml",
   "tqdm",
   "requests",
-  "huggingface-hub==0.23.1",
+  "huggingface-hub",
 ]
 
 [project.scripts]
--- a/.gitignore
+++ b/.gitignore
@@ -50,6 +50,9 @@ build
 build-*
 # pymarian wheels
 dist/
+tmp
+tmp-*
+tmp.*
 
 # Examples
 examples/*/*.gz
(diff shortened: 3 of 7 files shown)


thammegowda:
Turns out PyPI has file size limit of 100MB by default. 
 Our statically linked cuda supported native extensions are ~600MB. `CMAKE_BUILD_TYPE=Slim` did not help reduce the extension size when CUDA is enabled.  
So I was unable to upload our packages today.


 There is a process to request the increase of file limit. For instance, pytorch packages are ~800MB and they have successfully uploaded to PyPI. I have followed their process for requesting file size limit increase. 
Reference: https://github.com/pypi/support/issues/4520 
I am not sure how long they will take to review our request and approve it. Fingers crossed.

thammegowda:
update: the limit was increased this morning, and I have uploaded packages to PyPI, which are compatible with both CUDA and intel MKL backends.

cmake/PythonModules.cmake
```diff
@@ -34,10 +34,8 @@ macro(py_exec)
     endif()
 endmacro()
 
-set(PYBIND11_NOPYTHON On)
-# this wont work if pybind11 is git submodule
-#find_package(pybind11 REQUIRED)
-
+# NOTE: this property must be set before including pybind11
```
snukky, line 37: This comment seems confusing given the commented line below. Please check if it's okay and explain in the comment or fix.

src/python/pyproject.toml
```diff
@@ -47,7 +47,7 @@ demos = [
   "flask",
   "sacremoses",
   "pyqt5",
-  "sentence-splitter@git+https://github.com/mediacloud/sentence-splitter",
+#  "sentence-splitter@git+https://github.com/mediacloud/sentence-splitter",
```
snukky, line 50: Please explain in the comment why it's commented (because not used yet but will be in pymarian-webapp?) or remove.

snukky (approved):

thammegowda:
Moved this PR to internal fork. Leaving it open here in case somebody wants to build manylinux wheels and we should close this PR once the code is synced

